PR #49, opened 2016-01-21
handle multiple return values from gopass

timothyknight:
latest gopass returns multiple values for GetPasswd


arekkas:
Hey thanks! I will take a closer look at this tomorrow

arekkas:
Thanks!

PR #54, opened 2016-02-18
Update README.md

leetal:
README: Updated smaller typo


arekkas:
Nice thanks! :) SOO cool that rethinkdb is now implemented. I hope you're getting somewhere with hydra :) 

I will deploy Hydra in it's current form into production very soon and see how it works when facing the public net!

Issue #60, opened 2016-03-11
Add Dockerfile for autobuild


arekkas:
https://hub.docker.com/r/oryam/hydra/

sbani:
My suggestions what should be done here:

## Alpine and/or debian
Create an image based on alpine `golang:1.6-alpine` and debian `golang:1.6`
### Info from the golang image
> This variant is highly recommended when final image size being as small as possible is desired. The main caveat to note is that it does use musl libc instead of glibc and friends, so certain software might run into issues depending on the depth of their libc requirements.
### Image sizes
```
$ sudo docker images -a
REPOSITORY             TAG                 IMAGE ID            CREATED             SIZE
golang                 1.6                 2529f72145a7        8 days ago          744 MB
golang                 1.6-alpine          c40da134e949        4 weeks ago         238 MB
```
## All in one package
Should we create an "all in one package" with postgres? Could be useful for testing purpose like the vagrant machine is.

## Build and push
One task is to build and push the image with the CI server (travis)

## Docker compose
Create a compose yaml to run database and hydra with one command. I don't know how to "hide" the database passwords - by environments vars?

arekkas:
Looks good! As of right now, we have multiple executables (see [cli](https://github.com/ory-am/hydra/tree/master/cli)). I was thinking of splitting those up in separate environments as this is tricky do deal with in buildpacks. Could multiple executables work with Docker or is it there an issue as well?

arekkas:
Ok, of course we could simply build multiple executables in the Dockerfile and expose them on different ports. So it shouldn't be a huge deal

sbani:
The [docker best practice](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#run-only-one-process-per-container) doc says:
> In almost all cases, you should only run a single process in a single container.

That's why we should create multiple containers for the different executables.

arekkas:
Ok, I don't think that would work with docker autobuild. Looks like we have to create multiple repositories.

PR #66, opened 2016-05-01
Build Docker images

sbani:
PR to close issue #60.

I found a way to create the Dockerfiles within' this repo and added docker stuff, tests and a (very) simple README info.

`hydra-signin` and `hydra-signup` are very simple container which are a little too big with 700MB. But it's because of the base image. When there is an `1.6-alpine-onbuild`, we should switch.

The base image is `golang:1.6-onbuild`. It's the easiest way to build the app. I will test to use alpine as the base image, but this should be the second step.

`hydra-host`: The entrypoint script checks for a linked database container and sets the `DATABASE_URL` automatically. This is a common way I borrowed from example [Word(de)press](https://github.com/docker-library/wordpress/blob/master/fpm/docker-entrypoint.sh). But I can extend this script to a better version with deeper checks. 

At the moment, `hydra-host` fails at docker install command because of a vendoring problem. Please check the error we will see in travis and help me fix it.

Unfortunately, I cannot run the official-images docker test on my Mac because it's only working on a linux system. I will need to upgrade this PR multiple times. So please wait until it's a green build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,22 @@ install:
   - go get github.com/golang/lint/golint
   - go get github.com/pierrre/gotestcover
   - go get -t $(go list ./... | grep -v /vendor/)
+  - git clone https://github.com/docker-library/official-images.git ~/official-images
 
 script:
   # - go vet -x ./...
   - gotestcover -coverprofile="cover.out" -race -covermode="count" $(go list ./... | grep -v /vendor/)
   - goveralls -coverprofile="cover.out"
+  # Test dockerfiles
+  - cd cli/hydra-host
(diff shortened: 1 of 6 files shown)


sbani:
I will wait for #67 and rebase here.

arekkas:
superseded by https://github.com/ory-am/hydra/pull/62/commits/9507df3c5b8fcae8de24e4650a6469d61106df10#diff-4e5e90c6228fd48698d074241c2ba760R1

However, I'm keeping this PR around in case of database env magic stuff.

arekkas:
this is now merged in master :)

PR #79, opened 2016-05-30
Fix idiom in README

neuhaus:
"What it looks like", not "how it looks like"
(Very common mistake)


arekkas:
Thanks, that happens a lot when it's late and you're german.. :D

PR #90, opened 2016-06-03
0.1-beta2

arekkas:
This release has backwards compability issues, because the system secret generation algorithm has changed. See #86 

* [x] import tls certificates
* [x] document tls import
* [x] store certificates in jwk store
* [x] key / env should override store jwks! (use case: started without cert, added cert later to container)
* [x] remove secret flag
* [x] `clients import clients.json`
* [x] wait for https://github.com/square/go-jose/issues/95
* [x] offline scope
* [x] allow http only
* [x] add --dry option
* [ ] mailchimp


arekkas:
I think it would be good to wait for https://github.com/square/go-jose/issues/95 to be implemented before this gets merged. The current workaround feels hacky

arekkas:
https://github.com/ory-am/fosite/pull/48

arekkas:
:open_hands: https://github.com/square/go-jose/issues/95 is now merged. let's roll

arekkas:
```
go run main.go keys create foo123 --dry --skip-tls-verify
Warning: Skipping TLS Certificate Verification.
Could not generate keys: Because you are using the dry option, the request will not be executed. You can execute this command using:

curl -X POST -d "{\"alg\":\"\",\"keys\":null}" -H "Content-Type: application/json" 'https://localhost:4444/keys/foo123
```

Issue #109, opened 2016-06-20
Warden handlers are not mounted


faxal:
@dboyne on gitter mentioned this issue.

arekkas:
closed by #110

PR #110, opened 2016-06-20
Mount warden handler

faxal:
#109


arekkas:
Whoops, thanks! I will run some tests (because travis doesn't??) and merge when everything works :)

PR #149, opened 2016-07-20
ensure client endpoint is initialised for CLI "clients import" command

boyvinall:
Signed-off-by: Matt Vinall <boyvinall@gmail.com>


arekkas:
Thanks!

PR #153, opened 2016-07-21
client: add GetName() accessor method

boyvinall:
Fairly simple addition.  Would like to add this to fosite in a sec too.

Signed-off-by: Matt Vinall <boyvinall@gmail.com>

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -22,6 +22,10 @@ func (c *Client) GetID() string {
 	return c.ID
 }
 
+func (c *Client) GetName() string {
+	return c.Name
+}
+
 func (c *Client) GetRedirectURIs() []string {
 	return c.RedirectURIs
 }


arekkas:
What's the benefit of having a method for this? `GetHashedSecret`, `GetGrantedScopes`, ... are being used to satisfy [`fosite.Client`](https://github.com/ory-am/fosite/blob/master/client.go#L12)

arekkas:
Ah wow, I overread the "Would like to add this to fosite in a sec too." part - haha :D

boyvinall:
https://github.com/ory-am/hydra/blob/master/client/manager_http.go#L17 returns a fosite.Client interface, so I can't read the name :)

arekkas:
Still, there's no reason to bloat the `fosite.Client` interface - it should only have methods that are useful to the fosite library.

arekkas:
Ah I see

arekkas:
Why not simply do `hm.GetClient('123').(*client.Client)`?

boyvinall:
Ah yeah, sorry

skymeyer:
@arekkas wouldn't it be more consistent to return ```Client``` instead of ```fosite.Client``` to line up with with ```GetClients``` ? Just a detail though.

```go
func (m *HTTPManager) GetClient(id string) (*Client, error) {
```

arekkas:
No, because `GetClient(id string) (*Client, error)` does not implement `fosite.Storage` :)

arekkas:
However, I see how this is an issue for the HTTPManager

Issue #229, opened 2016-08-22
Vendor: Upgrade to jwt-go 3.0.0


leetal:
This is essentially the same that have been done at fosite [issue 76](https://github.com/ory-am/fosite/issues/76). 

Basically the new version of jwt-go have better support for claims, and right now hydra won't build when using the head of fosite due to versioning clashes. 

I will start fixing this tomorrow and supply a PR tomorrow evening.

leetal:
@arekkas I think i have successfully updated jwt-go in hydra now as well to 3.0.0. All tests in the oauth2 package pass. BUT, the following commit [7faee6bb @ fosite](https://github.com/ory-am/fosite/commit/7faee6bbd53ee762ddfe194fb2ea5e7d0205e46d) really messes up the tests in hydra/warden. Thus, i cannot verify the changes i've made and i really think you should take a look at what that commit does to hydra.

arekkas:
ok, will do, thanks for the effort so far

arekkas:
yeah that commit is faulty

leetal:
I think it actually accepts more or less all scopes right now...

arekkas:
no i think it works fine, the test cases seem ok:

```go
	scopes = []string{"hydra", "openid", "offline"}
	assert.False(t, strategy(scopes, "foo.bar"))
	assert.False(t, strategy(scopes, "foo"))
	assert.True(t, strategy(scopes, "hydra"))
	assert.True(t, strategy(scopes, "hydra.bar"))
	assert.True(t, strategy(scopes, "openid"))
	assert.True(t, strategy(scopes, "openid.baz.bar"))
	assert.True(t, strategy(scopes, "offline"))
	assert.True(t, strategy(scopes, "offline.baz.bar.baz"))
```

arekkas:
could you open a PR with your changes so i know which branch to look at?

leetal:
On its way! Keep in mind that tests might/will fail. I cannot guarantee anything ;)

arekkas:
that's ok

arekkas:
resolved

faxal:
@arekkas @leetal I was a in a hurry with this commit: https://github.com/ory-am/fosite/commit/7faee6bbd53ee762ddfe194fb2ea5e7d0205e46d

I realized after I pushed that it can be simplified to this:

```
if current != needle {
    break
}
```

arekkas:
Ah yes, if you want, create a PR

PR #231, opened 2016-08-23
vendor: updated jwt-go to 3.0.0. Also fixed a few go vet issues.

leetal:
This is in reply to #229.

**WIP:** This uses the current HEAD ref in .glide for Fosite. This MUST be changed to a semantic version of Fosite prior to accepting this PR.

_**NOTE:** Untested_

### DIFF
--- a/oauth2/introspector_test.go
+++ b/oauth2/introspector_test.go
@@ -1,29 +1,30 @@
 package oauth2_test
 
 import (
-	"testing"
-	"time"
-	"github.com/stretchr/testify/assert"
-	"golang.org/x/net/context"
-	"github.com/ory-am/hydra/pkg"
 	"net/http/httptest"
 	"net/url"
-	"github.com/ory-am/hydra/warden"
-	"github.com/ory-am/fosite"
-	"github.com/julienschmidt/httprouter"
-	"github.com/ory-am/hydra/herodot"
-	foauth2 "github.com/ory-am/fosite/handler/oauth2"
-	goauth2 "golang.org/x/oauth2"
+	"testing"
+	"time"
+
 	"github.com/Sirupsen/logrus"
+	"github.com/julienschmidt/httprouter"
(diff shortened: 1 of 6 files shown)


oauth2/introspector_test.go
```diff
@@ -62,36 +63,36 @@ var localWarden = &warden.LocalWarden{
...
-	ar2.GrantedScopes = fosite.Arguments{"core"}
+	ar2.GrantedScopes = fosite.Arguments{0: "core"}
 	ar2.RequestedAt = now
 	ar2.Session.(*oauth2.Session).Extra = map[string]interface{}{"foo": "bar"}
 	ar2.Client = &fosite.DefaultClient{ID: "siri"}
 	fositeStore.CreateAccessTokenSession(nil, tokens[1][0], ar2)
 
 	ar3 := fosite.NewAccessRequest(oauth2.NewSession("siri"))
-	ar3.GrantedScopes = fosite.Arguments{"core"}
+	ar3.GrantedScopes = fosite.Arguments{0: "core"}
```
arekkas: this looks weird?

leetal: yeah, that is due to go vet complaining about "composite literal uses unkeyed fields". The only thing that 0 does is silencing that go vet error. Looks really ugly, but does the job.

arekkas: ah right, that's the problem I have with go vet because `gofmt -s` actually removes those

leetal: Yeah ;) I think i'll remove the 0's. Will just confuse lesser bewildered gophers.

arekkas:
ah right, the error is because the client does not have the privilege to access the warden api which was previously possible due to my mistake which @faxal solved here: https://github.com/ory-am/fosite/commit/7faee6bbd53ee762ddfe194fb2ea5e7d0205e46d

this is now #232

PR #232, opened 2016-08-23
Update jwt-go and resolve warden regression issue

arekkas:


arekkas:
This PR should not break the database. Although the Signature of JWT Containers changed, they are not being saved to the database in Hydra. Plus, the float64 change won't matter to json anyways.

@leetal let me know if you disagree

leetal:
Seems good to me! 
Will have to fix a new version of fosite as well. Time for a 0.2.5? :)

arekkas:
I released fosite as 0.3.0 because it's still a possible bc break. :)

leetal:
Ah, okey! :+1:

PR #293, opened 2016-10-09
0.6.0

arekkas:
* apiary/gb: document introspection endpoint, document separate token requirement
 * [x] apiary
 * [x] gb
* [x] apiary: document public client
* [x] gb: document public client
* [x] cmd: add public client flag
* [x] apiary: remove connection
* [x] gb: update to mysql/postgres
* [x] docker: update to mysql/postgres
* [x] readme: remove wall of text
* [x] oauth2: write client + tests for token revocation
* [x] all: sql adapters, obviously
* [x] client: come up with a better storage strategy for sql
* [x] better fosite storage tests
* [x] document warden parameters in gitbook
* [x] example is now http only, update docs
* [x] fix issue where jwks can not be found after re-boot
* [x] document `?parseTime=true` mysql requirement
* [x] document how to connect to postgres
* [x] add test for empty oauth2 client id
* [x] add aud claim to token introspection

bc breaks:
* removed connection
* mysql


andrewmcnamara:
The `DATABASE_URL` environment variable doesn't seem to be picked up anymore when starting under this branch.

arekkas:
@andrewmcnamara this branch is wip :)

arekkas:
@janekolszak check out https://github.com/ory-am/hydra/pull/293/commits/1d20c6e8f6b98d608292d45790fef10a8a6bb285

client/manager_sql.go
```diff
@@ -21,12 +34,75 @@ type SQLManager struct {
...
-	Client  []byte `db:"client"`
+	ID                string `db:"id"`
+	Name              string `db:"client_name"`
+	Secret            string `db:"client_secret"`
+	RedirectURIs      string `db:"redirect_uris"`
+	GrantTypes        string `db:"grant_types"`
+	ResponseTypes     string `db:"response_types"`
+	Scope             string `db:"scope"`
+	Owner             string `db:"owner"`
+	PolicyURI         string `db:"policy_uri"`
```
janekolszak: I'd divide this table into two - one with OIDC specific data and one with the ClientInfo. 
I'm not sure Hydra should handle ClientInfo part at all. It looks IdP specific and I know I will have to add some fields in my service and not use others. I might as well store my own version of ClientInfo

client/manager_sql.go
```diff
@@ -1,18 +1,31 @@
...
+	"strings"
 )
 
 var sqlSchema = []string{
 	`CREATE TABLE IF NOT EXISTS hydra_client (
-	id      varchar(255) NOT NULL PRIMARY KEY,
-	version int NOT NULL DEFAULT 0,
-	client  json NOT NULL)`,
+	id      	varchar(255) NOT NULL PRIMARY KEY,
+	client_name  	text NOT NULL,
```
janekolszak: Remember NOT NULL on not indexed fields is expensive (not important here, since there isn't a lot of clients)

Are all of the fields really required?

arekkas: I didn't know that, do you have any resources on that? In general it's nicer to deal with non-nullable strings in Go because Go doesn't know nil strings. tl;dr SQL null fields are a bit tricky to deal with in Go iirc.

janekolszak: damn, sorry. NOT NULL is ok. 

It's UNIQE that I was thinking about :)

arekkas:
looks like this was introduced by a recent viper commit:
```
 - name: github.com/spf13/viper
-  version: 654fc7bb54d0c138ef80405ff577391f79c0c32d
+  version: 50515b700e02658272117a72bd641b6b7f1222e5
```

arekkas:
yup, confirmed regression issue in viper: https://github.com/spf13/viper/pull/195#issuecomment-255517435

arekkas:
investigate:

```
Could not retrieve validation key: {"request":"2b5eee08-393b-4af4-8f5d-7a0f284d6539","message":": : : json: cannot unmarshal array into Go value of type map[string]ladon.jsonCondition","code":500,"description":"Could not unwrap error of type *json.UnmarshalTypeError","name":"internal-error"}
```

andrewmcnamara:
The flag "--dangerous-auto-logon" does not add the "client_id:"  to  ~/.hydra.yml.

When I checked the hydra_client table there was no 'id' or 'client_name' set. 
```bash
mysql> select * from  hydra_client \G
*************************** 1. row ***************************
            id:
   client_name:
 client_secret: $2a$10$FH07lupmVLyXL8npc/8qMOIE58h5TIxE1UQ.XqB14pMfA35JKSpfa
 redirect_uris: http://localhost:4445/callback
   grant_types: implicit|refresh_token|authorization_code|password|client_credentials
response_types: id_token|code|token
         scope: hydra openid offline
         owner:
    policy_uri:
       tos_uri:
    client_uri:
      logo_uri:
      contacts:
        public: 0
`

**Comment by andrewmcnamara** [comment]
Additionally is there documentation for how to specify the database connection to postgres or mysql ? 

hydra help host only returns configuration details for RethinkDB

```bash 
- DATABASE_URL: A URL to a persistent backend. Hydra supports various backends:
  - None: If DATABASE_URL is empty, all data will be lost when the command is killed.
  - RethinkDB: If DATABASE_URL is a DSN starting with rethinkdb://, RethinkDB will be used as storage backend.
	Example: DATABASE_URL=rethinkdb://user:password@host:123/database

	Additionally, these controls are available when using RethinkDB:
	- RETHINK_TLS_CERT_PATH: The path to the TLS certificate (pem encoded) used to connect to rethinkdb.
		Example: RETHINK_TLS_CERT_PATH=~/rethink.pem

	- RETHINK_TLS_CERT: A pem encoded TLS certificate passed as string. Can be used instead of RETHINK_TLS_CERT_PATH.
		Example: RETHINK_TLS_CERT_PATH="-----BEGIN CERTIFICATE-----\nMIIDZTCCAk2gAwIBAgIEV5xOtDANBgkqhkiG9w0BAQ0FADA0MTIwMAYDVQQDDClP..."
```

arekkas:
> Additionally is there documentation for how to specify the database connection to postgres or mysql ?

Not yet, I will add that. Thanks for reminding me :) You can check the docker-compose.yml file though!

> The flag "--dangerous-auto-logon" does not add the "client_id:" to ~/.hydra.yml.
> When I checked the hydra_client table there was no 'id' or 'client_name' set.

That's weird, it worked on my system. What was the command you ran? Did you set any environment variables? If so, which ones?

andrewmcnamara:
@arekkas Only the client_id was missing from hydra.yml.

arekkas:
But [this comment](https://github.com/ory-am/hydra/pull/293#issuecomment-255617544) is the MySQL store, no?

andrewmcnamara:
@arekkas Yes. Here is the full command I was Hydra running with.
```sh
PORT=4444 SYSTEM_SECRET=4d371c2d4265feb6ae9f3e76e7eccdab DATABASE_URL="mysql://root@tcp(127.0.0.1:3306)/hydra" CONSENT_URL=http://sample-app.dev/consents/new hydra host --dangerous-auto-logon --dangerous-force-http
```

arekkas:
Thanks, is the log file showing the client id?

andrewmcnamara:
Logfile for command
```sh
08:38:16 hydra-mysql.1 | time="2016-10-24T08:38:16+11:00" level=warning msg="No clients were found. Creating a temporary root client..."
08:38:16 hydra-mysql.1 | time="2016-10-24T08:38:16+11:00" level=info msg="Temporary root client created."
08:38:16 hydra-mysql.1 | time="2016-10-24T08:38:16+11:00" level=info msg="client_id: "
08:38:16 hydra-mysql.1 | time="2016-10-24T08:38:16+11:00" level=info msg="client_secret: 8_tpO>F468sLF&v0"
08:38:16 hydra-mysql.1 | time="2016-10-24T08:38:16+11:00" level=warning msg="WARNING: YOU MUST delete this client once in production, as credentials may have been leaked logfiles."
```

arekkas:
Thanks, I made a patch :)

arekkas:
I only tested mysql store with a fixed client id / secret and apparently this bit is not covered by unit tests, whichi s why I missed it.

andrewmcnamara:
@arekkas Thanks. That patch fixed it. I can now see the client_id in hydra.yml.

```sh
09:26:49 hydra-mysql.1 | time="2016-10-24T09:26:49+11:00" level=warning msg="No clients were found. Creating a temporary root client..."
09:26:49 hydra-mysql.1 | time="2016-10-24T09:26:49+11:00" level=info msg="Temporary root client created."
09:26:49 hydra-mysql.1 | time="2016-10-24T09:26:49+11:00" level=info msg="client_id: af1c0242-c11a-49d3-9fe9-45455998b4f9"
09:26:49 hydra-mysql.1 | time="2016-10-24T09:26:49+11:00" level=info msg="client_secret: zCS>S4iBShqvcIn6"
```

andrewmcnamara:
@arekkas Token introspection no longer returns the "aud" field which is part of the JWT spec. Instead it returns a client_id field. Can the introspect response be change to include "aud"?


Current Hydra (version 5)
```sh
hydra token validate TBpyZhWoJGB7xvuyPzL2jZPuEazYnmdm7wYfG32k3qc.mr3OtSkj90d_lf584s4dktwlO7BB0oNnkq6FNI072zk
{
	"sub": "f13d49c3-b61d-4fbc-b0a7-048a5f138115",
	"scopes": [
		"hydra"
	],
	"iss": "hydra.localhost",
	"aud": "f13d49c3-b61d-4fbc-b0a7-048a5f138115",
	"iat": "2016-10-24T14:23:57.897+11:00",
	"exp": "2016-10-24T15:23:57.897+11:00",
	"ext": null
}
```

New Hydra
```sh 
hydra token validate "qLmMAnvRtXeGs3fvKYqtsv9XqKeTe3ORZ0aDmZ6PKN4.fMIFVa7LpFYXwZ61CI_WsWNEbrLlKBsnOIgBCI6IT5g"
{
	"active": true,
	"scope": "hydra",
	"client_id": "828c4958-4359-49d6-83be-02e235841c64",
	"sub": "828c4958-4359-49d6-83be-02e235841c64",
	"exp": 1477280498,
	"iat": 1477276899
}
```

arekkas:
@andrewmcnamara per spec, client_id is the right value. we could add the aud claim too though. What's your use case?

waynerobinson:
Our use case is preventing issues with backwards compatibility based on what we've implemented. 

The existing version uses aud, not client_id as they are practically the same thing. 

And it's the way we identify a client credentials token from a standard auth one, because these fields match.

And what spec?

If you're going to use the JWT fields of iat, exp, etc. Mixing in the non-JWT client_id when aud represents the same thing is very strange.

arekkas:
https://tools.ietf.org/html/rfc7662#section-2.2

>    client_id
      OPTIONAL.  Client identifier for the OAuth 2.0 client that
      requested this token.
> 
>   aud
      OPTIONAL.  Service-specific string identifier or list of string
      identifiers representing the intended audience for this token, as
      defined in JWT [RFC7519].

We could include both, though. Btw BC breaks are part of the minor versioning and the reason why Hydra is not marked as stable (major version 1) yet. :)

waynerobinson:
But you use the Client ID as the AUD in the OpenID token now right? So the behaviour is inconsistent. And you really need some way of marking an access token as a client credentials (subject-free) version.

arekkas:
> And you really need some way of marking an access token as a client credentials (subject-free) version.

I don't think that the subject of a client credentials token should be empty. It should be the client id. In which case it's pretty easy to figure out by checking if client_id == subject.

> But you use the Client ID as the AUD in the OpenID token now right?

Yes, this is why I offered two times to include the aud claim as well!

waynerobinson:
Adding the `aud` claim would be great! 

Sorry, didn't notice that above.

waynerobinson:
Our use case is preventing issues with backwards compatibility based on
what we've implemented.

The existing version uses aud, not client_id as they are practically the
same thing.

And it's the way we identify a client credentials token from a standard
auth one, because these fields match.

waynerobinson:
And what spec?

If you're going to use the JWT fields of iat, exp, etc. Mixing in the
non-JWT client_id when aud represents the same thing is very strange.

waynerobinson:
Is the scope/s key also different?

PR #303, opened 2016-11-04
Fix client SQL manager missing client_name

johnwu96822:
Found the bug when I couldn't set/update the client name of a client.

Signed-off-by: John Wu <johnwu96822@gmail.com>

### DIFF
--- a/client/manager_sql.go
+++ b/client/manager_sql.go
@@ -73,6 +73,7 @@ var sqlParams = []string{
 func sqlDataFromClient(d *Client) *sqlData {
 	return &sqlData{
 		ID:                d.ID,
+		Name:              d.Name,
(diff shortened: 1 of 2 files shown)


arekkas:
Thanks! Could you add a test case for it too? I think it should be enough to add the additional fields [here](https://github.com/ory-am/hydra/blob/master/client/manager_test.go#L284-L289), [here](https://github.com/ory-am/hydra/blob/master/client/manager_test.go#L296-L301) and [here](https://github.com/ory-am/hydra/blob/master/client/manager_test.go#L318-L322). Not sure if it's required in all three places, one create and one update should be enough :)

arekkas:
Thanks!

PR #313, opened 2016-11-22
store/redis: redis backend for hydra

115100:
Commits for https://github.com/ory-am/hydra/issues/306

### DIFF
--- a/client/manager_redis.go
+++ b/client/manager_redis.go
@@ -1,0 +1,143 @@
+package client
+
+import (
+	"encoding/json"
+
+	"github.com/imdario/mergo"
+	"github.com/ory-am/fosite"
+	"github.com/ory-am/hydra/pkg"
+	"github.com/pborman/uuid"
+	"github.com/pkg/errors"
+	"gopkg.in/redis.v5"
+)
+
+type RedisManager struct {
+	DB        *redis.Client
+	Hasher    fosite.Hasher
+	KeyPrefix string
+}
+
+const redisClientTemplate = "hydra:client"
+
+func (m *RedisManager) redisClientKey() string {
+	return m.KeyPrefix + redisClientTemplate
+}
+
+func (m *RedisManager) GetConcreteClient(id string) (*Client, error) {
+	resp, err := m.DB.HGet(m.redisClientKey(), id).Bytes()
+	if err == redis.Nil {
+		return nil, errors.Wrap(pkg.ErrNotFound, "")
+	} else if err != nil {
+		return nil, errors.Wrap(err, "")
+	}
+
+	var d Client
+	if err := json.Unmarshal(resp, &d); err != nil {
+		return nil, errors.Wrap(err, "")
+	}
+
+	return &d, nil
+}
+
+func (m *RedisManager) GetClient(id string) (fosite.Client, error) {
+	return m.GetConcreteClient(id)
+}
+
+func (m *RedisManager) UpdateClient(c *Client) error {
+	o, err := m.GetClient(c.ID)
+	if err != nil {
+		return err
+	}
+
+	if c.Secret == "" {
+		c.Secret = string(o.GetHashedSecret())
+	} else {
+		h, err := m.Hasher.Hash([]byte(c.Secret))
+		if err != nil {
+			return errors.Wrap(err, "")
+		}
+		c.Secret = string(h)
+	}
+	if err := mergo.Merge(c, o); err != nil {
+		return errors.Wrap(err, "")
+	}
+
+	s, err := json.Marshal(c)
+	if err != nil {
+		return errors.Wrap(err, "")
+	}
+
+	if err := m.DB.HSet(m.redisClientKey(), c.ID, string(s)).Err(); err != nil {
+		return errors.Wrap(err, "")
+	}
+
+	return nil
+}
+
+func (m *RedisManager) Authenticate(id string, secret []byte) (*Client, error) {
+	c, err := m.GetConcreteClient(id)
+	if err != nil {
+		return nil, errors.Wrap(err, "")
+	}
+
+	if err := m.Hasher.Compare(c.GetHashedSecret(), secret); err != nil {
+		return nil, errors.Wrap(err, "")
+	}
+
+	return c, nil
+}
+
+func (m *RedisManager) CreateClient(c *Client) error {
+	if c.ID == "" {
+		c.ID = uuid.New()
+	}
+
+	hash, err := m.Hasher.Hash([]byte(c.Secret))
+	if err != nil {
+		return errors.Wrap(err, "")
+	}
+	c.Secret = string(hash)
+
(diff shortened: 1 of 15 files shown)


arekkas:
Awesome! @waynerobinson would you mind taking a look, too?

jwk/manager_redis.go
```diff
@@ -0,0 +1,128 @@
...
+
+	"github.com/ory-am/hydra/pkg"
+	"github.com/pkg/errors"
+	"github.com/square/go-jose"
+	"gopkg.in/redis.v5"
+)
+
+type RedisManager struct {
+	DB        *redis.Client
+	Cipher    *AEAD
```
arekkas: It doesn't look like you make any use of the Cipher? The idea is that JWKs are encrypted with the SYSTEM_SECRET and then stored in the datastore, reducing the likelihood of compromise.

client/manager_redis.go
```diff
@@ -0,0 +1,143 @@
...
+func (m *RedisManager) GetConcreteClient(id string) (*Client, error) {
+	resp, err := m.DB.HGet(m.redisClientKey(), id).Bytes()
+	if err == redis.Nil {
+		return nil, errors.Wrap(pkg.ErrNotFound, "")
+	} else if err != nil {
+		return nil, errors.Wrap(err, "")
+	}
+
+	var d *Client
+	if err := json.Unmarshal(resp, &d); err != nil {
```
arekkas: This can't work. You're referencing a nil-Pointer. This should be something like this instead:

```
var d Client
if err := json.Unmarshal(resp, &d); err != nil {
    return nil, errors.Wrap(err, "")
}
return &d, nil
```

I don't know if you've seen it, but I think the Go tour gives good ideas how this works: https://tour.golang.org/moretypes/1

oauth2/fosite_store_redis.go
```diff
@@ -0,0 +1,282 @@
+package oauth2
```
arekkas: note-to-self: needs review

client/manager_redis.go
```diff
@@ -0,0 +1,143 @@
...
+	"gopkg.in/redis.v5"
+)
+
+type RedisManager struct {
+	DB        *redis.Client
+	Hasher    fosite.Hasher
+	KeyPrefix string
+}
+
+const redisClientTemplate = "hydra:clients"
```
arekkas: Do you want singular for the table names or plural? Because here it is `hydra:clients` but in the JWK it's singular `hydra:jwk`. Since it's singular in sql/rethink I would recommend having it singular here as well

jwk/manager_redis.go
115100: Yes, I forgot about this completely. Fixed.

client/manager_redis.go
115100: Yeah, I'll keep it consistent and make this singular.

client/manager_redis.go
115100: Strangely enough, it hasn't broken tests locally :/. I'll change it to that because I have no idea why it wouldn't fail.

115100: Hmm. It's panicking for me with the newer commit.

arekkas:
So sorry it took some time. I will review now!

jwk/manager_redis.go
arekkas: Thanks!

cmd/server/handler_client_factory.go
```diff
@@ -40,6 +40,12 @@ func newClientManager(c *config.Config) client.Manager {
 		}
 		m.Watch(context.Background())
 		return m
+	case *config.RedisConnection:
+		m := &client.RedisManager{
```
arekkas: does redis need something like schema generation?

cmd/host.go
```diff
@@ -44,6 +44,9 @@ CORE CONTROLS
 	- RETHINK_TLS_CERT: A pem encoded TLS certificate passed as string. Can be used instead of RETHINK_TLS_CERT_PATH.
 		Example: RETHINK_TLS_CERT_PATH="-----BEGIN CERTIFICATE-----\nMIIDZTCCAk2gAwIBAgIEV5xOtDANBgkqhkiG9w0BAQ0FADA0MTIwMAYDVQQDDClP..."
 
+  - Redis: If DATABASE_URL is a DNS starting with redis:// Redis will be used as a storage backend.
+		Example: DATABASE_URL=redis://x:password@host:123/0
```
arekkas: What is the `/0` for?

arekkas: I think it would be better if the port are actually the defaults of the databases (same goes for mysql/postgres/rethink). I will fix it for those three, but could you replace `:123` with the redis default port?

cmd/server/handler_client_factory.go
115100: No, I've kept everything quite similar to the memory manager so it's all done in Hydra.

cmd/host.go
115100: The `/0` is the database number to connect to:  https://redis.io/commands/SELECT.

I'll fix the port number later.

cmd/server/handler_client_factory.go
arekkas: Ok so it works with a fresh redis instance?

115100:
I've found a small bug so I'll fix that too.

arekkas:
One more thing, could you add redis to the [docker-compose.yml](https://github.com/ory-am/hydra/blob/master/docker-compose.yml#L43), so it's easy to spin up that environment and test it?

cmd/server/handler_client_factory.go
115100: Yeah, I ran my local tests after a FLUSHDB and it started up fine.

115100:
I've added a few commits to a WIP branch here: https://github.com/115100/hydra/tree/feature/redis-backend-wip

I don't have a proper environment set up here so I'll have to test on Monday before adding to the PR.

arekkas:
Ok, you still need to link to redis [here](https://github.com/ory-am/hydra/commit/38e4c984ac358b53bcc76741628c2b02ad8b6c4b#diff-4e5e90c6228fd48698d074241c2ba760R11) and use `redis` instead of localhost [here](https://github.com/ory-am/hydra/compare/master...115100:feature/redis-backend-wip?expand=1#diff-4e5e90c6228fd48698d074241c2ba760R25)

docker-compose.yml
```diff
@@ -53,6 +56,9 @@ services:
     environment:
       - MYSQL_ROOT_PASSWORD=secret
 
+  redisd:
```
arekkas: maybe uncomment this and mysql too, so it doesn't pull unneccessary images

arekkas:
thank you so much for your hard work, @115100 !

115100:
Thanks for merging!

arekkas:
thank you for contributing!

PR #324, opened 2016-12-12
Log fixes

johnwu96822:
1. Fix log level missing the "error" level set from LOG_LEVEL
2. Do not display datastore username and password in the log

Signed-off-by: John Wu <johnwu96822@gmail.com>


arekkas:
thanks!

Issue #327, opened 2016-12-15
cmd: add configuration options for `hydra token user`


tamarrow:
Config options for redirect url, client secrets and scope would be most helpful.

Issue #326, opened 2016-12-14
sql: deleting policies does not delete associated records with mysql driver


johnwu96822:
In MySQL there are four tables for ladon policies. When deleting a policy, it is only deleted from "ladon_policy" table, but the data for this policy in three other ladon_policy_* tables are not removed.

johnwu96822:
This happens on MySQL

arekkas:
@johnwu96822 is this a critical issue for you?

johnwu96822:
@arekkas Yes we have deployed Hydra for testing. If this issue doesn't get fixed and we modify policies (Hydra currently modifies a policy by **deleting** and recreating it), we'll get a bunch of policy records in those tables and we wouldn't know which record we'd get.

arekkas:
see https://github.com/ory-am/hydra/pull/329/commits/070520d3d862e39ce1a199be7f50f95380ad7e1e

arekkas:
resolved by #329

Issue #194, opened 2016-08-09
include a migration routine for databases


arekkas:

janekolszak:
I think all database code should be in a separate folder, in separate sql scripts. All tables in one place.
First script would be `001-create-db.sql`. When there's an update we add a script `002-some-update.sql`
In order to create the database hydra would:
- create transaction
- run scripts in order
- commit

In order to update the database schema hydra would:
- create transaction
- run scripts from the updated version to the latest
- commit

If hydra does this automatically we get rid of the problematic (for user) schema updates.

arekkas:
a list of sql migration tools:

* https://github.com/DavidHuie/gomigrate
* https://github.com/rubenv/sql-migrate
* https://github.com/mattes/migrate

20zinnm:
I highly recommend migrate--you can see a great usage of it [here](https://github.com/drone/drone/tree/master/store).

arekkas:
awesome, thanks!

arekkas:
see https://github.com/ory-am/hydra/pull/329/commits/070520d3d862e39ce1a199be7f50f95380ad7e1e

Issue #68, opened 2016-05-03
warden: add group management and group based policy checks


arekkas:
Subpackage to policy:

``` go
package group

type Group struct {
    ID       string   `json:"id"`
    Members  []string `json:"members"`
    Policies []string `json:"policies"`
}

type Groups map[string]*Group

type Manager interface {
    GetGroup(id string) (Group, error)

    JoinGroup(user, group string) error

    LeaveGroup(user, group string) error

    CreateGroup(group Group) error

    AddGroupPolicy(policy, group string) error

    RemoveGroupPolicy(policy, group string) error

    DeleteGroup(id string) error
}
```

PR #329, opened 2016-12-19
0.7.0: SQL Migrate, Groups, Hardening

arekkas:
This PR bumps Hydra to version 0.7.0. We expect this release to be the forelast release before the stable 1.0.0. Version 0.8.0 is anticipated to include #297 .

This PR includes the following changes:

* cmd: add configuration options for `hydra token user` #327 
* warden: add group management and group based policy checks #68 
* docs: improve 5 min tutorial
* sql: include a migration routine for databases #194

This PR includes the follwoing **breaking** changes:

## sql: deleting policies does not delete associated records with mysql driver #326 

MySQL has a [12 year old bug](http://bugs.mysql.com/bug.php?id=4919) that ignores shorthand foreign key constraints in `CREATE TABLE` statements. Those were used in ladon, the library responsible for access control policies. This patch resolves this issue by replacing shorthand constraints with explicit indices. You will be able to run your existing database with this patch, but it will not resolve the bug described in #326. However, you can easily update the foreign key constraints manually in your database, the schema [is here](https://github.com/ory-am/ladon/blob/master/manager_sql.go#L25-L41).

Postgres is not affected.

## oauth2/consent: force jti echo in consent response #322

Previously, Hydra accepted signed consent tokens without requiring a special JSON Token ID in the JWT payload. This was deemed secure enough, as expiry times are usually short and OAuth2 clients use random states. However, to make things more secure, the consent app has to include the `jti` value from the consent challenge in the payload of the consent response. Hydra sets a cookie when redirecting to the consent app with the JTI value for validation of the consent response. The cookie is encrypted with a secret passphrase given by `COOKIE_SECRET` and if none is given, falls back to `SYSTEM_SECRET`. We recommend using a dedicated `COOKIE_SECRET` in production.

Read more [here](https://ory-am.gitbooks.io/hydra/content/oauth2.html#consent-app-flow).


arekkas:
If you have open questions regarding this release, feel free to post them here.

arekkas:
Currently, the group check ignores `ladon.DenyAccess` if one of the other groups has `ladon.AllowAccess`. Same goes for subjects. But the order should be `deny > allow > none`

PR #331, opened 2016-12-19
cmd: replace newline in HTTP_TLS

ewilde:
HTTPS_TLS_CERT and HTTPS_TLS_KEY environment variables can contain `\n`
see:https://github.com/ory-am/hydra/blob/master/cmd/host.go
This commit replaces the `\n` characters with an actual newline to allow
the tls package to correctly create a X509 key pair.

Otherwise the X509 key pair fails to get created.

Signed-off-by: Edward Wilde <ewilde@gmail.com>

### DIFF
--- a/cmd/server/helper_cert.go
+++ b/cmd/server/helper_cert.go
@@ -17,6 +17,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"github.com/square/go-jose"
+	"strings"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


arekkas:
Does this work on windows / linux / osx alike?

arekkas:
Because windows has `\r\n` right?

ewilde:
I will test it on windows and get back to you. I've only tried is on mac and linux

ewilde:
Seems to load okay on window

image

arekkas:
Cool, thanks for investigating!

PR #389, opened 2017-03-08
docs: add example policy for consent app signing

arekkas:


docs/access-control/policies/allow-client-consent-key
```diff
@@ -0,0 +1,15 @@
+{
+  "description": "Allow a client to read the keys neccessary for performing the consent flow. Make sure to replace <my-client-id> with the client id you chose.",
+  "subjects": [
+    "<my-client-id>"
+  ],
+  "effect": "allow",
+  "resources": [
+    "rn:hydra:keys:hydra.consent.response:private",
```
Moggers: What is response public vs response private?

arekkas: The private key is used for signing the consent response, the public key is looked up by hydra to verify the consent response. The client does not neccessarily need acces to the public consent key. Please check out https://github.com/ory/hydra-js for an easy to use js API on this topic

PR #404, opened 2017-03-30
Updated ladon version in glide.lock

ericalandouglas:
The new custom policy `StringPairsEqualCondition ` in Ladon is included with the latest commit to the project. Now updating Hydra's `glide.lock` file to respect this newest commit in Ladon.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -131,7 +131,7 @@ imports:
   - token/hmac
   - token/jwt
 - name: github.com/ory-am/ladon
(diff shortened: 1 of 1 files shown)


ericalandouglas:
I have not updated the `glide.yaml` file present in this project, only the `glide.lock` file has been modified. If there is a particular version I should use and a place where I can derive the right version number I will updated accordingly.

arekkas:
Thanks, that works for now :)

PR #416, opened 2017-04-12
Custom consent URIs

janekolszak:


arekkas:
Is this a draft? It does not look like the consent URL is used anywhere.

I feel very discomfortable with this feature though. Of the top of my head, here's an attack scenario:

1. Malicious user (Mali) sets up a client in bob's developer portal. He sets the consent url to `http://malicious.com/login`
2. Mali duplicates the log in screen from bob's website `http://bobwebsite.com/login`
3. Mali tells alice to use his app. Alice performs the oauth2 flow and lands at `http://malicious.com/login`. She thinks however that she is at `http://bobwebsite.com/login` because the sites look the same.
4. Alice enters her username and password.
5. Mali has now alice's username and password.

janekolszak:
Yeah, it's work in progress. I'll use CI from time to time. Do you get a lot of spam because of that?

This scenario is a valid problem, but it's Bob's role not to let Mali register a malicious client. 
There's a difference between trusted clients (Bob's desktop app, Bob's mobile app etc.) and third party clients. 

That's why by default if there's no client_uri or it's empty the default client_uri should be used. The old behavior will be preserved, don't worry.

arekkas:
> This scenario is a valid problem, but it's Bob's role not to let Mali register a malicious client.
There's a difference between trusted clients (Bob's desktop app, Bob's mobile app etc.) and third party clients.

This is the problem exactly, because now we assume that Bob has implicit (because he read the documentation) knowledge of this issue. The same thing is achievable with getting the `aud` claim from the consent challenge and looking it up in a "routing" table, e.g.

```
switch aud:
  case "my-app":
    redirect "http://my-app/login"
  default:
    redurect "http://my-other-app/login"
```

Here, Bob has explicit control and knowledge of the routing logic.

janekolszak:
You are right, but there's this extra redirect that we could avoid.

arekkas:
That is true, however the benefits outweigh the downsides IMO. However, I'll sleep over it and try to think of a way that we could add this without risking the attack scenario above to system administrators who are not very cautious :)

arekkas:
I thought about it and I feel even more uncomfortable with this change. The downside of having another redirect is minimal (averaging on 80ms?) while the downside of risking attacks such as the one mentioned above is maximal.

I would like to thank you for your initiative on this, but I believe that it's not the right path forward for Hydra. If you feel strongly about this, feel free to chat with me on gitter or in this issue.

janekolszak:
There are some other options you can sleep on:
- Switch this on, during build time
- Switch this on with CLI parameter

Fixing this in IdP mixes the roles. Hydra should store all OIDC data. It will be much harder to break into Hydra than to break into someones IdP since Hydra get a lot of reviews.

arekkas:
> Fixing this in IdP mixes the roles.

In my opinion, identity brokerage (which this is effectively) is subject to the consent app, as the consent app would also broker identities to Google/GitHub/Facebook auth :)

janekolszak:
OK, have it your way, but I still think it's overprotective.

PR #423, opened 2017-04-14
Remove uname references from docs

### LINKED ISSUE #418: Uname in session
Am I right in believing that the uname is not set in the session?
https://github.com/ory/hydra/blob/master/oauth2/consent_strategy.go#L93

I wanted to make use of it: 

> uname: You can set an arbitrary, non-unique username which will be echoed in the token introspection. (optional)

matteosuppo:
Fix #418


arekkas:
Thank you!

PR #427, opened 2017-04-17
Wellknown

Grillz:
Adds discovery endpoints that were started in branch `fix-379`. Tested an working, opening this as a WIP, let me know if there is anything you want added/changed. Thanks


oauth2/handler.go
```diff
@@ -49,6 +58,20 @@ func (h *Handler) SetRoutes(r *httprouter.Router) {
 	r.GET(ConsentPath, h.DefaultConsentHandler)
 	r.POST(IntrospectPath, h.IntrospectHandler)
 	r.POST(RevocationPath, h.RevocationHandler)
+	r.GET(WellKnownPath, h.WellKnownHandler)
+}
+
+func (h *Handler) WellKnownHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	fmt.Printf("handler: %+v", h)
```
arekkas: Please remove this

cmd/server/handler_oauth2_factory.go
```diff
@@ -155,6 +155,8 @@ func newOAuth2Handler(c *config.Config, router *httprouter.Router, km jwk.Manage
 		H:                   &herodot.JSON{},
 		AccessTokenLifespan: c.GetAccessTokenLifespan(),
 		CookieStore:         sessions.NewCookieStore(c.GetCookieSecret()),
+		Issuer:              c.Issuer,
+		BaseURL:             c.ClusterURL,
```
Grillz: This was the one part I was unsure of, where should the `BaseURL` come from? Issuer might actually make the most sense

arekkas: Hm, good question, this was actually a problem previously. The cluster url is usually set on the client side, not the server side. We could either use the issuer url, or make it easier to understand by renaming ISSUER env var to something like PUBLIC_URL. what do you think?

Sorry by the way for taking so long on responding, the last days were really busy.

Grillz: I think `Issuer` is fine actually, thats what its generally called by most implementers. I suppose I should add a section to the docs that emphasizes its use. Would it be ok to remove `BaseURL` then?

arekkas: Yes, sounds good!

arekkas: Besides adding it in the docs, also make the usage clear in `hydra help host` :)

Grillz:
@arekkas this guy is good to go, I'm getting blocked on the coveralls, in order to increase the code coverage we would have to implement HTTP tests, which seems non-trivial. Let me know how you want to proceed here

arekkas:
Any idea why travis is failing?

oauth2/handler.go
```diff
@@ -40,6 +45,9 @@ type Handler struct {
 
 	AccessTokenLifespan time.Duration
 	CookieStore         sessions.Store
+
+	Issuer  string
+	BaseURL string
```
arekkas: This is no longer needed right?

Grillz:
CI is good now, not sure what that blip was, also removed the unnecessary parameter from the oAuth handler

arekkas:
Thanks! That's perfect. However, I have to be the mean maintainer here and insist on a test case for this :( It helps keep the code quality high and also test for potential errors not visible to code review.

Testing this should be easy. For the oauth2 package, set up a `handler_test.go`, bootstrap the handler itself (should be easy, just set the issuer - there is no other logic required) and do a http request against it, and test the output against what you expect.

Once that is done, I'll help you setting up the test for the JWK.

Grillz:
ok no worries, I'll cook those up

Grillz:
Ok @arekkas, I've added tests for both routes. Some additional changes were necessary per https://github.com/ory/hydra/issues/433, having key ids as "public" or "private" is not sufficient when they are returned in the well-known response. The key id must be unique, so I added `set` to the key generator here https://github.com/ory/hydra/pull/427/files#diff-98aed7eaac1441068d28c70047dd9fdbR21 . Let me know if you think there is a better way of handling that. Thanks!

arekkas:
It's weird that the tests are failing, I don't think it's your code's fault, I tracked it as #443

cmd/server/helper_keys.go
```diff
@@ -18,7 +18,7 @@ func createRS256KeysIfNotExist(c *config.Config, set, kid, use string) {
 	if _, err := ctx.KeyManager.GetKey(set, kid); errors.Cause(err) == pkg.ErrNotFound {
 		logrus.Infof("Key pair for signing %s is missing. Creating new one.", set)
 
-		keys, err := generator.Generate("")
+		keys, err := generator.Generate(set)
```
arekkas: I'm not entirely sure why I did `generator.Generate("")` here. Can you confirm that the docker-compose example works with your patch?

oauth2/handler.go
```diff
@@ -49,6 +65,20 @@ func (h *Handler) SetRoutes(r *httprouter.Router) {
 	r.GET(ConsentPath, h.DefaultConsentHandler)
 	r.POST(IntrospectPath, h.IntrospectHandler)
 	r.POST(RevocationPath, h.RevocationHandler)
+	r.GET(WellKnownPath, h.WellKnownHandler)
+}
+
+func (h *Handler) WellKnownHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	known := WellKnown{
+		Issuer:        h.Issuer,
+		AuthURL:       h.Issuer + AuthPath,
```
arekkas: I don't feel comfortable about this, what if the issuer is formated as `http://foo.bar/foo/` and the AuthPath as `/foo`, then we get `http://foo.bar/foo//foo` as a result. For resolving this, see [this SO question](http://stackoverflow.com/questions/34668012/combine-url-paths-with-path-join)

Grillz: I agree I added a trim suffix to `root.go` https://github.com/ory/hydra/pull/427/files#diff-ff7686b39bf90dc2520886fb874371a4R125 , do we want to do that for any of the other urls there?

cmd/server/helper_keys.go
Grillz: The docker-compose example works, I went through all of the flows in the tutorial. So this changes the generic key names to `public:id-token-foo`, which in turn changes a couple policies as documented here https://github.com/ory/hydra/pull/427/files#diff-08781ef00709d5059315ac5095fafabbR73 and https://github.com/ory/hydra/pull/427/files#diff-0704e4259cd334d9340c20bd02095628R8 . This has the downside of making the policy a bit more confusing. I suppose there aren't many policies that need to be written for the root keys, but let me know if you see a better way.

If we do want to go ahead with that, there are quite a few references to the `kid` being "public" or "private", the easiest way to get by that may be to add a `getKeyFromType` function that could be swapped out fairly easily.

oauth2/handler.go
arekkas: Oh, I totally missed that! Then this is fine of course :)

cmd/server/helper_keys.go
arekkas: > So this changes the generic key names to public:id-token-foo, which in turn changes a couple policies as documented here

Yeah that's what I thought which is why I asked the question. I don't know how I feel about that. Does it change the URLs which are used to look up the keys as well?

Grillz: The API contract would remain the same, but there may need to be some changes to `/keys/set` on the backend. 

Google:   
https://www.googleapis.com/oauth2/v3/certs   
MS:   
https://login.windows.net/common/discovery/keys  

They are using just a uuid for the `kid`, so thats another option.

arekkas: Ok, I see, I now also understand the change better. I think going with a timestamp, a uuid or some other unique identifier is the way to go here, the doubling of the name feels kinda weird.

However, this will impact the way we query for those keys within hydra, and maybe also in the SDK. So I think this is your suggestion for the key type?

I understand now the need for this, but it also complicates things a bit, so we need to document this, especially if we expose the KeyFromType functionality to the SDK/REST.

Grillz: Ok, I've been thinking on the least breaking change here, and let me run this by you. What if `getKey()` supported wildcards? Then the policy could be `mykeyset:public:29873489234`, to retrieve that key you could run `getKey("mykeyset", "public*")` which would return all public keys in that set.

Grillz: so I forgot wildcards aren't supported in url schemes. Perhaps we could do a `getKey("mykeyset", "public:all")`

arekkas: Couldn't get key just return the most recent key?

Grillz: I think that would work as long as we are ok with reserved keywords of `public` and `private` referring to the most recent key of that type in the set. Policy could still be written as normal, but the actual key-id would be `public:12345`. Obviously we would need heavy documentation on the matter

arekkas: Yeah you're right here. Damn this one is tricky. Couldn't we be doing the following:

Public and private are reserved keys. Only one key can have the key-id. On, for example, key rotation, the keys with kid public and private get reassigned something like `private:<timestamp/uuid>` and the new keys get public/private?

Grillz: The issue with that is when you return all the public keys in the `well-known` response, they would all have the `kid` of `public`, and the breaks a lot of client libraries that are expecting a unique `kid` there, which is matched to the `kid` in the jwt header. So the `kid` needs to be more enriched than just `public` but perhaps we can still reference it that way with a layer of abstraction

arekkas: Oh I think you misunderstood, the kid `public` and `private` are unique - like a fixed primary key. However, if multiple keys exists, they have a different id, for example `1234` or `public-1234`. This way we only have one public kid in the set. Does that make sense?

Grillz: So that could work, but when I add the `kid` to the JWT header, which would need to happen here https://github.com/ory/hydra/blob/master/oauth2/consent_strategy.go#L102 I would need to know what that ID is. We could have the understanding that the `kid` in the JWT header is appended with a known value -- say a hash of the set name, and then when we send the keys in the `well-known` response we append that as well. 

The downside of this is that if a person looks at the key in `well-known` and tries to make an API call to hydra modifying it, that `kid` it wouldn't exist. The same could be said for writing policy.

A bit of a rock and a hard place 😄

arekkas: I think I wasn't able to get the point across yet :D 

When booting Hydra on a new database, the signing keys for OIDC will be generated with:

```
{
set: "openid",
keys: [{kid: "private"}, {kid: "public"}]
}
```

In this case, the kids private and public are unique because the set does not have any other keys. Let's say the administrator of Hydra now adds a new keys to the set by accident:


```
{
set: "openid",
keys: [{kid: "private"}, {kid: "public"}, {kid: "public:just-any-key-whatever"}]
}
```

The private and public keys are still unique in this case - so no problem. Let's say the admin now wants to do some fancy key rotation. First, he renames the keys private and public:

```
{
set: "openid",
keys: [{kid: "private:old"}, {kid: "public:old"}]
}
```

next he adds the new keys:

```
{
set: "openid",
keys: [
  {kid: "private"}, {kid: "public"},
  {kid: "private:old"}, {kid: "public:old"}
]
}
```

Now, we have four keys. The old ones plus the fresh ones from key rotation. In a later stage I want to provide automatic key rotation, so this process would be automated by, for example, calling `PUT /keys/rotate/openid` with the result being something like this:


```
{
set: "openid",
keys: [
  {kid: "private"}, {kid: "public"},
  {kid: "private:some-uuid-or-a-timestamp-or-something-else"}, {kid: "public:some-uuid-or-a-timestamp-or-something-else"}
]
}
```

Does this make more sense now?

arekkas:
Thank you @Grillz for staying on this issue for so long, I really appreciate it!

cmd/server/helper_keys.go
Grillz: So the confusion here is stemming from which keys are returned in the `well-known` response. I am assuming that all of the public keys are returned, hence each set would have a `kid` of `public` and then you put all of those public keys in a single set and return them. Leading to 3 keys with the `kid` of `public`. 

Whats interesting as I read through the OpenID spec http://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata again, they don't specify which key types should be returned. Looking at Okta http://developer.okta.com/docs/api/resources/oauth2.html#get-authorization-server-keys they are returning just one key type, but with an added `status` parameter.

I was seeing 3 keys from google and ms thinking it was `consent, challenge, and idtoken`, but it may be `past, present, future` of the same key.

arekkas: Oh wow, this took me some time to get it. I didn't get that well-known is aggregating all those keys and putting them in the same set. Now this all makes sense, sorry for being so slow on understanding this :D 

Ok, so how about we simply prefix the kid with the set id in the well-known endpoint? Or is that a hack job?

Grillz: Yeah so the more I look at this and think about it, I think you may have been right to assume that its just the `idtoken` key that needs returned. I can't find any reason to have the `consent` or `challenge` keys in that set. I'll update to reflect that, sorry about the run around. 

On the upside we've had a chance to look at key rotation within the sets. I'll leave that for another PR though 😉

arekkas: Yeah, only returning the keys for openid connect makes sense IMO :)

arekkas:
@Grillz to get this working I think you need to run glide update

Grillz:
ok @arekkas I think we are close on this guy 😄, the one piece I couldn't figure out is how you are generating the swagger spec. I see https://github.com/ory/hydra/blob/master/gen-swagger.sh but thats generating a json file, are you just converting that to yaml?

Let me know if there is anything else you see, and I will squash before we merge.

arekkas:
Yeah I'm actually converting that manually via [here](http://swagger.io/swagger-editor/) (just paste the json and it will ask you if it's ok to convert to yaml). I'll take another 👀 tomorrow, 11.55pm over here :)

docs/api.swagger.yaml
```diff
@@ -169,7 +233,6 @@ paths:
           description: The id of the OAuth 2.0 Client.
           name: id
           in: path
-          required: true
```
arekkas: This is actually required, was it removed by accident?

docs/api.swagger.yaml
```diff
@@ -287,7 +350,6 @@ paths:
...
```
arekkas: This is actually required, was it removed by accident?

docs/api.swagger.yaml
```diff
@@ -637,9 +699,7 @@ paths:
       summary: The OAuth 2.0 Auth endpoint
       operationId: oauthAuth
       responses:
-        '302': {
-          description: "found"
-        }
+        '302': {}
```
arekkas: Without description, this causes a schema error, could you revert this change?

docs/api.swagger.yaml
```diff
@@ -697,8 +757,7 @@ paths:
                 type: string
                 x-go-name: RefreshToken
       responses:
-        '200':
-          description: "ok"
+        '200': {}
```
arekkas: Without description, this causes a schema error, could you revert this change?

docs/api.swagger.yaml
```diff
@@ -900,7 +958,6 @@ paths:
           description: The id of the policy.
           name: id
           in: path
-          required: true
```
arekkas: This is actually required, was it removed by accident?

docs/api.swagger.yaml
```diff
@@ -950,7 +1007,6 @@ paths:
...
```
arekkas: This is actually required, was it removed by accident?

docs/api.swagger.yaml
```diff
@@ -1114,7 +1170,6 @@ paths:
           description: The id of the group to look up.
           name: id
           in: path
-          required: true
```
arekkas: This is actually required, was it removed by accident?

docs/api.swagger.yaml
```diff
@@ -1157,7 +1212,6 @@ paths:
           description: The id of the group to modify.
           name: id
           in: path
-          required: true
```
arekkas: This is actually required, was it removed by accident?

docs/api.swagger.yaml
```diff
@@ -1202,7 +1256,6 @@ paths:
...
```
arekkas: This is actually required, was it removed by accident?

docs/api.swagger.yaml
```diff
@@ -1713,6 +1902,7 @@ definitions:
     type: object
     properties:
       Body:
+        description: 'in: body'
```
arekkas: This looks like a fluke?

docs/api.swagger.yaml
```diff
@@ -1732,6 +1922,7 @@ definitions:
       - set
     properties:
       Body:
+        description: 'in: body'
```
arekkas: This looks like a fluke?

docs/api.swagger.yaml
```diff
@@ -1778,6 +1969,7 @@ definitions:
...
```
arekkas: This looks like a fluke?

docs/api.swagger.yaml
```diff
@@ -1798,6 +1990,7 @@ definitions:
...
```
arekkas: This looks like a fluke?

docs/api.swagger.yaml
```diff
@@ -1829,12 +2022,14 @@ definitions:
     type: object
     properties:
       Body:
+        description: 'in: body'
```
arekkas: This looks like a fluke?

docs/api.swagger.yaml
```diff
@@ -1829,12 +2022,14 @@ definitions:
...
         $ref: '#/definitions/policy'
     x-go-package: github.com/ory/hydra/policy
   swaggerUpdatePolicyParameters:
     type: object
     properties:
       Body:
+        description: 'in: body'
```
arekkas: This looks like a fluke?

glide.lock
```diff
@@ -43,6 +43,7 @@ imports:
   - pkg/system
   - pkg/term
   - pkg/term/windows
+  - pkg/testutil/assert
```
arekkas: I think you imported the wrong assert package somewhere :)

jwk/handler.go
```diff
@@ -62,6 +67,63 @@ type joseWebKeySetRequest struct {
 	Keys []json.RawMessage `json:"keys"`
 }
 
+// swagger:route GET /.well-known/jwks.json jwk WellKnown
```
arekkas: Awesome, thank you so much! Could you replace the `jwk` tag with `oauth2` and `jwks` and `openid-connect`? :)

jwk/handler_test.go
```diff
@@ -0,0 +1,75 @@
...
+	. "github.com/ory/hydra/jwk"
+	"github.com/julienschmidt/httprouter"
+	"github.com/ory/fosite"
+	"github.com/ory/hydra/compose"
+	"github.com/ory/herodot"
+	"github.com/ory/ladon"
+	"net/http"
+	"encoding/json"
+	"github.com/square/go-jose"
+	"github.com/docker/docker/pkg/testutil/assert"
```
arekkas: There it is! Replace this with testify assert :)

jwk/handler_test.go
```diff
@@ -0,0 +1,75 @@
...
+	testServer = httptest.NewServer(router)
+}
+
+
+
+func TestHandlerWellKnown(t *testing.T) {
+
+	JWKPath := "/.well-known/jwks.json"
+	res, err := http.Get(testServer.URL + JWKPath)
+	if err != nil {
```
arekkas: This can be simplified with:

```
require.NoError(t, err, "problem in http request")
```

jwk/handler_test.go
```diff
@@ -0,0 +1,75 @@
...
+	defer res.Body.Close()
+
+	var known jose.JsonWebKeySet
+	err = json.NewDecoder(res.Body).Decode(&known)
+	if err != nil {
+		t.Errorf("problem decoding well known response: %v", err)
+	}
+
+	resp := known.Key("public")
+	if resp == nil {
```
arekkas: This can be simplified with:

```
require.NotNil(t, err, "Could not find key public")
```

oauth2/handler.go
```diff
@@ -3,12 +3,11 @@ package oauth2
...
 	"net/http"
 	"net/url"
-
 	"encoding/json"
-	"strings"
-	"time"
 
 	"github.com/Sirupsen/logrus"
+	"strings"
+	"time"
```
arekkas: I'm not super picky on this but since you probably will take another pass at this anyways, could you run `goimports` on this file? Please on this file only so we get a clean changeset!

oauth2/handler.go
```diff
@@ -42,8 +44,21 @@ type Handler struct {
...
-
+	
 	L logrus.FieldLogger
+
+	Issuer string
+}
+
+// swagger:model WellKnown
+type WellKnown struct {
+	Issuer        string   `json:"issuer"`
```
arekkas: Nice! Thanks for adding the model here :) You could copy the descriptions from the JWK RFC here so it's clear what each of those values are.

oauth2/handler.go
```diff
@@ -53,6 +68,41 @@ func (h *Handler) SetRoutes(r *httprouter.Router) {
 	r.GET(ConsentPath, h.DefaultConsentHandler)
 	r.POST(IntrospectPath, h.IntrospectHandler)
 	r.POST(RevocationPath, h.RevocationHandler)
+	r.GET(WellKnownPath, h.WellKnownHandler)
+}
+
+// swagger:route GET /.well-known/openid-configuration oauth2 WellKnownHandler
```
arekkas: Could you add the tag openid-connect here? :)

docs/api.swagger.yaml
Grillz: @arekkas I'm not sure what was going on with the swagger changes here, I just ran the `gen-swagger.sh` and ran it through the conversion. I was hoping you knew lol

Grillz:
No worries @arekkas I'm happy to get it right 👍 , I updated all the changes you listed and ran `goimports` and `gofmt` on all the files. Turns out the odd swagger stuff was due to a slightly older version of `go-swagger` I was running. There were still a couple of pieces that didn't generate right that I manually fixed. 

https://github.com/ory/hydra/blob/master/oauth2/handler.go#L207 is producing `'302': {}` where you said it needed a description to parse right. I manually added that, but we may want to look at having a custom response there. Also the `scopes` piece of the `securityDefinitions` I had to manually add to the swagger spec.

Let me know if you want to find a more automated solution to those in this PR. Thanks!

arekkas:
Awesome! Thanks!

PR #438, opened 2017-04-28
vendor: resolve issues with glide lock file

arekkas:


glide.yaml
```diff
@@ -1,26 +1,30 @@
 package: github.com/ory-am/hydra
 import:
 - package: github.com/Sirupsen/logrus
-  version: ~0.11.0
+  version: v0.11.5
```
janekolszak: I think it should be 0.11.5

arekkas: doesn't matter :)

PR #445, opened 2017-05-02
0.8.0: Towards production friendliness

arekkas:
This PR tackles various things.

## Breaking Changes

* [x] FindPoliciesForSubject was removed -> `/policies` no longer supports subject query parameter and instead offers pagination (offset (default 0), limit (default 500)) and is able to fetch all policies
* [x] RethinkDB + Redis removed
* [x] Upgrade ladon 0.5.0 -> 0.6.0 - requires database changes
* [x] Go 1.8 required
* [x] Move to `ory` namespace
* [x] Log format is now JSON
* [x] Policy for listing policies is now `action=list`, no longer `action=find`

## Changes

* [x] Move to ory/herodot
* [x] Migrations need to be done by hand now
* [x] Moved everything to one logger instance
* [x] sql connections managable through DSN, introduced `max_conns`, `max_idle_conns` and `max_conn_lifetime`
* [x] Introduce LOG_FORMAT which supports `json` and none (text) for now.
* [x] swagger
* [x] creating a JWK previously required keys `alg` and `id`. `id` got renamed to `kid` now.
* [x] explain offline scope somewhere

## To be done before merge

* [x] Code review
* [x] Test docker compose example extensively
* [x] Apply migrations to an existing hydra database
* [x] Test if policies work as expected in create scenarios


arekkas:
Ok, tests are now passing locally and all changes I wanted to address are now in. I'll give myself one or two days away from this issue and then review it and look for things that don't seem right.

PR #447, opened 2017-05-03
glide.yaml: move dockertest and testify to testImport

hlian:
A small patch! I noticed these two packages got pulled in when I was including it as a dependency.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -49,11 +49,6 @@ import:
   version: v1.1.0
   subpackages:
   - json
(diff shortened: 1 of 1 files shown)


hlian:
@arekkas

arekkas:
Thanks, not sure what's up with glide here. I'll probably rebase this on top of the 0.8.0 release #445

arekkas:
@hlian thank you for this, I'm closing it because I pushed this change in #445 :)

PR #448, opened 2017-05-04
Allow setting SkipTLSVerify Option value

faxal:
Allow setting the value of TLS verification option to given value. This allows one to set value from another source such as a configuration file, which is not known ahead of time.

Variadic parameters are there for backwards compatibility reasons, so that `SkipTLSVerify()` alone still works.


arekkas:
Thanks!

PR #454, opened 2017-05-09
Add Key Id to Header

Grillz:
This PR adds Key ID to the JWT header of the ID token.

Signed-off-by: pbarker <pbarker@datapipe.com>

### DIFF
--- a/oauth2/handler.go
+++ b/oauth2/handler.go
@@ -250,6 +250,9 @@ func (h *Handler) AuthHandler(w http.ResponseWriter, r *http.Request, _ httprout
 		h.writeAuthorizeError(w, authorizeRequest, errors.Wrapf(fosite.ErrServerError, "Could not store session cookie: %s", err))
 		return
 	}
+	
+	// add key id to header
+	session.Headers.Add("kid", "public")
 
 	// done
 	response, err := h.OAuth2.NewAuthorizeResponse(ctx, r, authorizeRequest, session)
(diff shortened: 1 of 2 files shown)


oauth2/handler.go
```diff
@@ -250,6 +250,9 @@ func (h *Handler) AuthHandler(w http.ResponseWriter, r *http.Request, _ httprout
 		h.writeAuthorizeError(w, authorizeRequest, errors.Wrapf(fosite.ErrServerError, "Could not store session cookie: %s", err))
 		return
 	}
+	
+	// add key id to header
+	session.Headers.Add("kid", "public")
```
Grillz: @arekkas I know you had mentioned putting this here https://github.com/ory/hydra/blob/master/oauth2/session.go#L21 but fosite appears to not be passing the session headers though entirely. 

Even if the token headers are populated in the session, https://github.com/ory/hydra/blob/master/oauth2/handler.go#L169 appears to strip them. If we want a different solve we will likely need to move into fosite.

arekkas: Adding the kid [here](https://github.com/Grillz/hydra/blob/e868ef9e860003e8035bc40da5d02c6fda26fe90/oauth2/consent_strategy.go#L93) should work

Grillz:
@arekkas I think we are good here now 👍  Thanks!

arekkas:
Nice, looks much better now :)

PR #473, opened 2017-05-25
date//userconfig: connect to cleaned DSN

### LINKED ISSUE #464: max_conns and max_conn_lifetime breaks db.Ping
```
$ export DATABASE_URL="postgres://*:*@*:5432/auth?max_conn_lifetime=1m&max_conns=20&max_idle_conns=4"
$ ./hydra host --dangerous-force-http                                                                                                                                   
INFO[0000] Connecting with postgres://*:*@*:5432/auth?max_conn_lifetime=1m&max_conns=20&max_idle_conns=4 
ERRO[0000] An error occurred                             error="Could not connect to SQL: pq: unrecognized configuration parameter "max_idle_conns""
```
Investigating I discovered that it's this call that fails: https://github.com/ory/hydra/blob/master/config/backend_connections.go#L41

Unfortunately the custom parameters hydra uses to control the number of connections appear to break the connection since they're not recognized by the database library.

Not sure if this happens with all postgres databases. I tried with our production one and a docker one summoned with 

```
docker run --rm --name postgres -e POSTGRES_PASSWORD=password -p 5432:5432 postgres -d 5
```

salvajito8:
Closes #464

### DIFF
--- a/config/backend_connections.go
+++ b/config/backend_connections.go
@@ -46,12 +46,12 @@ func (c *SQLConnection) GetDatabase() *sqlx.DB {
 
 	if err = pkg.Retry(c.L, time.Second*15, time.Minute*2, func() error {
 		c.L.Infof("Connecting with %s", c.URL.Scheme+"://*:*@"+c.URL.Host+c.URL.Path+"?"+clean.RawQuery)
-		u := c.URL.String()
-		if c.URL.Scheme == "mysql" {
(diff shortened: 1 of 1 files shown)


arekkas:
This is weird, why is this a PR with my commit that I already merged into master? https://github.com/ory/hydra/pull/470

PR #479, opened 2017-06-02
Use correct logrus path

ionutzp:


arekkas:
This breaks the build, I can't merge it but will investigate

Issue #352, opened 2017-01-05
Hydra URLs mounted to a subpath


waynerobinson:
Allow Hydra URLs to be mounted to a subpath so that instead of URLs like:

`/oauth2/auth`
`/oauth2/token`

It will respond to URLs like:

`/some/other/path/oauth2/auth`
`/some/other/path/oauth2/token`

This makes proxying from a single domain easier as no URL rewriting is necessary.

URL rewriting can effect the `redir` portion of a challenge token being set incorrectly and issues with the session cookie used to prevent replay attacks.

Easiest implementation option is probably to allow for the prefix to be set via a global config option.

janekolszak:
I think the easiest is to have a separate subdomain for hydra.

waynerobinson:
That doesn't work in all cases. For example, our app has self-service custom branding. To avoid users ever seeing our internal Hydra domain flash, it and each of our services need to exist on the customer's own (single) domain.

arekkas:
Most API gateways can solve this, in the example of kong, `preserve_host` and `strip_tag` are the settings to look for.

Closing as this has been open for a while and you guys have fixed this.

wyattanderson:
We're generally able to run Hydra behind a proxy that strips out the prefix, but we're running into the issue mentioned here:

> URL rewriting can effect the redir portion of a challenge token being set incorrectly and issues with the session cookie used to prevent replay attacks.

The `openid-configuration` response includes the right `authorization_endpoint` field by concatenating `ISSUER` (`https://foo.bar.com/hydra`) with the URL (see [the source](https://github.com/ory/hydra/blob/48c229b62af56ab16f26e827b221b9e04bb0c077/oauth2/handler.go#L139)), but the `redirectToConsent` function just [uses the incoming request](https://github.com/ory/hydra/blob/48c229b62af56ab16f26e827b221b9e04bb0c077/oauth2/handler.go#L376-L379), which isn't accurate since we strip `/hydra/` from the URL at the load balancer.

Would it make sense to update `redirectToConsent` to construct the redirect URL from the configured issuer URL and the `AuthPath`?

arekkas:
Yeah that makes sense, I would also add a check whether or not the endpoint is https or not (disallow https). Would you mind doing a PR?

wyattanderson:
I'll work on a PR.

> I would also add a check whether or not the endpoint is https or not (disallow https)

[...]

PR #509, opened 2017-06-12
oauth2: use issuer-prefixed auth URL in challenge redirect

wyattanderson:
In order to support running Hydra with a different path prefix behind
a proxy, issue a challenge token with an issuer-prefixed auth redirect
URL instead of the URL received with the auth request.

Based on discussion in #352

### DIFF
--- a/cmd/host.go
+++ b/cmd/host.go
@@ -77,7 +77,7 @@ OAUTH2 CONTROLS
 	Example: CONSENT_URL=https://id.myapp.com/consent
 
 - ISSUER: The issuer is used for identification in all OAuth2 tokens. Should be the public url of the server.
-	Defaults to ISSUER=hydra.localhost
+	Defaults to ISSUER=http://localhost:4444
 
 - AUTH_CODE_LIFESPAN: Lifespan of OAuth2 authorize codes. Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
 	Defaults to AUTH_CODE_LIFESPAN=10m
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -110,7 +110,7 @@ func initConfig() {
 	viper.SetDefault("PORT", 4444)
 
 	viper.BindEnv("ISSUER")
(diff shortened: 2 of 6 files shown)


arekkas:
Hey thank you for your contribution! Could we maybe add a fallback in case the `Issuer` is not set?

wyattanderson:
> Could we maybe add a fallback in case the Issuer is not set?

Sure, but what are valid uses of Hydra with an *empty* issuer (versus, say, the default of `hydra.localhost`)?

arekkas:
If you just want to play with it and run it locally, or as part of the tests, such a fallback makes sense. It should be easy to get started with Hydra (it's hard enough already ;) ) so let us make this as easy as possible!

wyattanderson:
100% agree; the `ISSUER` config option has a default value when not set, though, so this still works. I'd suggest (and was considering sending in a separate PR) changing the issuer logic:

When `ISSUER` is not set, construct an `ISSUER` URL from `HOST` (default `localhost`), `PORT` (default `4444`), and `--dangerous-force-http` to set the scheme. The current default of `hydra.localhost` seems out-of-spec, since it doesn't have a scheme:

> An Issuer Identifier is a case sensitive URL using the https scheme that contains scheme, host, and optionally, port number and path components and no query or fragment components.

arekkas:
HOST is unfortunately often not the real hostname, but rather the interface that hydra is listening on, this is why I used r.RequestURI in the first place here :)

wyattanderson:
Right, so, when starting out or testing locally it would work (because `HOST` defaults to localhost), and in production you'd probably be setting the `ISSUER` explicitly (`ISSUER` would not be constructed from `HOST`/`PORT` etc. when explicitly set), which I think is worth encouraging. This is roughly how Dex behaves.

arekkas:
Host defaults to empty catching all interfaces which is why hydra works per default in a docker container

wyattanderson:
🤦‍♂️  sorry; was reading the documentation wrong. I'll add the fallback.

arekkas:
Thank you!

PR #534, opened 2017-06-19
[DOC] Update "Build from source" section to actual state

dolbik:
"Build from source" section in https://ory.gitbooks.io/hydra/content/install.html#installing-hydra is different than https://github.com/ory/hydra/blob/master/README.md


arekkas:
Thank you!

PR #535, opened 2017-06-21
[Work in Progress] Cleanse Handler

matteosuppo:
A new handler that allows to delete old and useless tokens.

It's not finished by I needed to know if I'm going in the right direction.

### DIFF
--- a/oauth2/fosite_store_memory.go
+++ b/oauth2/fosite_store_memory.go
@@ -1,11 +1,12 @@
 package oauth2
 
 import (
-	"log"
 	"sync"
+	"time"
 
 	"context"
 
+	"github.com/fluxio/multierror"
 	"github.com/ory/fosite"
 	"github.com/ory/hydra/client"
 	"github.com/pkg/errors"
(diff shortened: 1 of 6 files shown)


oauth2/fosite_store_memory.go
```diff
@@ -182,3 +184,17 @@ func (s *FositeMemoryStore) RevokeAccessToken(ctx context.Context, id string) er
 	}
 	return nil
 }
+
+func (s *FositeMemoryStore) CleanseTokens(ctx context.Context, before time.Time) error {
+	errs := multierror.Accumulator{}
+
+	for sig, token := range s.AccessTokens {
+		if token.GetRequestedAt().Before(before) {
+			errs.Push(s.DeleteAccessTokenSession(ctx, sig))
```
arekkas: Is the idea here that you don't want to exit on error but rather process everything and then return the first error?

matteosuppo: Return all the errors. The accumulator returns an error containing a list of errors

arekkas:
you need to add the new dependency using `glide update`

arekkas:
If you ever come back to this issue feel free to comment & reopen

PR #605, opened 2017-10-09
fix go-jose imports

joshuarubin:
This is a patch against v0.9.13, but there was no branch based on that at present, so I apologize for requesting this against master.

When using the hydra sdk in other go packages, there is a problem with the go-jose import that makes it impossible, unless the other package uses glide, to resolve the vendor issues such that hydra still builds and tests successfully while the package that imports it builds and tests successfully. This is because it uses the `github.com/square/go-jose` import while expecting the v1 version of go-jose which is supposed to be imported with `gopkg.in/square/go-jose.v1`.

This patch fixes the imports so that go-jose v2 (which uses the `github.com/square/go-jose` import`) can be used instead and drops the v1 requirement from glide.

### DIFF
--- a/cmd/server/helper_cert.go
+++ b/cmd/server/helper_cert.go
@@ -84,8 +84,8 @@ func getOrCreateTLSCertificate(cmd *cobra.Command, c *config.Config) tls.Certifi
 
 		private := jwk.First(keys.Key("private"))
 		private.Certificates = []*x509.Certificate{cert}
-		keys = &jose.JsonWebKeySet{
-			Keys: []jose.JsonWebKey{
+		keys = &jose.JSONWebKeySet{
+			Keys: []jose.JSONWebKey{
 				*private,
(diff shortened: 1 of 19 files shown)


arekkas:
I created a new branch for 0.9.x: https://github.com/ory/hydra/tree/v0.9.x - can you rebase against that?

Also note that this might be an issue with fosite as well, go-jose is used there as well (version 1.0) so resolving this only here won't solve it properly (I think)

joshuarubin:
I haven't had an issue with fosite so far, but yes it can't hurt to add a Gopkg.toml there either.

I've rebased this as asked.

arekkas:
This can be closed right?

joshuarubin:
yes, it's included in #606

arekkas:
I made this change to master as well (and switched to dep) which will be released with #630

PR #606, opened 2017-10-09
Support dep

joshuarubin:
This is a patch against v0.9.13, and requires #605, but there was no branch based on that at present, so I apologize for requesting this against master.

This adds a Gopkg.toml file so that [dep](https://github.com/golang/dep) can be supported. You can continue to use glide if you keep the Gopkg.toml up-to-date too. The reason that this is useful, for packages that use the hydra sdk, is that dep will recursively resolve dependencies using the Gopkg.toml file.

Right now, we have to add overrides for each of hydra's dependencies which is less than ideal.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,8 @@ install:
   - go install github.com/ory/hydra
   - glide update
   - go install github.com/ory/hydra
+  - dep ensure
+  - go install github.com/ory/hydra
 
 script:
   - touch ./coverage.tmp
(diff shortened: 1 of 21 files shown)


arekkas:
I see, use of dep makes sense, the idea is to transition at some point anyhow. However, I had some issues with dep on windows which is why it's not included yet. I'm however open to accepting this change.

One thing though, could you add dep tests to travis.yml (similar to glide tests)?

Also, here's a 0.9.x branch: https://github.com/ory/hydra/tree/v0.9.x

joshuarubin:
OK, based on v0.9.x and all glide and dep tests now pass

arekkas:
Thank you!

Issue #594, opened 2017-10-03
Feature Request: ability to list all groups


joshuarubin:
We are building a management UI around hydra and would like to be able to list all groups in a single API call. Right now we build a list dynamically from the list of clients:
```
for client in range clientList {
    GET /warden/groups?member=client.ID
}
```
and policies:
```
for policy in range policyList {
    for subject in range policy.Subjects {
       GET /warden/groups?member=subject
    }
}
```

This is obviously very inefficient and even wasteful because policy subjects are not necessarily only groups.

This also has the possibility of missing groups because we can only find groups that have a known client id as a member or have a policy listing the group as a subject. We understand that the groups we might miss, orphan groups, necessarily can't have any affect on the system, but it would still be nice to find them. Additionally, it is challenging if we create a group with no members, and then intend to update/create a policy referencing it, the group doesn't show up as being available (e.g. for autocompletion) until after the policy is updated/created.

Thank you very much!

arekkas:
Yeah that makes totally sense!

arekkas:
If you want, you can take a shot at it yourself.

The handler is [here](https://github.com/ory/hydra/blob/master/warden/group/handler.go#L73-L92). You could probably build a check [after here](https://github.com/ory/hydra/blob/master/warden/group/handler.go#L75) to see if `members == ""`. If so, you could run a manager function that implements `ListGroups(limit, offset)`. Then you would have to implement that function in the managers:

* [sql](https://github.com/ory/hydra/blob/master/warden/group/manager_sql.go)
* [memory](https://github.com/ory/hydra/blob/master/warden/group/manager_memory.go)
* [http](https://github.com/ory/hydra/blob/master/warden/group/manager_http.go)

And of course add some test [somewhere around here](https://github.com/ory/hydra/blob/master/warden/group/manager_test_helper.go#L36) where you check if the groups in the manager match the expected amount of groups.

If you don't feel confident with Go you could either ask on Gitter or let me know and I'll try to address it soon (no time guarantees though, sorry).

PR #612, opened 2017-10-17
add ability to list all groups

joshuarubin:
This is to satisfy #594 

I'm not sure if the `api.swagger.yaml` file is complete/correct since the `gen-swagger.sh` script produces the uncommitted `api.swagger.json` file.

The API is also a little strange as the `GET /warden/groups` endpoint already existed and previously required a `member` param (and `GET /warden/groups/<id>` refers to the group id, not the member id).

Now `GET /warden/groups` _OPTIONALLY_ uses the `member` param. If it exists, the behavior is the same as before so there are no API regressions.

However, if `member` does not exist, then it is assumed that the user wants to list all groups. In this case, the required policy resource changes (from `rn:hydra:warden:groups:<member>` to `rn:hydra:warden:groups`). Additionally, the `limit` and `offset` parameters are used if they exist. `limit` defaults to `500` if not specified (the same as `GET /policies`).

### DIFF
--- a/docs/api.swagger.yaml
+++ b/docs/api.swagger.yaml
@@ -1376,8 +1376,8 @@ paths:
   /warden/groups:
     get:
       description: >-
-        The subject making the request needs to be assigned to a policy
-        containing:
+        The subject making the request, if member is specified, needs to be
+        assigned to a policy containing:
 
 
         ```
@@ -1385,6 +1385,23 @@ paths:
         {
 
         "resources": ["rn:hydra:warden:groups:<member>"],
+
+        "actions": ["get"],
+
+        "effect": "allow"
+
+        }
+
+        ```
+
+        If member is not specified, the subject making the request needs to be
+        assigned to a policy containing:
+
+        ```
+
+        {
+
+        "resources": ["rn:hydra:warden:groups"],
 
         "actions": ["get"],
 
@@ -1403,8 +1420,8 @@ paths:
       tags:
         - warden
         - groups
-      summary: Find group IDs by member
-      operationId: findGroupsByMember
+      summary: Find group IDs
+      operationId: findGroups
       security:
         - oauth2:
             - hydra.groups
@@ -1415,9 +1432,21 @@ paths:
           description: The id of the member to look up.
           name: member
           in: query
+        - type: integer
+          format: int64
+          x-go-name: Offset
+          description: The offset from where to start looking if member isn't specified.
+          name: offset
+          in: query
+        - type: integer
+          format: int64
+          x-go-name: Limit
+          description: The maximum amount of policies returned if member isn't specified.
+          name: limit
+          in: query
       responses:
         '200':
-          $ref: '#/responses/findGroupsByMemberResponse'
+          $ref: '#/responses/findGroupsResponse'
         '401':
           $ref: '#/responses/genericError'
         '403':
@@ -2721,7 +2750,7 @@ responses:
     headers:
       description:
         type: string
-  findGroupsByMemberResponse:
+  findGroupsResponse:
     description: A list of groups the member is belonging to
     schema:
       type: array
--- a/docs/api.swagger.yaml
+++ b/docs/api.swagger.yaml
@@ -1376,26 +1376,9 @@ paths:
   /warden/groups:
     get:
       description: >-
-        The subject making the request, if member is specified, needs to be
-        assigned to a policy containing:
+        The subject making the request needs to be assigned to a policy
+        containing:
 
-
-        ```
-
-        {
-
-        "resources": ["rn:hydra:warden:groups:<member>"],
-
-        "actions": ["get"],
-
-        "effect": "allow"
-
-        }
-
-        ```
-
-        If member is not specified, the subject making the request needs to be
-        assigned to a policy containing:
 
         ```
 
@@ -1420,8 +1403,8 @@ paths:
       tags:
         - warden
         - groups
-      summary: Find group IDs
-      operationId: findGroups
+      summary: List group IDs
+      operationId: listGroups
       security:
         - oauth2:
             - hydra.groups
@@ -1446,7 +1429,7 @@ paths:
           in: query
       responses:
         '200':
-          $ref: '#/responses/findGroupsResponse'
+          $ref: '#/responses/listGroupsResponse'
         '401':
           $ref: '#/responses/genericError'
         '403':
@@ -2750,7 +2733,7 @@ responses:
     headers:
       description:
         type: string
-  findGroupsResponse:
+  listGroupsResponse:
     description: A list of groups the member is belonging to
     schema:
       type: array
--- a/warden/group/handler.go
+++ b/warden/group/handler.go
@@ -41,21 +41,11 @@ func (h *Handler) SetRoutes(r *httprouter.Router) {
 	r.DELETE(GroupsHandlerPath+"/:id/members", h.RemoveGroupMembers)
 }
 
-// swagger:route GET /warden/groups warden groups findGroups
+// swagger:route GET /warden/groups warden groups listGroups
 //
-// Find group IDs
+// List group IDs
 //
-// The subject making the request, if member is specified, needs to be assigned to a policy containing:
-//
-//  ```
-//  {
-//    "resources": ["rn:hydra:warden:groups:<member>"],
-//    "actions": ["get"],
-//    "effect": "allow"
-//  }
-//  ```
-//
-// If member is not specified, the subject making the request needs to be assigned to a policy containing:
+// The subject making the request needs to be assigned to a policy containing:
 //
 //  ```
 //  {
@@ -77,7 +67,7 @@ func (h *Handler) SetRoutes(r *httprouter.Router) {
 //       oauth2: hydra.groups
 //
 //     Responses:
-//       200: findGroupsResponse
+//       200: listGroupsResponse
 //       401: genericError
 //       403: genericError
 //       500: genericError
@@ -107,20 +97,16 @@ func (h *Handler) FindGroupNames(w http.ResponseWriter, r *http.Request, _ httpr
 		return
 	}
 
-	limit, err := intFromQuery(r, "limit")
+	limit, err := intFromQuery(r, "limit", 500)
 	if err != nil {
 		h.H.WriteError(w, r, errors.WithStack(err))
 		return
 	}
 
-	offset, err := intFromQuery(r, "offset")
+	offset, err := intFromQuery(r, "offset", 0)
 	if err != nil {
 		h.H.WriteError(w, r, errors.WithStack(err))
 		return
-	}
-
-	if limit == 0 {
-		limit = 500
 	}
 
 	g, err := h.Manager.ListGroups(limit, offset)
@@ -133,10 +119,10 @@ func (h *Handler) FindGroupNames(w http.ResponseWriter, r *http.Request, _ httpr
 	h.H.Write(w, r, g)
 }
 
-func intFromQuery(r *http.Request, key string) (int64, error) {
+func intFromQuery(r *http.Request, key string, def int64) (int64, error) {
 	val := r.URL.Query().Get(key)
 	if val == "" {
-		val = "0"
+		return def, nil
 	}
 
 	return strconv.ParseInt(val, 10, 64)
--- a/warden/group/handler.go
+++ b/warden/group/handler.go
@@ -90,10 +90,6 @@ func (h *Handler) FindGroupNames(w http.ResponseWriter, r *http.Request, _ httpr
 		Action:   "get",
 	}
 
-	if member != "" {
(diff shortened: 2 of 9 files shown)


arekkas:
Hey @joshuarubin the way you solved it is perfect. It is common to have a resource endpoints `/resources` which lists everything, and then query options like `/resources?owner=1234` to request a subset (similar to limit & offset).

The only thing I would like to change is to remove `rn:hydra:warden:groups:<member>` and only do `rn:hydra:warden:groups` AC.

I will now review your code changes and leave comments where they fit.

docs/api.swagger.yaml
```diff
@@ -1376,8 +1376,8 @@ paths:
   /warden/groups:
     get:
       description: >-
-        The subject making the request needs to be assigned to a policy
-        containing:
+        The subject making the request, if member is specified, needs to be
```
arekkas: As discussed in my comment, I'd like to revert this and replace this with a policy check against `rn:hydra:warden:groups`

docs/api.swagger.yaml
```diff
@@ -1376,8 +1376,8 @@ paths:
...
+        assigned to a policy containing:
```
arekkas: As discussed in my comment, replace this with one check to `rn:hydra:warden:groups`

docs/api.swagger.yaml
```diff
@@ -1403,8 +1420,8 @@ paths:
       tags:
         - warden
         - groups
-      summary: Find group IDs by member
-      operationId: findGroupsByMember
+      summary: Find group IDs
```
arekkas: In fact, we're receiving the full group payload (no longer only their IDs). This isn't your fault, but mine since I forgot to update the header title. But since you're already at it, maybe you could change that too ... :)

warden/group/handler.go
```diff
@@ -40,11 +41,11 @@ func (h *Handler) SetRoutes(r *httprouter.Router) {
 	r.DELETE(GroupsHandlerPath+"/:id/members", h.RemoveGroupMembers)
 }
 
-// swagger:route GET /warden/groups warden groups findGroupsByMember
+// swagger:route GET /warden/groups warden groups findGroups
```
arekkas: This should actually be renamed to listGroups

warden/group/handler.go
```diff
@@ -40,11 +41,11 @@ func (h *Handler) SetRoutes(r *httprouter.Router) {
...
 //
-// Find group IDs by member
+// Find group IDs
```
arekkas: `List groups`

warden/group/handler.go
```diff
@@ -66,24 +77,65 @@ func (h *Handler) SetRoutes(r *httprouter.Router) {
 //       oauth2: hydra.groups
 //
 //     Responses:
-//       200: findGroupsByMemberResponse
+//       200: findGroupsResponse
```
arekkas: `listGroupsResponse`

warden/group/handler.go
```diff
@@ -66,24 +77,65 @@ func (h *Handler) SetRoutes(r *httprouter.Router) {
...
 	var member = r.URL.Query().Get("member")
 
-	g, err := h.Manager.FindGroupNames(member)
-	if err != nil {
+	accessReq := &firewall.TokenAccessRequest{
+		Resource: GroupsResource,
+		Action:   "get",
+	}
+
+	if member != "" {
```
arekkas: This can be removed as discussed in my comment

warden/group/handler.go
```diff
@@ -66,24 +77,65 @@ func (h *Handler) SetRoutes(r *httprouter.Router) {
...
+		if err != nil {
+			h.H.WriteError(w, r, err)
+			return
+		}
+
+		h.H.Write(w, r, g)
+		return
+	}
+
+	val := r.URL.Query().Get("offset")
```
arekkas: Can we wrap this and the parsing in it's own function? It would allow for better readability and a shorter function body, e.g.

func intFromQuery(r *http.Request, key string) (int, error)

warden/group/manager_memory.go
```diff
@@ -18,6 +19,8 @@ type MemoryManager struct {
 	sync.RWMutex
 }
 
+var _ Manager = (*MemoryManager)(nil)
```
arekkas: is this to assert that `MemoryManager` implements `Manager`?

arekkas:
Damn, halfway through the review I noticed that you made this against the 0.9.x branch. I would much rather see it done against the 0.10.0 one, is that an issue for you?

docs/api.swagger.yaml
arekkas: This applies to the 0.10.x branch - for the 0.9.x branch this is correct.

warden/group/handler.go
arekkas: For 0.9.x this should be `List group IDs`

warden/group/manager_sql.go
```diff
@@ -123,6 +125,28 @@ func (m *SQLManager) RemoveGroupMembers(group string, subjects []string) error {
 	return nil
 }
 
+func (m *SQLManager) ListGroups(limit, offset int64) ([]string, error) {
+	if limit < 0 {
```
arekkas: I think it should be fine to set limit to 0 (or 1) if it's below 0

warden/group/manager_sql.go
```diff
@@ -123,6 +125,28 @@ func (m *SQLManager) RemoveGroupMembers(group string, subjects []string) error {
...
+		return nil, errors.New("limit can't be less than 0")
+	}
+
+	if offset < 0 {
```
arekkas: I think it should be fine to set offset to 0 if it's below 0

warden/group/manager_test.go
```diff
@@ -85,6 +85,8 @@ func connectToPG() {
 }
 
 func TestManagers(t *testing.T) {
+	t.Parallel()
```
arekkas: This is nice, I totally did not know about that flag. That way, postgres and mysql will be run alongside. This should be everywhere. Thank you!

#617

warden/group/manager_sql.go
```diff
@@ -123,6 +125,28 @@ func (m *SQLManager) RemoveGroupMembers(group string, subjects []string) error {
...
+		return nil, errors.New("offset can't be less than 0")
+	}
+
+	var q []string
+
+	if limit == 0 {
+		limit = 500
+	}
+
+	if err := m.DB.Select(&q, m.DB.Rebind("SELECT id from hydra_warden_group ORDER BY id LIMIT ? OFFSET ?"), limit, offset); err != nil {
```
arekkas: I'm not sure about the order clause here, what it's goal?

joshuarubin:
We would like to be able to get this functionality ASAP which would be less risky if we can stay on 0.9.x. However, I don't think it would be too hard to cherry pick the commits and get it working on master. If we can merge this into 0.9.x, I'll set up another PR to get it into master.

warden/group/manager_memory.go
```diff
@@ -76,6 +79,39 @@ func (m *MemoryManager) RemoveGroupMembers(group string, subjects []string) erro
...
+	if offset >= int64(len(m.Groups)) {
+		return nil, nil
+	}
+
+	var res []string
+	for _, g := range m.Groups {
+		res = append(res, g.ID)
+	}
+
+	sort.Strings(res)
```
arekkas: Sorting makes sense here because maps don't have an order.

arekkas:
> We would like to be able to get this functionality ASAP which would be less risky if we can stay on 0.9.x. However, I don't think it would be too hard to cherry pick the commits and get it working on master. If we can merge this into 0.9.x, I'll set up another PR to get it into master.

I would love to see that. If you find the time, I'd also love to see that for #605 :)

warden/group/handler.go
joshuarubin: Sure, I didn't see many (any?) separate functions for these things, so I was just trying to maintain the existing style.

warden/group/manager_memory.go
joshuarubin: Yes. It will throw a compile time error if the interface isn't satisfied.

warden/group/manager_sql.go
joshuarubin: OK

warden/group/manager_sql.go
joshuarubin: OK

warden/group/handler.go
arekkas: No worries, most of them are encapsulated in their own libraries (e.g. the whole json/statuscode/error stuff) and thus a kind of invisible to the naked eye :)

warden/group/manager_memory.go
arekkas: Since it's used somewhere in the code, that should happen anyways, but it probably makes sense when testing only this specific package/file - maybe move this line to the tests?

warden/group/manager_sql.go
joshuarubin: ORDER BY is required to ensure that the paging is consistent.
https://www.postgresql.org/docs/8.2/static/queries-limit.html

arekkas: Oh, I did not know that!

warden/group/manager_memory.go
joshuarubin: Your call. I like putting it right after type definitions so I can see how it is intended to be used. Also, I think it was manager_http that didn't fail the build at all if it didn't satisfy Manager. However, the tests failed to build. This didn't feel right to me. Whatever you want though.

arekkas: Makes sense, let's keep it where it is

joshuarubin:
OK. Updated with changes from review. Still don't know how to generate the api.swagger.yaml so I hope it's correct.

arekkas:
Nice! I'll go over the changes one more time. Regarding swagger, check out [this script](https://github.com/ory/hydra/blob/master/scripts/run-genswag.sh) from the 0.10.x branch

warden/group/handler.go
arekkas: ping

warden/group/handler.go
arekkas: ping

warden/group/handler.go
```diff
@@ -40,11 +41,11 @@ func (h *Handler) SetRoutes(r *httprouter.Router) {
...
 }
 
-// swagger:route GET /warden/groups warden groups findGroupsByMember
+// swagger:route GET /warden/groups warden groups findGroups
 //
-// Find group IDs by member
+// Find group IDs
 //
-// The subject making the request needs to be assigned to a policy containing:
+// The subject making the request, if member is specified, needs to be assigned to a policy containing:
```
arekkas: This line and the following policy can be removed, as discussed in the comments

warden/group/handler.go
arekkas: ping

warden/group/handler.go
```diff
@@ -66,31 +77,71 @@ func (h *Handler) SetRoutes(r *httprouter.Router) {
...
+		return
+	}
+
+	limit, err := intFromQuery(r, "limit")
+	if err != nil {
+		h.H.WriteError(w, r, errors.WithStack(err))
+		return
+	}
+
+	offset, err := intFromQuery(r, "offset")
```
arekkas: If you supply a default value, no error checking is requried and you can save the `limit == 0` check!

joshuarubin:
OK. Updated again.

warden/group/manager_sql.go
```diff
@@ -123,6 +125,24 @@ func (m *SQLManager) RemoveGroupMembers(group string, subjects []string) error {
...
+func (m *SQLManager) ListGroups(limit, offset int64) ([]string, error) {
+	if limit <= 0 {
+		limit = 500
+	}
+
+	if offset < 0 {
+		offset = 0
+	}
+
+	var q []string
```
arekkas: If no results are returned, this will be nil which in turn will return `null` in the JSON response (see https://play.golang.org/p/7H6y1t2mdd ). To avoid this please initialize the slice (q := []string{})

warden/group/manager_memory.go
```diff
@@ -76,6 +79,35 @@ func (m *MemoryManager) RemoveGroupMembers(group string, subjects []string) erro
...
+
+	if offset < 0 {
+		offset = 0
+	}
+
+	if offset >= int64(len(m.Groups)) {
+		return nil, nil
+	}
+
+	var res []string
```
arekkas: If no results are returned, this will be nil which in turn will return null in the JSON response (see https://play.golang.org/p/7H6y1t2mdd ). To avoid this please initialize the slice (q := []string{})

joshuarubin:
Nice catch. I think everything should be ready now.

arekkas:
Thank you!

arekkas:
https://github.com/ory/hydra/releases/tag/v0.9.16

PR #625, opened 2017-10-21
docs: updated hydra version in the tutorial to v0.10.0-alpha.8 and consent app to v0.10.0-alpha.9

abusaidm:
this PR updated the installation docs to reflect the latest hydra version at the time and the consent app.

also it fixes a mistake in assigning the consent scope.

I have ran through the installation with this update and was able to reach the end without any issues.


arekkas:
THank you!

PR #626, opened 2017-10-21
updated links to apiary as the old ones didn't work

abusaidm:
links for apiary didn't link to correct section of the page as the # was incorrect.


arekkas:
Thank you once again!

Issue #627, opened 2017-10-24
Support ES256 JWK Algo


joshuarubin:
It looks like support for ES256 is available (jwk/generator_ecdsa256.go), but that the JWK handler doesn't support it (https://github.com/ory/hydra/blob/master/jwk/handler.go#L27-L38).

Go's P-256 is implementation is constant-time (which prevents certain types of attacks) while its P-384 and P-521 are not (https://github.com/gtank/cryptopasta).

joshuarubin:
closed in favor of #628

arekkas:
I think we should disable the P521 generator for now, as it's not safe to use them for repeated use in private keys.

joshuarubin:
[RFC7518](https://tools.ietf.org/html/rfc7518#section-3.1) lists ES256 as "Recommended+" meaning:

> The use of "+" in the Implementation Requirements column indicates
   that the requirement strength is likely to be increased in a future
   version of the specification.

I think that the stronger algorithm should continue to be supported.

Also, the [webcrypto](https://w3c.github.io/webcrypto/Overview.html#ecdsa-operations) spec has renamed `ES521` to `ES512`: 

>  If the "alg" field is equal to the string "ES512":
Let algNamedCurve be the string "P-521".

arekkas:
Ok, let's keep them then. As long as JWK/JWT spec talk of ES521 I think we should keep it that way. Definitely doesn't help though that there are naming inconsistencies now.

joshuarubin:
Can you link to the spec which still references ES521?

arekkas:
My bad, looks like I got mixed up, it's definitely ES512

arekkas:
I closed this because ES512 is now tracked in #651

PR #628, opened 2017-10-24
Support ES256 JWK Algo

joshuarubin:
I think this is all that's necessary to support #627 

It looks like support for ES256 is available (`jwk/generator_ecdsa256.go`), but that the JWK handler doesn't support it (https://github.com/ory/hydra/blob/master/jwk/handler.go#L27-L38).

Go's P-256 is implementation is constant-time (which prevents certain types of attacks) while its P-384 and P-521 are not (https://github.com/gtank/cryptopasta).

### DIFF
--- a/cmd/keys_create.go
+++ b/cmd/keys_create.go
@@ -27,6 +27,6 @@ var keysCreateCmd = &cobra.Command{
 
 func init() {
 	keysCmd.AddCommand(keysCreateCmd)
-	keysCreateCmd.Flags().StringP("alg", "a", "", "REQUIRED name that identifies the algorithm intended for use with the key. Supports: RS256, ES521, HS256")
+	keysCreateCmd.Flags().StringP("alg", "a", "", "REQUIRED name that identifies the algorithm intended for use with the key. Supports: RS256, ES256, ES521, HS256")
 
 }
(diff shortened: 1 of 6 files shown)


joshuarubin:
I agree with you, that's why I have updated the `master` branch of our repo to be in sync with your `master` and all my recent changes. After you approved this, I planned on making that pull request. You can see it here https://github.com/zvelo/hydra

arekkas:
Ok awesome, that works for me - I'll make a code review and merge if everything checks out!

cmd/keys_create.go
```diff
@@ -27,6 +27,6 @@ var keysCreateCmd = &cobra.Command{
 
 func init() {
 	keysCmd.AddCommand(keysCreateCmd)
-	keysCreateCmd.Flags().StringP("alg", "a", "", "REQUIRED name that identifies the algorithm intended for use with the key. Supports: RS256, ES521, HS256")
+	keysCreateCmd.Flags().StringP("alg", "a", "", "REQUIRED name that identifies the algorithm intended for use with the key. Supports: RS256, ES256, ES521, HS256")
```
arekkas: Do you think we should remove ES521 completely?

joshuarubin: See my https://github.com/ory/hydra/issues/627#issuecomment-339326814

PR #642, opened 2017-11-05
(WIP) client: use serial for primary key and query surrogate id

arekkas:


client/manager_test.go
```diff
@@ -25,7 +25,7 @@ func TestMain(m *testing.M) {
 	flag.Parse()
 	if !testing.Short() {
 		connectToPG()
-		connectToMySQL()
+		// connectToMySQL()
```
arekkas: This needs to be re-added

PR #665, opened 2017-11-27
cli: Fix hydra cli adding policy subjects on subject remove

jamesnicolas:
Removed the line that added a subject back after it removed it.

I also added a couple of `\n` on fmt.Printfs that didn't have them.


arekkas:
Perfect, thank you!

Issue #661, opened 2017-11-24
swagger: scope pattern requires a space


pnicolcev-tulipretail:
The pattern definition for scope in the client requires a space after each scope. This is causing problems when generating a client that checks against this pattern.

https://github.com/ory/hydra/blob/master/client/client.go#L57

I've generated a PHP client and so far it works, but when creating a client I get errors when the scope is returned. 

For reference, the generated code:
```
     /**
      * Sets scope
      * @param string $scope Scope is a string containing a space-separated list of scope values (as described in Section 3.3 of OAuth 2.0 [RFC6749]) that the client can use when    requesting access tokens.
      * @return $this
      */
     public function setScope($scope)
     {

         if (!is_null($scope) && (!preg_match("/([a-zA-Z0-9\\.\\*]+\\s)+/", $scope))) {
             throw new \InvalidArgumentException("invalid value for $scope when calling OAuth2Client., must conform to the pattern /([a-zA-Z0-9\\.\\*]+\\s)+/.");
         }

         $this->container['scope'] = $scope;

         return $this;
     }
```

Changing it to `/([a-zA-Z0-9\\.\\*]+\\s?)+/` resolves the issue.

Am I reading this correctly? Want a PR?

arekkas:
Ah yes, because if only one scope is allowed, that won't match. Feel free to create a PR!

arekkas:
This is resolved now

PR #668, opened 2017-11-28
docs: Make space optional in scope regex (#661)

pnicolcev-tulipretail:
Signed-off-by: Philip Nicolcev <phil.nicolcev@tulip.io>

### DIFF
--- a/docs/api.swagger.json
+++ b/docs/api.swagger.json
@@ -2254,12 +2254,12 @@
           "x-go-name": "Public"
         },
         "redirect_uris": {
+          "description": "RedirectURIs is an array of allowed redirect urls for the client, for example http://mydomain/oauth/callback .",
           "type": "array",
           "items": {
             "type": "string"
           },
-          "x-go-name": "RedirectURIs",
-          "example": "http://mydomain/oauth/callback ."
+          "x-go-name": "RedirectURIs"
         },
         "response_types": {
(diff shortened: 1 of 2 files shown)


pnicolcev-tulipretail:
Running genswag caught another change in there. Can remove if that's a problem.

docs/api.swagger.json
```diff
@@ -2254,12 +2254,12 @@
           "x-go-name": "Public"
         },
         "redirect_uris": {
-          "description": "RedirectURIs is an array of allowed redirect urls for the client, for example: http://mydomain/oauth/callback .",
```
arekkas: This is weird, why is the description gone?

pnicolcev-tulipretail: I don't know, i assumed you'd removed it. Strange that's the only thing that changed. I tried running genswag again, same result. I'll investigate what I did wrong.

arekkas: Hm strange, maybe a version mismatch in go-swagger. The corresponding line is here: https://github.com/ory/hydra/blob/master/client/client.go#L39-L40

I think the issue is `, for example: http://mydomain/oauth/callback .` which is apparently parsed as an example section. You could try rewriting it to `, for example http://mydomain/oauth/callback`

arekkas: Damn, I was just a bit late! :D

pnicolcev-tulipretail: @arekkas fixed

arekkas:
Thank you!

PR #700, opened 2017-12-11
feat(default root client): enable overwriting redirect uris for the r…

### LINKED ISSUE #699: cmd: Add cli helper for importing and exporting environments (clients, policies, keys)
It makes developing against hydra a bit difficult if the redirect uri for the root client is hard coded to localhost and you're trying emulate the production architecture a bit better. With for example a reverse proxy routing all traffic to the correct endpoints using urls. By allowing to overwrite the redirect uris in the same way as the root clients' credentials it would make hydra flexible while developing against it.

Rio:
…oot client

Just like `FORCE_ROOT_CLIENT_CREDENTIALS`, `FORCE_ROOT_CLIENT_REDIRECT_URI` will set the redirect uris for the root client.

closes #699

### DIFF
--- a/cmd/host.go
+++ b/cmd/host.go
@@ -65,6 +65,12 @@ CORE CONTROLS
 	and the secret: "FORCE_ROOT_CLIENT_CREDENTIALS=urlencode(id):urlencode(secret)".
 	Example: FORCE_ROOT_CLIENT_CREDENTIALS=admin:h6hy92tK4dQcZ2EaFsGNRtqg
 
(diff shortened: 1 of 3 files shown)


arekkas:
Thank you for your contribution. As explained in https://github.com/ory/hydra/issues/699#issuecomment-350829516 I don't feel this is a smart move. It puts too much emphasis in a client that should only be used for management.

Closing, but feel free to persuade me :)

Issue #304, opened 2016-11-08
oauth2: allow token revocation without knowing the token (i.e. per user)


joelpickup:
We would love a mechanism for the revocation of all tokens associated with a specific user. This would allow us to suspend user accounts from our admin panel.

arekkas:
if you have any proposals or ideas how to achieve that feel free to share them here!

arekkas:
I'm still not sure how to resolve this properly. Maybe only for priviledged clients?

20zinnm:
How about having some memory-based messaging channel or whatnot between the auth server and the resource server, such that, if the auth server revokes access for a user, it sends out a message to resource servers to reject tokens issued for a user from before the message came out? (I.e. some RabbitMQ deal?)

Otherwise, I'd handle this the OAuth2 way--issue short-lived access tokens and longer refresh tokens.

arekkas:
Hm a message queue would add another depedency (what if you don't want to use rabbitmq). The most obvious way is to write a small SQL script that scans the subject and audience from the table and deletes the rows.

I think there is no way around a dedicated endpoint that is protected, similar like the JWK endpoint.

joshuarubin:
I'm just curious if a solution to this issue is required to do the following:

1. End user is taken through the user auth flow and approves the `RequestedScopes` for a single "app" (ClientID) which include the `offline` scope. User may never have access to the token or refresh_token.
2. User later would like to revoke their approval for that particular "app".

As I understand it, this would only be possible if the "app" provides a mechanism for the user to do this since the app (which has access to the tokens) can revoke the token and refresh_token with the `/oauth2/revoke` endpoint.

We, however, would like to have a single interface for the user to see all the applications they have approved and have an opportunity to revoke access for any particular app.

We keep track of the applications (`claims.Audience`) that the user has approved by hooking into the consent app. So listing the applications for the user is not an issue. We just don't appear to have any way to revoke the refresh_token for that application/user without having it in hand.

caseyf:
Could this be added as a REST API that requires `rn:hydra:whatever` permissions? 

My use case is the usual: 
* users need to be able to revoke permission that they have granted from a web dashboard outside of any client apps
* staff needs to be able to revoke all tokens for a specific client + user combination

arekkas:
Yes, this should in fact be a REST API in Hydra which is protected by ACL!

caseyf:
[...]

PR #720, opened 2017-12-16
oauth2: Adds ability to detect previous consent

### LINKED ISSUE #692: oauth2: Require consent for OAuth 2.0 public clients
This is currently up to the implementation of the consent app but should be baked into the consent protocol. See https://tools.ietf.org/html/rfc6819#section-5.2.3.2

### LINKED ISSUE #697: oauth2: Remember authentication and application authorization
The goal of this issue is to make implementing an OIDC conformant consent app easier.

* [ ] Hydra should remember user authentication based on cookies
* [ ] Hydra should remember previously authorized applications and their scopes. If a subset is used, the consent screen can be skipped.
* [ ] Hydra should listen to `maxAge`, `prompt=login`, `prompt=none` and `prompt=consent` and `acr` accordingly. `acr` should be the same as `prompt=login`
* [ ] Hydra should inform the consent app what steps have to be taken:
  * [ ] `mustAuthenticateUser` indicates that the user must be logged out (if signed in already) and logged back in again. This should happen when `prompt=login` or when no user session exists
  * [ ] `mustAuthorizeConsent` indicates that the user must give his/her consent. This should happen when `prompt=consent`, when an unknown application requests authorization, or when new scopes are requested
  * [ ] `requestedAuthenticationContext` (*Authentication Context Class Reference*) indicates that some level of assurance was required. Setting this to anything other than empty or `0` has the same effect as `prompt=login`
* [ ] When user session is revoked, require re-authentication and re-authorization

For starters, subject ID and authorized clients (+ scopes) could be stored in the encrypted session cookie, or alternatively in the consent manager.

arekkas:
This commit aims at improving OpenID Connect conformity whilst making it
as easy as possible to implement the consent app.

For that reason, ORY Hydra is now capable of remembering user sessions
and previous consent requests, and properly handles OpenID Connect's
maxAge and prompt parameters.

Additionally, public OAuth 2.0 clients always require the full consent
flow.

* [x] Closes #692
* [x] Closes #697
* [ ] Consent session sign out
* [ ] #304
* [ ] Check that, if user id from cookie mismatches with user id from consent request, either an error is thrown or the user from the consent request is used


arekkas:
There are multiple things which need to be addressed:

* [ ] Being able to revoke a subjects's (user + oauth2 client) all access + refresh tokens
* [ ] Being able to revoke access of a specific application (oa2 client) for a user 
* [ ] Being able to destroy a user's session with regards to the consent flow

Let's take a closer look at the options.

### Destroying previous consent session

I think this one's pretty easy. We could add something like `DELETE /consent/requests?subject=<subject>[&client=<client>]` which would delete all previous consent sessions, thus requiring the user to re-authenticate.

This could probably also revoke the access & refresh tokens that were issued with that request.

### User logout

This would only destroy the user cookie but not revoke any tokens. Maybe along the lines of `/oauth2/destroy-session`?

arekkas:
This has been solved by another PR

PR #765, opened 2018-02-06
client: do not show/send secret when client is public

### LINKED ISSUE #737: Do not show client secret when client is public in CLI

zepatrik:
Previously a newly created public client had a secret send with the initial response and this secret was displayed in the CLI.
Now it is clear that there is no secret needed for public clients. It is not displayed in the CLI anymore.

closes #737

### DIFF
--- a/client/sdk_test.go
+++ b/client/sdk_test.go
@@ -71,7 +71,7 @@ func TestClientSDK(t *testing.T) {
 	c := hydra.NewOAuth2ApiWithBasePath(server.URL)
 	c.Configuration.Transport = httpClient.Transport
 
-	t.Run("client is created and updated", func(t *testing.T) {
+	t.Run("case=client is created and updated", func(t *testing.T) {
 		createClient := createTestClient("")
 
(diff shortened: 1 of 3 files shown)


client/sdk_test.go
```diff
@@ -71,7 +71,7 @@ func TestClientSDK(t *testing.T) {
 	c := hydra.NewOAuth2ApiWithBasePath(server.URL)
 	c.Configuration.Transport = httpClient.Transport
 
-	t.Run("foo", func(t *testing.T) {
+	t.Run("client is created and updated", func(t *testing.T) {
```
arekkas: Add `case=client is ...` here

client/sdk_test.go
```diff
@@ -106,4 +106,18 @@ func TestClientSDK(t *testing.T) {
 		require.NoError(t, err)
 		assert.Equal(t, http.StatusNotFound, response.StatusCode)
 	})
+
+	t.Run("public client is transmitted without secret", func(t *testing.T) {
```
arekkas: Add `case=public client is ...` here

PR #767, opened 2018-02-06
docs: Fix dead link to example policy

gr-eg:
The policy linked to as an example has since been removed. Just point to
a different policy instead.


arekkas:
Thank you!

PR #768, opened 2018-02-07
warden/groups: add a PUT method to overwrite members & some refactoring

### LINKED ISSUE #745: Add PUT method for /warden/groups/:id
Add a `PUT` method to overwrite the whole members array.

zepatrik:
There was no way to overwrite all group members meaning just sending the actual value, instead one had to keep track of changes or remove all members to add the new ones again.
Now a http PUT method allows to set the members array.

closes #745


zepatrik:
not quite ready to merge, just for review if I am doing it right

warden/group/handler.go
```diff
@@ -172,19 +168,14 @@ func (h *Handler) FindGroupNames(w http.ResponseWriter, r *http.Request, member
 //       403: genericError
 //       500: genericError
 func (h *Handler) CreateGroup(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
-	var g Group
-	var ctx = r.Context()
-
-	if err := json.NewDecoder(r.Body).Decode(&g); err != nil {
-		h.H.WriteError(w, r, errors.WithStack(err))
+	if err := h.checkRequest(r, h.PrefixResource(GroupsResource), "create"); err != nil {
```
zepatrik: isn't it better to always first check the request?

warden/group/handler.go
```diff
@@ -384,27 +363,58 @@ func (h *Handler) AddGroupMembers(w http.ResponseWriter, r *http.Request, ps htt
...
-	}, Scope); err != nil {
+	if err := h.Manager.RemoveGroupMembers(id, m.Members); err != nil {
 		h.H.WriteError(w, r, err)
 		return
 	}
 
-	if err := h.Manager.RemoveGroupMembers(id, m.Members); err != nil {
+	w.WriteHeader(http.StatusNoContent)
+}
+
```
zepatrik: missing comment

warden/group/manager_sql.go
```diff
@@ -183,3 +187,33 @@ func (m *SQLManager) ListGroups(limit, offset int) ([]Group, error) {
...
+
+func (m *SQLManager) OverwriteGroupMembers(group string, members []string) error {
+	if err := m.applyInTransaction(m.deleteGroup(group), m.createGroup(group), m.addGroupMembers(group, members)); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (m *SQLManager) applyInTransaction(requests ...func(tx *sqlx.Tx) error) error {
```
zepatrik: check if rollbacks are correct

zepatrik:
shouldn't [this](https://github.com/ory/hydra/pull/768/files#diff-f5038ba6594bf60ef28399e90b84dc8aR120) function (and the next one) be non capital because it is not checking the request at all, it is just a helper right?

warden/group/handler.go
```diff
@@ -384,27 +363,58 @@ func (h *Handler) AddGroupMembers(w http.ResponseWriter, r *http.Request, ps htt
...
+func (h *Handler) OverwriteGroupMembers(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
```
arekkas: I would probably rename it to "UpdateGroup"

warden/group/handler.go
arekkas: Yes, maybe I was using `g` here for some reason which was why it was decoded first, but as it stands now the order can be reversed´the way you did it.

warden/group/manager_sql.go
arekkas: `requests` are really `transactions`, right?

warden/group/manager_sql.go
```diff
@@ -183,3 +187,33 @@ func (m *SQLManager) ListGroups(limit, offset int) ([]Group, error) {
...
+	tx, err := m.DB.Beginx()
+	if err != nil {
+		return errors.Wrap(err, "Could not begin transaction")
+	}
+
+	for _, req := range requests {
+		if err := req(tx); err != nil {
+			tx.Rollback()
```
arekkas: If rollback fails, the error should be returned instead

```
if err := tx.Rollback(); err != nil {
  return err
}
```

warden/group/manager_sql.go
```diff
@@ -183,3 +187,33 @@ func (m *SQLManager) ListGroups(limit, offset int) ([]Group, error) {
 
 	return groups, nil
 }
+
+func (m *SQLManager) OverwriteGroupMembers(group string, members []string) error {
+	if err := m.applyInTransaction(m.deleteGroup(group), m.createGroup(group), m.addGroupMembers(group, members)); err != nil {
```
arekkas: this can be simplified to `return m.ApplyIn...`

warden/group/manager_sql.go
```diff
@@ -88,58 +94,56 @@ func (m *SQLManager) GetGroup(id string) (*Group, error) {
...
-			if err := tx.Rollback(); err != nil {
+func (m *SQLManager) AddGroupMembers(group string, subjects []string) error {
+	return m.applyInTransaction(m.addGroupMembers(group, subjects))
+}
+
+func (m *SQLManager) removeGroupMembers(group string, subjects []string) func(tx *sqlx.Tx) error {
+	return func(tx *sqlx.Tx) error {
+		for _, subject := range subjects {
+			if _, err := tx.Exec(m.DB.Rebind("DELETE FROM hydra_warden_group_member WHERE member=? AND group_id=?"), subject, group); err != nil {
+				if err := tx.Rollback(); err != nil {
```
arekkas: You have duplicate rollbacks here, let `applyInTransaction` handle rollbacks instead!

warden/group/manager_sql.go
```diff
@@ -88,58 +94,56 @@ func (m *SQLManager) GetGroup(id string) (*Group, error) {
...
 
-	for _, subject := range subjects {
-		if _, err := tx.Exec(m.DB.Rebind("INSERT INTO hydra_warden_group_member (group_id, member) VALUES (?, ?)"), group, subject); err != nil {
-			if err := tx.Rollback(); err != nil {
-				return errors.WithStack(err)
+func (m *SQLManager) addGroupMembers(group string, subjects []string) func(tx *sqlx.Tx) error {
+	return func(tx *sqlx.Tx) error {
+		for _, subject := range subjects {
+			if _, err := tx.Exec(m.DB.Rebind("INSERT INTO hydra_warden_group_member (group_id, member) VALUES (?, ?)"), group, subject); err != nil {
+				if err := tx.Rollback(); err != nil {
```
arekkas: Rollback is already covered by `applyInTransaction`

warden/group/manager_sql.go
```diff
@@ -57,16 +57,22 @@ func (m *SQLManager) CreateSchemas() (int, error) {
 	return n, nil
 }
 
+func (m *SQLManager) createGroup(group string) func(tx *sqlx.Tx) error {
+	return func(tx *sqlx.Tx) error {
+		if _, err := tx.Exec(m.DB.Rebind("INSERT INTO hydra_warden_group (id) VALUES (?)"), group); err != nil {
+			return err
```
arekkas: Normal errors do not have stack traces in go, to add a stack trace do `return errors.WithStack(err)`

warden/group/manager_sql.go
```diff
@@ -88,58 +94,56 @@ func (m *SQLManager) GetGroup(id string) (*Group, error) {
 	}, nil
 }
 
-func (m *SQLManager) DeleteGroup(id string) error {
-	if _, err := m.DB.Exec(m.DB.Rebind("DELETE FROM hydra_warden_group WHERE id=?"), id); err != nil {
-		return errors.WithStack(err)
+func (m *SQLManager) deleteGroup(id string) func(tx *sqlx.Tx) error {
+	return func(tx *sqlx.Tx) error {
+		if _, err := tx.Exec(m.DB.Rebind("DELETE FROM hydra_warden_group WHERE id=?"), id); err != nil {
+			return err
```
arekkas: Normal errors do not have stack traces in go, to add a stack trace do `return errors.WithStack(err)`

warden/group/manager_sql.go
```diff
@@ -88,58 +94,56 @@ func (m *SQLManager) GetGroup(id string) (*Group, error) {
...
-			if err := tx.Rollback(); err != nil {
-				return errors.WithStack(err)
+func (m *SQLManager) addGroupMembers(group string, subjects []string) func(tx *sqlx.Tx) error {
+	return func(tx *sqlx.Tx) error {
+		for _, subject := range subjects {
+			if _, err := tx.Exec(m.DB.Rebind("INSERT INTO hydra_warden_group_member (group_id, member) VALUES (?, ?)"), group, subject); err != nil {
+				if err := tx.Rollback(); err != nil {
+					return err
+				}
+				return err
```
arekkas: Normal errors do not have stack traces in go, to add a stack trace do `return errors.WithStack(err)`

warden/group/manager_sql.go
zepatrik: Oh now I got it, I thought that `errors.WithStack(err)` just has to be called once to get the whole stacktrace.

warden/group/manager_sql.go
zepatrik: A transaction is a sequence of requests I would rather say.

warden/group/manager_sql.go
arekkas: Exactly, and it also needs to be called at the location which is the deepest in the stack trace and there only once - otherwise you'll overwrite the stack trace. This is a bit tricky unfortunately. The rule is that we wrap all errors coming from other libraries but don't wrap errors coming from our own. That way we ensure that traces are always as deep as possible in the stack.

warden/group/manager_sql.go
arekkas: How about `executors`?

warden/group/handler.go
zepatrik: I think I'm going for `UpdateGroupMembers` because the http PUT url is `GroupsHandlerPath + "/:id/members"` so only one group is affected, not all of them

warden/group/manager_sql.go
zepatrik: So [this](https://github.com/ory/hydra/pull/768/files#diff-70ba1b1b4893bb9e2acf833a21da73b7R182) should be actually without a stacktrace?

zepatrik:
ready for review @arekkas

warden/group/handler.go
```diff
@@ -384,27 +363,88 @@ func (h *Handler) AddGroupMembers(w http.ResponseWriter, r *http.Request, ps htt
...
+//     Produces:
+//     - application/json
+//
+//     Schemes: http, https
+//
+//     Security:
+//       oauth2: hydra.warden.groups
+//
+//     Responses:
+//       204: emptyResponse
```
arekkas: `PUT` requests typically don't return 204 responses but instead 200 + body. 204 is typically used for `DELETE` responses to aknowledge that the resource is gone

warden/group/manager_sql.go
```diff
@@ -57,16 +57,20 @@ func (m *SQLManager) CreateSchemas() (int, error) {
 	return n, nil
 }
 
+func (m *SQLManager) createGroup(group string) func(tx *sqlx.Tx) error {
+	return func(tx *sqlx.Tx) error {
+		_, err := tx.Exec(m.DB.Rebind("INSERT INTO hydra_warden_group (id) VALUES (?)"), group)
+
+		return errors.WithStack(err)
```
arekkas: Not sure if this works for nil values? So if error is nil, does this return nil as well or not?

zepatrik: https://godoc.org/github.com/pkg/errors#WithStack
I looked it up it works as expected

warden/group/handler.go
zepatrik: I may not have changed everything correctly when copying the comments

warden/group/manager_sql.go
arekkas: Absolutely, it returns nil in that case: https://github.com/pkg/errors/blob/master/errors.go#L145

warden/group/handler.go
arekkas: It's also in the response, see `w.WriteHeader(http.StatusNoContent)`

arekkas:
This is now tracked as https://github.com/ory/keto/issues/10

PR #780, opened 2018-02-13
Update newsletter text

### LINKED ISSUE #779: make --skip-newsletter the default
At least to me it is quite weird to have a cli ask for a newsletter. Is this really the way to treat users to do that by default, especially if --id and --secret are used (which is incredibly helpful for tests, setting up testusers etc)

zepatrik:
Before newsletter text did not seem to make clear that it is just for security information.
The text is updated according to #779, closes #779


arekkas:
Nice, thank you!

PR #783, opened 2018-02-15
Remove coveralls token from circleci config

### LINKED ISSUE #782: Don't push to coveralls in CI when PR comes from fork
See https://circleci.com/gh/ory/hydra/1225?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

zepatrik:
Before forked PRs failed on circleci because $COVERALLS_REPO_TOKEN was not set. This token does seem not to be necessary at all because hydra is a public repository (see #782).
closes #782

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,7 +38,7 @@ jobs:
       - run: go-acc -o coverage.txt ./...
       - run: go test -race -short $(go list ./... | grep -v cmd)
       - run: ./scripts/test-e2e.sh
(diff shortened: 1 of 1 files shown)


arekkas:
Failing test is not your fault, re-running

arekkas:
@zepatrik ok seems like this still fails...but now due to coveralls :)

zepatrik:
The [coveralls api reference](https://docs.coveralls.io/api-reference) says the token is required. Trying the bash test approach for the moment.

PR #845, opened 2018-05-09
add /health/version endpoint

zepatrik:
Add the `/healt/version` endpoint.
It serves the same version string that can be obtained by running `hydra version`


arekkas:
Please resolve the conflict and add a test :)

PR #849, opened 2018-05-17
5 minutes tutorial url is broken

gonzalesraul:
Previously, the URL was https://www.ory.sh/docs/1-hydra/0-tutorial/0-readme, which result on *Page Not Found*.
Updated to  the correct one https://www.ory.sh/docs/guides/latest/hydra/1-tutorial

Signed-off-by: Raul Gonzales <gonzalesraul03@gmail.com>


arekkas:
I think this is fixed now and the redirections work as they should, so I'm closing this. In any case, thank you for raising this PR!

PR #870, opened 2018-05-23
Client secret expires

### LINKED ISSUE #778: client: Add field `client_secret_expires_at` to create
>  client_secret_expires_at
      REQUIRED if "client_secret" is issued.  Time at which the client
      secret will expire or 0 if it will not expire.  The time is
      represented as the number of seconds from 1970-01-01T00:00:00Z as
      measured in UTC until the date/time of expiration.

Since we're currently not expiring secrets, it should just show a 0

zepatrik:
introduces the `client_secret_expires_at` field without any functionality
it should always be 0
closes #778


arekkas:
Tests fail

PR #954, opened 2018-07-29
Add api endpoint to list all authorized clients by user

### LINKED ISSUE #953: consent: expose  a list of all clients authorized by a user
**Do you want to request a *feature* or report a *bug*?**
Feature

There is currently no direct way to get a list of all clients for which a particular user has given consent. @joshuarubin has already mentioned in #304 that:

>We, however, would like to have a single interface for the user to see all the applications they have approved and have an opportunity to revoke access for any particular app.

As he also wrote, this is currently feasible by storing the authorized clients in the Consent application. This information is already available in Hydra (you can even delete all tokens since #304, but you can't list them).

Although I am a Go beginner, I don't think it should be too complicated, since `FindPreviouslyGrantedConsentRequests` already provides a similar functionality. If the idea meets with approval here, I would try a pull request.

kingjan1999:
Resolves #953 

As already mentioned in #953, I'm rather beginner in Go, so I'm open to comments and suggestions for improvement and ask for leniency for possible mistakes.

Signed-off-by: Jan Beckmann <king-jan1999@hotmail.de>


consent/handler.go
```diff
@@ -133,6 +134,47 @@ func (h *Handler) DeleteUserClientConsentSession(w http.ResponseWriter, r *http.
...
+//       500: genericError
+
+func (h *Handler) GetConsentSessions(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	user := ps.ByName("user")
+	if user == "" {
+		h.H.WriteError(w, r, errors.WithStack(fosite.ErrInvalidRequest.WithDebug("Parameter user is not defined")))
+		return
+	}
+
+	sessions, err := h.M.FindPreviouslyGrantedConsentRequestsByUser(user)
```
arekkas: I really think we need pagination here, because this can be a lot! Examples can be found [here](https://github.com/ory/hydra/blob/master/client/handler.go#L192-L193), [here](https://github.com/ory/hydra/blob/master/client/manager_sql.go#L377-L381) and [here](https://github.com/ory/hydra/blob/master/client/manager_memory.go#L151-L154).

consent/manager_memory.go
```diff
@@ -162,6 +162,22 @@ func (m *MemoryManager) VerifyAndInvalidateConsentRequest(verifier string) (*Han
 }
 
 func (m *MemoryManager) FindPreviouslyGrantedConsentRequests(client string, subject string) ([]HandledConsentRequest, error) {
+	var rs []HandledConsentRequest
+	filteredByUser, _ := m.FindPreviouslyGrantedConsentRequestsByUser(subject)
+	for _, c := range filteredByUser {
+		if client != c.ConsentRequest.Client.GetID() {
```
arekkas: You can revert this logic to `==` and add `rs = append(rs, c)` in the loop, it is easier to read!

consent/types.go
```diff
@@ -96,7 +96,7 @@ type HandledConsentRequest struct {
 	// authorization will be remembered indefinitely.
 	RememberFor int `json:"remember_for"`
 
-	ConsentRequest  *ConsentRequest     `json:"-"`
+	ConsentRequest  *ConsentRequest     `json:"consent_request"`
```
arekkas: The problem with exposing this is that acceptance of the consent request could overwrite this value and potentially mess stuff up. I haven't checked if that's actually the case but it could be an issue in some regression. I think it would be ok to copy this struct 1:1 and there set the `consent_request` value. If the order of the fields is the same, you can easily convert the type:

```go
var foo HandledConsentRequest
var bar = NewTypeWithJSON(HandledConsentRequest)
```

arekkas:
I'm pinging you @kingjan1999 so this doesn't get lost in your notifications. This looks already very solid, just a few things which I mentioned during review :)

consent/handler.go
kingjan1999: Thanks, I adapted it. However, I was a bit unsure in `consent/manager_memory.go ` as `FindPreviouslyGrantedConsentRequests` depends now on the newly created `FindPreviouslyGrantedConsentRequestsByUser` (which requires a limit and offset now) and does not support limit and offset. As you'll see, I've used a little workaround that I'm personally not 100% satisfied with.

consent/types.go
kingjan1999: I called it `HandledConsentRequestResponse` now but other suggestions are very welcome

kingjan1999:
Thanks for your help @arekkas !  I hope I have implemented all your proposals. 

When Swagger generated the SDK, some files (concerning JSONWebKeys) appeared for unknown reasons, which I have now simply commited.

consent/types.go
arekkas: How about `PreviousConsentSession`? :)

consent/handler.go
arekkas: I see, I don't think this solution will work with SQL as they don't support negative offset (I think).

arekkas: Oh sorry, I didn't see that you duplicated the other method. Then it will obviously work. I agree that it's not the cleanest solution but it does get the job done for now. So from my side, this is fine.

arekkas:
The only thing left is the naming of the one struct and then this is good to get merged from my side! Thank you so much for your contribution :)

kingjan1999:
Thanks for your naming suggestion @arekkas ! I've implemented it accordingly.

arekkas:
Perfect!

consent/manager_memory.go
```diff
@@ -162,6 +163,21 @@ func (m *MemoryManager) VerifyAndInvalidateConsentRequest(verifier string) (*Han
 }
 
 func (m *MemoryManager) FindPreviouslyGrantedConsentRequests(client string, subject string) ([]HandledConsentRequest, error) {
+	var rs []HandledConsentRequest
+	filteredByUser, _ := m.FindPreviouslyGrantedConsentRequestsByUser(subject, -1, -1)
```
arekkas: The error *must* be handled here!

consent/manager_sql.go
```diff
@@ -364,8 +364,35 @@ WHERE
...
+	AND
+		(h.error='{}' AND h.remember=TRUE)
+LIMIT ? OFFSET ?
+`), subject, limit, offset); err != nil {
+		if err == sql.ErrNoRows {
+			return nil, errors.WithStack(errNoPreviousConsentFound)
+		}
+		return nil, sqlcon.HandleError(err)
+	}
+	aa, err := m.resolveHandledConsentRequests(a)
```
arekkas: You can rewrite this to `return m.resolveHandledConsentRequests(a)
`

consent/manager_sql.go
```diff
@@ -376,12 +403,10 @@ WHERE
 		if v.RememberFor > 0 && v.RequestedAt.Add(time.Duration(v.RememberFor)*time.Second).Before(time.Now().UTC()) {
 			continue
 		}
-
 		va, err := v.toHandledConsentRequest(r)
 		if err != nil {
 			return nil, err
 		}
-
```
arekkas: Re-add newline

consent/manager_sql.go
```diff
@@ -390,4 +415,5 @@ WHERE
 	}
 
 	return aa, nil
+
```
arekkas: Remove newline

consent/manager_sql.go
```diff
@@ -376,12 +403,10 @@ WHERE
 		if v.RememberFor > 0 && v.RequestedAt.Add(time.Duration(v.RememberFor)*time.Second).Before(time.Now().UTC()) {
 			continue
 		}
-
```
arekkas: Re-add newline

consent/manager_sql.go
```diff
@@ -364,8 +364,35 @@ WHERE
...
+LIMIT ? OFFSET ?
+`), subject, limit, offset); err != nil {
+		if err == sql.ErrNoRows {
+			return nil, errors.WithStack(errNoPreviousConsentFound)
+		}
+		return nil, sqlcon.HandleError(err)
+	}
+	aa, err := m.resolveHandledConsentRequests(a)
+	return aa, err
+
```
arekkas: Remove newline

consent/manager_memory.go
```diff
@@ -202,8 +214,11 @@ func (m *MemoryManager) FindPreviouslyGrantedConsentRequests(client string, subj
 	if len(rs) == 0 {
 		return []HandledConsentRequest{}, nil
 	}
-
-	return rs, nil
+	if limit < 0 && offset < 0 {
+		return rs, nil
+	}
+	start, end := pagination.Index(limit, offset, len(rs))
```
arekkas: Add a newline on top of `start, end := ..`

consent/manager_memory.go
```diff
@@ -202,8 +214,11 @@ func (m *MemoryManager) FindPreviouslyGrantedConsentRequests(client string, subj
 	if len(rs) == 0 {
 		return []HandledConsentRequest{}, nil
 	}
-
```
arekkas: Re-add this newline

consent/handler.go
```diff
@@ -133,6 +135,55 @@ func (h *Handler) DeleteUserClientConsentSession(w http.ResponseWriter, r *http.
 	w.WriteHeader(http.StatusNoContent)
 }
 
+// swagger:route GET /oauth2/auth/sessions/consent/{user} oAuth2 listUserClientConsentSessions
```
arekkas: I think the naming should be `listUserConsentSessions`. With the current naming, I would expect the method to be like `listUserClientConsentSessions(user, client string)`

consent/manager_sql.go
```diff
@@ -364,8 +364,35 @@ WHERE
...
+JOIN
+	hydra_oauth2_consent_request as r ON (h.challenge = r.challenge)
+WHERE
+		r.subject=? AND r.skip=FALSE
+	AND
+		(h.error='{}' AND h.remember=TRUE)
+LIMIT ? OFFSET ?
+`), subject, limit, offset); err != nil {
+		if err == sql.ErrNoRows {
+			return nil, errors.WithStack(errNoPreviousConsentFound)
```
arekkas: So one issue with this is that the HTTP handler function will return `errNoPreviousConsentFound` which is a "dumb" error (meaning it will return a 500 status code with no context in the API). On the same hand, this error is required to be returned for the `consent_strategy` to work. Since you duplicated the methods and this particular method is not used in the `consent_strategy`, you can remove the `err == sql.ErrNoRows` check and just return `return nil, sqlcon.HandleError(err)`.

Please also add a test for this. It should check that `sqlcon.ErrNoRows` is returned when we expect that no session exist in the store. This will return a proper 404 error in the API which I think is to be expected here.

In the handler, you may want to choose to ignore the error of type `sqlcon.ErrNoRows` and instead just write an empty array. This will return `[]` instead of a 404 which might be a better suited response.

kingjan1999: Thanks for your help and your feedback! I have gladly implemented your suggestions. 

However, I'm a little unsure at this point and need your help: Does `db.Select` really return an `ErrNoRows` error if no rows could be found? According to my understanding of SQLX / Go SQL documentation, this does not happen here. I have not been able to force such behaviour in my experiments.

Or did I misunderstand and the manager should generate and return an `sqlcon.ErrNoRows`?

arekkas:
Thank you for the updates! The newly generated SDK breaks tests because the methods have been renamed:

```
consent/sdk_test.go:121:24: sdk.ListUserClientConsentSessions undefined (type *hydra.CodeGenSDK has no field or method ListUserClientConsentSessions)
consent/sdk_test.go:128:23: sdk.ListUserClientConsentSessions undefined (type *hydra.CodeGenSDK has no field or method ListUserClientConsentSessions)
```

arekkas:
Thank you so much for your hard work!

Issue #949, opened 2018-07-24
[Enhancement/Proposal] Update Plugin System


someone1:
Continuing discussion started in Issue #248 

> The *sqlx.DB reference for plugins is required as when its loaded an [immediate](https://github.com/ory/hydra/blob/master/config/backend_plugin.go#L84) call to db.Ping() is made. I can open a separate issue to drop the reference altogether as the plugin should be able to manage the connection on its own and doesn't need the reference it returns to be passed into its own exposed functions as a parameter.

Additionally, it would be great to extend plugins for not only database backends but for signing mechanisms as well (e.g. for HashiCorp Vault or GCP IAM API).

Finally, I also propose to enhance the system in general and export minor parts of the bootstrap process for those of us that want to customize hydra directly vs loading a plugin.

arekkas:
I think all of them are very sensible requests. Happy to talk proposals and PRs (please discuss first)!

someone1:
I actually have a PR for the first part done (database plugin) - I think it simplifies a lot of the database config process around hydra. I'll submit a PR shortly for this and maybe we can discuss the changes I made there (though no hard feelings if you reject it!)

I'd like to discuss the plugin extension around external JWT signing mechanisms - I think hydra needs to decouple the jwk bits by adding more interfaces like there is around the other parts of the code first.

someone1:
Actually, just read the contributing document. Here is my work in a [single commit](https://github.com/someone1/hydra/commit/93961d6c5dd3c2849c400c6ce517d5b834049fff).

The gist of the changes are:
- Introduce a backend interface for databases
- Create both memory and sql implementations of the new interface
- Update plugin load sequence to leverage new backed interface
- Update config/context loading for the new interface
- Update cmd/server bootstrap for the changes to config (greatly simplified)
- Enhancement to cmd/server: create an exported function that takes in a hydra.Config and spits out a http.Handler

It's a relatively small change architecture-wise (IMHO) - and it removes much more code than it adds while increasing flexibility of backends supported.

Let me know if you're open to a PR for this, I've run tests (with -short) and can follow the remaining steps in the contribution guide. Open to feedback as well!

arekkas:
This looks pretty solid. The dependency checks have been removed (they check if the dependency has been initialized and if not fatalf) which should definitely be re-added. I can't promise right now that this will stay 100% as is in the future as there might still be issues which we'll yet have to surface but from a high level perspective this looks good. Feel free to PR.

[...]

Issue #928, opened 2018-07-15
Panic when calling oauth2/auth/sessions/consent/{user} or oauth2/auth/sessions/consent/{user}/{client}


robotlovesyou:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
A panic when calling http://localhost:4444/oauth2/auth/sessions/consent/{user} or http://localhost:4444/oauth2/auth/sessions/consent/{user}/{client}

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

Follow the '5 minute tutorial' to create the auth-code-client and follow the auth code grant flow until the token is displayed by the browser.

Then 
```
curl -X "DELETE" "http://localhost:4444/oauth2/auth/sessions/consent/foo@bar.com"
```
or
```
curl -X "DELETE" "http://localhost:4444/oauth2/auth/sessions/consent/foo@bar.com/auth-code-client"
```

The request will cause a non HTTP error to return. When you observe the docker-compose log you will see the trace from the panic

**What is the expected behavior?**

That the user's consent is deleted and their tokens are revoked

**Which version of the software is affected?**
v1.0.0-beta.6

I've done a little digging and the panic is because the store property of the sql manager is nil

robotlovesyou:
The root cause of this is that in register routes in `cmd/server/handler.go` the injectConsentManager is called before the store is created by injectFositeStore. Changing the code so that those functions are called the other way around fixes the issue and the tests still pass. I would submit a patch but I don't have time to work out how to write a meaningful test right now. Sorry!

PR #956, opened 2018-07-31
all: Introduce database backend interface and update plugin system an…

someone1:
…d boostrap accordingly (#949)

### DIFF
--- a/cmd/server/handler_client_factory.go
+++ b/cmd/server/handler_client_factory.go
@@ -27,33 +27,11 @@ import (
 	"github.com/ory/herodot"
 	"github.com/ory/hydra/client"
 	"github.com/ory/hydra/config"
-	"github.com/ory/sqlcon"
 )
 
 func newClientManager(c *config.Config) client.Manager {
 	ctx := c.Context()
-
-	switch con := ctx.Connection.(type) {
-	case *config.MemoryConnection:
-		expectDependency(c.GetLogger(), ctx.Hasher)
-		return client.NewMemoryManager(ctx.Hasher)
-	case *sqlcon.SQLConnection:
-		expectDependency(c.GetLogger(), ctx.Hasher, con.GetDatabase())
-		return &client.SQLManager{
-			DB:     con.GetDatabase(),
-			Hasher: ctx.Hasher,
-		}
-	case *config.PluginConnection:
-		if m, err := con.NewClientManager(); err != nil {
-			c.GetLogger().Fatalf("Could not load client manager plugin %s", err)
-		} else {
-			return m
-		}
-		break
-	default:
-		panic("Unknown connection type.")
-	}
-	return nil
+	return ctx.Connection.NewClientManager(ctx.Hasher)
 }
 
 func newClientHandler(c *config.Config, router *httprouter.Router, manager client.Manager) *client.Handler {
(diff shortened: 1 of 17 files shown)


cmd/server/handler_client_factory.go
```diff
@@ -27,33 +27,13 @@ import (
...
-			c.GetLogger().Fatalf("Could not load client manager plugin %s", err)
-		} else {
-			return m
-		}
-		break
-	default:
-		panic("Unknown connection type.")
-	}
-	return nil
+	expectDependency(c.GetLogger(), ctx.Hasher, ctx.Connection)
```
arekkas: We don't really need the ctx.Hash/logger dependencies here. This could be the case for plugins for example (they might use a different hasher). I think it would make more sense to put `expectDependency` in the concrete implementations.

config/config.go
```diff
@@ -238,34 +237,34 @@ func (c *Config) Context() *Context {
...
+		scheme = u.Scheme
+	}
+
+	if backend, ok := backends[scheme]; ok {
+		if err := backend.Init(c.DatabaseURL, c.GetLogger()); err != nil {
+			c.GetLogger().Fatalf(`Could not connect to database backend: %s`, err)
 		}
+		connection = backend
+	} else {
+		c.GetLogger().Fatalf(`Unknown DSN "%s" in DATABASE_URL: %s`, scheme, c.DatabaseURL)
```
arekkas: I think this should read: `Unknown DSN scheme "%s" in DATABASE_URL "%s\", schemes %v supported`

config/backend_manager.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/ory/hydra/jwk"
+	"github.com/ory/hydra/pkg"
+)
+
+var (
+	backends = make(map[string]BackendManager)
+	bmutex   sync.Mutex
+)
+
+type BackendManager interface {
```
arekkas: I think `Connection` or `Connector` would be a better choice here :)

cmd/server/handler_client_factory.go
someone1: Hmmm - so move the function and all calls to config and call it as required there, e.g. for the SQL Manager? I'm not entirely sure why these were added so I just copied them back as best I could.

config/backend_manager.go
someone1: Sure, so `BackendConnector` then? Any thoughts on what variable we should load from the database plugins? I'm using `Manager` right now but can update this to `Connector` to match this name change.

cmd/server/handler_client_factory.go
arekkas: The problem is that some services have dependencies to others. In some cases (in the past) these had not been properly instantiated during the set up of a particular service/manager. This caused nil point exceptions (see #928 for example). To solve that I added this little helper that makes sure that no dependency is nil.

Adding this to the SQL or Memory Managers (where required) prevents that.

config/backend_manager.go
arekkas: Yup, `BackendConnector` or `Connector` both work fine. Choose one you like. The plugin should export the same name, so also either `BackendConnector` or `Connector`.

someone1:
Updated though one thing to note is dependency checks are still used in `cmd/server` unrelated to injecting so I basically just copied this out to `config` and left it there.

arekkas:
> Updated though one thing to note is dependency checks are still used in cmd/server unrelated to injecting so I basically just copied this out to config and left it there.

That makes total sense and is much cleaner. Thank you for your work on this, I truly think this is a step forwards to reusability and customization of Hydra and I really like the design you chose. Thank you for this contribution!

PR #961, opened 2018-08-05
cli: fix reporting of epected vs. received status codes

rjw57:
Asking for a non-existent client results in the following confusing error message:

```
Command failed because calling "GET http://hydra:4444/clients/no-such-client" resulted in status code "200" but code "404" was expected.
{"error":"Unable to locate the resource","error_description":"","status_code":404}
```

This commit fixes the expectedStatusCode and response.StatusCode arguments to fmt.Fprintf which were reversed.


arekkas:
Thank you!

Issue #970, opened 2018-08-06
Allow logging out and deleting a single session cookie


mderazon:
In version 1.0.0, Hydra started handling the user's session. You can ask Hydra to remember the session so next time the user logs in they will not be shown the login screen.

Currently, the only way to log the user out is by calling
```
DELETE https://hydra/oauth2/auth/sessions/login/{user}
```
This, however, deletes the cookie and logs the user out of **all** devices.

Following a discussion [in the forum](https://community.ory.sh/t/how-to-implement-remember-me-option-on-login-page/537/3), the plan is to add an endpoint
```
GET http://hydra/oauth2/login/signout
```
That will basically delete the saved login cookie and redirect the user to a whitelisted url.

@arekkas How should the redirect work ?
Should the user define the redirect by adding a `redirect=` query param to the signout url and if the domain matches the whitelisted one then redirection is allowed ?
What should be the behaviour in case the redirect does not match the whitelisted one ? (need a fallback url ? or do we just return a 400 response ?)

arekkas:
I think there should be an environment variable `LOGOUT_REDIRECT_URL` to which the browser is redirected upon successful logout. There should not be an option to set the redirect URL from the browser. If wee need it, we can add that later.

The method for logging out is [already implemented](https://github.com/ory/hydra/blob/master/consent/strategy_default.go#L241). All we need to do is expose this method via a URL. The HTTP handler should probably be in [this file](https://github.com/ory/hydra/blob/master/consent/handler.go#L59).

I think the URL should be something like `/oauth2/auth/logout`. Be aware that the router being used there is the one for port 4445. You will have to get the second router (frontend router on port 4444) for this to work.

PR #971, opened 2018-08-06
consent: Add logout api endpoint

mderazon:
ory/hydra#970

### DIFF
--- a/consent/handler.go
+++ b/consent/handler.go
@@ -56,19 +56,21 @@ func NewHandler(
 	}
 }
 
-func (h *Handler) SetRoutes(r *httprouter.Router) {
-	r.GET(LoginPath+"/:challenge", h.GetLoginRequest)
-	r.PUT(LoginPath+"/:challenge/accept", h.AcceptLoginRequest)
-	r.PUT(LoginPath+"/:challenge/reject", h.RejectLoginRequest)
+func (h *Handler) SetRoutes(frontend, backend *httprouter.Router) {
+	backend.GET(LoginPath+"/:challenge", h.GetLoginRequest)
+	backend.PUT(LoginPath+"/:challenge/accept", h.AcceptLoginRequest)
+	backend.PUT(LoginPath+"/:challenge/reject", h.RejectLoginRequest)
 
-	r.GET(ConsentPath+"/:challenge", h.GetConsentRequest)
-	r.PUT(ConsentPath+"/:challenge/accept", h.AcceptConsentRequest)
-	r.PUT(ConsentPath+"/:challenge/reject", h.RejectConsentRequest)
+	backend.GET(ConsentPath+"/:challenge", h.GetConsentRequest)
+	backend.PUT(ConsentPath+"/:challenge/accept", h.AcceptConsentRequest)
+	backend.PUT(ConsentPath+"/:challenge/reject", h.RejectConsentRequest)
 
-	r.DELETE("/oauth2/auth/sessions/login/:user", h.DeleteLoginSession)
-	r.GET("/oauth2/auth/sessions/consent/:user", h.GetConsentSessions)
-	r.DELETE("/oauth2/auth/sessions/consent/:user", h.DeleteUserConsentSession)
-	r.DELETE("/oauth2/auth/sessions/consent/:user/:client", h.DeleteUserClientConsentSession)
+	backend.DELETE("/oauth2/auth/sessions/login/:user", h.DeleteLoginSession)
+	backend.GET("/oauth2/auth/sessions/consent/:user", h.GetConsentSessions)
+	backend.DELETE("/oauth2/auth/sessions/consent/:user", h.DeleteUserConsentSession)
+	backend.DELETE("/oauth2/auth/sessions/consent/:user/:client", h.DeleteUserClientConsentSession)
+
+	frontend.GET("/oauth2/auth/logout", h.LogoutUser)
 }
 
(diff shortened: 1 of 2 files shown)


mderazon:
@arekkas the fix is very small, however since it's my first time dealing with Go, I was hitting some issues running tests (missing packages, even after `go get`) and also a weird behaviour something making changes to files in the project that I didn't ask for.

I checked and the changes were forced even after I closed VSC, very bizzare. No way to revert them, they just keep popping back.

Anyway, would probably take me a lot more time to understand the tooling around it than to actually make the changes.


Regarding the changes themselves:
1. I am not sure I wrote the actual redirect correctly: https://github.com/ory/hydra/pull/971/commits/171fb6d8d26d98a4a409fff3666b6decae46e51a#diff-ac4ce4ab0ca6c69035745a50d3108c65R603. Also, should it be 302 redirect ?
2. Please review the swagger decorators for the route. Not sure it's good (see description and return type etc)
3. Please review the route handler name. Naming is hard...
4. Couldn't find a place to write test for this route.

Anyway, hope it's not creating more work than it would for you to write it :-/

consent/handler.go
```diff
@@ -573,4 +575,37 @@ func (h *Handler) RejectConsentRequest(w http.ResponseWriter, r *http.Request, p
...
+//       500: genericError
+
+func (h *Handler) LogoutUser(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	err := revokeAuthenticationSession(w, r)
+	if err != nil {
+		h.H.WriteError(w, r, err)
+		return
+	}
+
+	http.Redirect(w, r, os.Getenv("LOGOUT_REDIRECT_URL"), 302)
```
arekkas: This solution is ok, but it makes setting defaults and testing the configuration harder. Typically what we do here is:

1. Add a config item in [config/config.go](https://github.com/ory/hydra/blob/master/config/config.go#L69)
2. Add a field (e.g. `LogoutRedirectURL`) to set this in the [handler.go](https://github.com/ory/hydra/blob/master/consent/handler.go#L41) and the respective factors (`new...`)
3. Bind and set a default in `cmd/root.go` like [here](https://github.com/ory/hydra/blob/master/cmd/root.go#L119-L120).
4. Use `LogoutRedirectURL` here instead of `os.GetEnv()`

Additionally, we should set up a default handler for the log out screen, similar to [this one](https://github.com/ory/hydra/blob/master/oauth2/handler_fallback_endpoints.go#L52) and add it to the frontend router [here](https://github.com/ory/hydra/blob/master/oauth2/handler.go#L161). The layout of the page doesn't matter. It should read something "You have been logged out successfully. ... Admin forgot to set up the redirect URL after logout ... to fix that set environment variable `LogoutRedirectURL`, e.g. `export LogoutRedirectURL=http://...` or on windows `set LogoutRedirectURL=http://...` ..."

consent/handler.go
```diff
@@ -56,19 +56,21 @@ func NewHandler(
...
+	backend.GET(ConsentPath+"/:challenge", h.GetConsentRequest)
+	backend.PUT(ConsentPath+"/:challenge/accept", h.AcceptConsentRequest)
+	backend.PUT(ConsentPath+"/:challenge/reject", h.RejectConsentRequest)
+
+	backend.DELETE("/oauth2/auth/sessions/login/:user", h.DeleteLoginSession)
+	backend.GET("/oauth2/auth/sessions/consent/:user", h.GetConsentSessions)
+	backend.DELETE("/oauth2/auth/sessions/consent/:user", h.DeleteUserConsentSession)
+	backend.DELETE("/oauth2/auth/sessions/consent/:user/:client", h.DeleteUserClientConsentSession)
+
+	frontend.GET("/oauth2/auth/logout", h.LogoutUser)
```
arekkas: Let's rename this to `UserLogout` (because it's `UserConsent` :) )

arekkas:
One more thing, please document `LOGOUT_REDIRECT_URL` somewhere around [here](https://github.com/ory/hydra/blob/master/cmd/serve.go#L96)

mderazon:
Thanks for the valuable feedback.
One more thing - What should be the behaviour if the user does not set `LOGOUT_REDIRECT_URL` ?

arekkas:
> One more thing - What should be the behaviour if the user does not set LOGOUT_REDIRECT_URL ?
Two options

See this comment: https://github.com/ory/hydra/pull/971#discussion_r208136347

> Additionally, we should set up a default handler for the log out screen, similar to this one and add it to the frontend router here. The layout of the page doesn't matter. It should read something "You have been logged out successfully. ... Admin forgot to set up the redirect URL after logout ... to fix that set environment variable LogoutRedirectURL, e.g. export LogoutRedirectURL=http://... or on windows set LogoutRedirectURL=http://... ..."

arekkas:
I know now why this fails. Apparently, `gofmt` tries to format the .php and .js files which causes this error (uppercasing). Not sure why it does that but that's the error. Happened to me as well.

mderazon:
Were you able to disable it ?

arekkas:
It was really weird, I commited all the changes to a new branch, then switched back. Only then did they disappear. I honestly have no idea what's going on here...

arekkas:
Did you manage to get it fixed? Maybe it was a broken git index? I had the issue several times too. I finally fixed it though, I think it had something to do with lower/uppercase of the filename. If this blocks you from working on the PR let me know.

arekkas:
Just letting you know that I'll take your changes and integrate my feedback so this can be released with beta.8 today

arekkas:
Thank you for your hard work - new PR is #984

arekkas:
No problem!

PR #972, opened 2018-08-07
oauth2: Refactor JWT strategy

### LINKED ISSUE #960: oauth2: abstract oauth2/handler JWT Strategies
I mentioned this in the chat a week or two ago and I believe @arekkas responded with leaning towards introducing an interface to implement this.

Basically, I'd like to use an external JWT signing mechanism. I'd like to bootstrap Hydra on my own in order to do this and for the most part I can, however, the [introduction](https://github.com/ory/hydra/commit/c932ab4571f1ae75c526e9b19d5a7c60d533ca41#diff-405b3db63f972dfe0172d5c39904ebce) of `OpenIDJWTStrategy *jwk.RS256JWTStrategy` and `AccessTokenJWTStrategy *jwk.RS256JWTStrategy` in the `oauth2/handler` struct is preventing me from doing so.

I know there was some talk about introducing a plugin system for external key management/signing services though I think this is a good first step in implementing such a system. If we can introduce a basic interface for these two I can start devoting time to coming up with a system by hacking in my own deployment before proposing something here.

From what I can tell, the `oauth2/handler` only expects the following for both of these:

```go
import (
	jwt2 "github.com/dgrijalva/jwt-go"
	"github.com/ory/fosite/token/jwt"
)

type JWTStrategyer interface {
	GetPublicKeyID() (string, error)
	Generate(claims jwt2.Claims, header jwt.Mapper) (string, string, error)
}
```

`AccessTokenJWTStrategy` only appears to be used when its not nil, meaning hydra was bootstrapped to utilize JWT for Access Tokens. The `oauth2/handler` only cares about the PublicKeyId this has to return which is passed along the Token and Auth handlers.

`OpenIDJWTStrategy` expects to be present so hydra can use it's PublicKeyId in the Auth handler response, and if a client is using signed responses, the UserInfo handler will sign it's response with this strategy.

If it's alright with you, I can submit a simple PR to introduce the interface and update the handler accordingly.

someone1:
(closes #960)

### DIFF
--- a/cmd/server/handler_oauth2_factory.go
+++ b/cmd/server/handler_oauth2_factory.go
@@ -194,6 +194,7 @@ func newOAuth2Handler(c *config.Config, frontend, backend *httprouter.Router, cm
 		L:                      c.GetLogger(),
 		OpenIDJWTStrategy:      openIDJWTStrategy,
 		AccessTokenJWTStrategy: accessTokenJWTStrategy,
+		AccessTokenStrategy:    c.OAuth2AccessTokenStrategy,
 		IDTokenLifespan:        c.GetIDTokenLifespan(),
 	}
 
--- a/jwk/jwt_strategy.go
+++ b/jwk/jwt_strategy.go
@@ -29,6 +29,12 @@ import (
 	"github.com/pkg/errors"
 )
 
+type JWTStrategy interface {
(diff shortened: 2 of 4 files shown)


arekkas:
LGTM

PR #986, opened 2018-08-13
config: disable plugin backend through 'noplugin' tag 

glerchundi:
Debugging Hydra in Go 1.10 and 1.11 (confirmed by one of its members),
is not possible due to [this unresolved
bug](https://github.com/golang/go/issues/23733) which is related to the
use of the plugin functionality.

This change allows passing a build tag which will disable plugin
implementation and therefore allow to debug in all the use-cases where
plugin backend is not needed.

Signed-off-by: Gorka Lerchundi Osa <glertxundi@gmail.com>


arekkas:
Thanks!

PR #1015, opened 2018-08-30
config: Fix use of uninitialized logger

vHanda:
The MustValidate() function is sometimes called before any other logging
function has been called and this results in a crash. An easy way to
reproduce the crash is to change OAUTH2_ACCESS_TOKEN_STRATEGY=jwt in the
default docker-compose.yml

Signed-off-by: Vishesh Handa <vishesh.handa@telefonica.com>


aeneasr:
thanks!

PR #1019, opened 2018-09-01
Add support for distributed tracing

aaslamin:
## So what does this PR do?

• Adds support for the [Jaeger Distributed Tracing System](https://www.jaegertracing.io/docs/)
• Docker setup for loading up Jaeger with Hydra

#### Notes 🗒:

• The tracer can be configured w/ sampling via config https://www.jaegertracing.io/docs/sampling/
• By using the OpenTracing API, which is vendor neutral, will make adding support for other OpenTracing compliant tracers straightforward.
• This is the first step...the real work is to follow as we will need to span & propagate all teh things! (_e.g. db lookups, encryption, calls to the consent app, etc._)

## Cool, how can I get this up and running locally? 🎩 💻 

I have updated the docker-compose to bring up the Jaeger tracer and added config for the Hydra container to use it as the collector.

• Clone Hydra to your GOPATH
• Checkout to this branch
• Run `docker-compose up --build`
    _(Good time to grab a ☕️...)_
• Make a request to Hydra now, for example:
```
docker exec -it hydra_hydra_1 \
    hydra clients create \
    --endpoint http://localhost:4445 \
    --id my-client \
    --secret secret \
    -g client_credentials
```
• Open up your browser to: http://localhost:16686/search
• From the dropdown, select _Hydra_
image
• At the bottom, click on _Find Traces_
• You should be able to see your first trace 🎉:
image

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -197,6 +197,24 @@ func initConfig() {
 	viper.BindEnv("OIDC_SUBJECT_TYPE_PAIRWISE_SALT")
 	viper.SetDefault("OIDC_SUBJECT_TYPE_PAIRWISE_SALT", "public")
 
+	viper.BindEnv("TRACING_PROVIDER")
+	viper.SetDefault("TRACING_PROVIDER", "")
+
+	viper.BindEnv("TRACING_PROVIDER_JAEGER_SAMPLING_SERVER_URL")
+	viper.SetDefault("TRACING_PROVIDER_JAEGER_SAMPLING_SERVER_URL", "http://localhost:5778/sampling")
+
+	viper.BindEnv("TRACING_PROVIDER_JAEGER_SAMPLING_TYPE")
+	viper.SetDefault("TRACING_PROVIDER_JAEGER_SAMPLING_TYPE", "const")
+
+	viper.BindEnv("TRACING_PROVIDER_JAEGER_SAMPLING_VALUE")
+	viper.SetDefault("TRACING_PROVIDER_JAEGER_SAMPLING_VALUE", float64(1))
+
+	viper.BindEnv("TRACING_PROVIDER_JAEGER_LOCAL_AGENT_HOST_PORT")
+	viper.SetDefault("TRACING_PROVIDER_JAEGER_LOCAL_AGENT_HOST_PORT", "127.0.0.1:6831")
+
+	viper.BindEnv("TRACING_SERVICE_NAME")
+	viper.SetDefault("TRACING_SERVICE_NAME", "Hydra")
+
 	// If a config file is found, read it in.
 	if err := viper.ReadInConfig(); err != nil {
 		fmt.Printf(`Config file not found because "%s"`, err)
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -201,7 +201,7 @@ func initConfig() {
 	viper.SetDefault("TRACING_PROVIDER", "")
 
 	viper.BindEnv("TRACING_PROVIDER_JAEGER_SAMPLING_SERVER_URL")
-	viper.SetDefault("TRACING_PROVIDER_JAEGER_SAMPLING_SERVER_URL", "http://localhost:5778/sampling")
+	viper.SetDefault("TRACING_PROVIDER_JAEGER_SAMPLING_SERVER_URL", "")
 
 	viper.BindEnv("TRACING_PROVIDER_JAEGER_SAMPLING_TYPE")
 	viper.SetDefault("TRACING_PROVIDER_JAEGER_SAMPLING_TYPE", "const")
@@ -210,10 +210,10 @@ func initConfig() {
 	viper.SetDefault("TRACING_PROVIDER_JAEGER_SAMPLING_VALUE", float64(1))
 
 	viper.BindEnv("TRACING_PROVIDER_JAEGER_LOCAL_AGENT_HOST_PORT")
-	viper.SetDefault("TRACING_PROVIDER_JAEGER_LOCAL_AGENT_HOST_PORT", "127.0.0.1:6831")
+	viper.SetDefault("TRACING_PROVIDER_JAEGER_LOCAL_AGENT_HOST_PORT", "")
 
 	viper.BindEnv("TRACING_SERVICE_NAME")
-	viper.SetDefault("TRACING_SERVICE_NAME", "Hydra")
+	viper.SetDefault("TRACING_SERVICE_NAME", "Ory Hydra")
 
 	// If a config file is found, read it in.
 	if err := viper.ReadInConfig(); err != nil {
--- a/cmd/server/handler.go
+++ b/cmd/server/handler.go
@@ -130,6 +130,11 @@ func setup(c *config.Config, cmd *cobra.Command, args []string, name string) (ha
 	handler = NewHandler(c, w)
 	handler.RegisterRoutes(frontend, backend)
 	c.ForceHTTP, _ = cmd.Flags().GetBool("dangerous-force-http")
+	tracer := c.GetTracer()
+
+	if tracer.IsLoaded() {
+		middlewares = append(middlewares, tracer)
+	}
 
 	if !c.ForceHTTP {
 		if c.Issuer == "" {
@@ -206,6 +211,13 @@ func serve(c *config.Config, cmd *cobra.Command, handler http.Handler, address s
 		TLSConfig: &tls.Config{
 			Certificates: []tls.Certificate{cert},
 		},
+	})
+
+	srv.RegisterOnShutdown(func() {
+		tracer := c.GetTracer()
+		if tracer.IsLoaded() {
+			tracer.Close()
+		}
 	})
 
 	err := graceful.Graceful(func() error {
--- a/cmd/server/handler.go
+++ b/cmd/server/handler.go
@@ -213,7 +213,7 @@ func serve(c *config.Config, cmd *cobra.Command, handler http.Handler, address s
 		},
 	})
 
-	if tracer, _ := c.GetTracer(); tracer.IsLoaded() {
+	if tracer, err := c.GetTracer(); err == nil && tracer.IsLoaded() {
 		srv.RegisterOnShutdown(tracer.Close)
 	}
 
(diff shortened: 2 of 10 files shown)


cmd/root.go
```diff
@@ -197,6 +197,24 @@ func initConfig() {
 	viper.BindEnv("OIDC_SUBJECT_TYPE_PAIRWISE_SALT")
 	viper.SetDefault("OIDC_SUBJECT_TYPE_PAIRWISE_SALT", "public")
 
+	viper.BindEnv("TRACING_PROVIDER")
+	viper.SetDefault("TRACING_PROVIDER", "")
+
+	viper.BindEnv("TRACING_PROVIDER_JAEGER_SAMPLING_SERVER_URL")
+	viper.SetDefault("TRACING_PROVIDER_JAEGER_SAMPLING_SERVER_URL", "http://localhost:5778/sampling")
```
aeneasr: This should be empty, we can't assume that this service is running on that port and doing so could even be a security issue (it's a different service/malware not explicitly configured)

cmd/root.go
```diff
@@ -197,6 +197,24 @@ func initConfig() {
...
+
+	viper.BindEnv("TRACING_PROVIDER_JAEGER_SAMPLING_TYPE")
+	viper.SetDefault("TRACING_PROVIDER_JAEGER_SAMPLING_TYPE", "const")
+
+	viper.BindEnv("TRACING_PROVIDER_JAEGER_SAMPLING_VALUE")
+	viper.SetDefault("TRACING_PROVIDER_JAEGER_SAMPLING_VALUE", float64(1))
+
+	viper.BindEnv("TRACING_PROVIDER_JAEGER_LOCAL_AGENT_HOST_PORT")
+	viper.SetDefault("TRACING_PROVIDER_JAEGER_LOCAL_AGENT_HOST_PORT", "127.0.0.1:6831")
```
aeneasr: Same here, don't assume default hostnames!

cmd/root.go
```diff
@@ -197,6 +197,24 @@ func initConfig() {
...
+
+	viper.BindEnv("TRACING_SERVICE_NAME")
+	viper.SetDefault("TRACING_SERVICE_NAME", "Hydra")
```
aeneasr: This should be `ORY Hydra`

config/config.go
```diff
@@ -187,6 +195,28 @@ func (c *Config) GetLogger() *logrus.Logger {
 	return c.logger
 }
 
+func (c *Config) GetTracer() *tracing.Tracer {
+	if c.tracer == nil {
```
aeneasr: Are there any stats on reliability and/or network performance, specifically wrt to memory use under high load? If it's minimal we can have this always enabled. If it isn't, it should be either enabled by default (but possible to disable) or disabled by default (maybe the better option).

docker-compose.yml
```diff
@@ -80,6 +84,17 @@ services:
       - POSTGRES_PASSWORD=secret
       - POSTGRES_DB=hydra
 
+  jaeger:
+    image: jaegertracing/all-in-one:latest
```
aeneasr: The problem with `:latest` tags is that, while they do fetch the latest tag, they don't fetch those the next time you're running `:latest`. This can lead to problems where machine A uses version 1.0 and machine B uses 2.0 although both have the same tag. This can then lead to a ton of confusion and issues. It's much better (and safer) to explicitly define the version.

That being said, I don't think jaeger belongs in this docker-compose file. Instead I'd love to see either a docker command (e.g. `docker run --network --rm ..... jeagertracing/all-in-one-latest`) that you can just run "on top" of this docker-compose set up, or alternatively add this to one of the [ory/examples](https://github.com/ory/examples). If it's the latter, please coordinate with me as I'm currently working on that repository and changing some substantial things.

tracing/middleware.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	defer span.Finish()
+
+	r = r.WithContext(opentracing.ContextWithSpan(r.Context(), span))
+
+	next(rw, r)
+
+	ext.HTTPMethod.Set(span, r.Method)
+	if negroniWriter, ok := rw.(negroni.ResponseWriter); ok {
```
aeneasr: Not sure about this line, are you sure this is `negroni`? I think it would be better do define your own interface here or use one from the standard library

config/config.go
aaslamin: With the current setup, the `opentracing.GlobalTracer()` will be set to the [`NoopTracer`](https://github.com/opentracing/opentracing-go/blob/master/globaltracer.go#L4) if one does not provide any tracing config (_or if they provide invalid tracing config_) when booting up Hydra. It is disabled by default.

I use the `IsLoaded` helper later to determine as to whether to append the tracer middleware.

aeneasr: Oh I see, so omitting `TRACING_PROVIDER` disables tracing?

aaslamin: Exactly.

aeneasr: Perfect

tracing/tracer.go
```diff
@@ -0,0 +1,75 @@
...
+
+type JaegerConfig struct {
+	LocalAgentHostPort string
+	SamplerType        string
+	SamplerValue       float64
+	SamplerServerUrl   string
+}
+
+func (t *Tracer) Setup() {
+	switch t.Provider {
```
aeneasr: I think it would be good to lowercase this so `JAEGER` and `jaeger` are equivalent

tracing/tracer.go
```diff
@@ -0,0 +1,75 @@
...
+
+		if err != nil {
+			t.Logger.Warnf("Could not initialize jaeger tracer: %s", err.Error())
+			return
+		}
+
+		t.closer = closer
+		t.tracer = opentracing.GlobalTracer()
+		t.Logger.Infof("Jaeger tracer configured!")
+	default:
```
aeneasr: IMO we should have two cases here:

```
case "":
// disabled
default:
// fatal because unknown tracer
```

By explicitly failing on an unknown tracer, we make debugging much easier.

aaslamin: Good point, will update.

tracing/tracer.go
aaslamin: Good idea 👍, will update

config/config.go
aaslamin: With that said, people that configure tracing should really know what they are doing. The Google dapper paper is a great read - https://ai.google/research/pubs/pub36356

tracing/tracer.go
aaslamin: @aeneasr should it be a fatal error in this case if someone misconfigured their tracing config or a typo? 

The rest of Hydra can still run so I am not sure if we want to bring down everything. Maybe logging at error level is more appropriate?

tracing/middleware.go
aaslamin: I am sure that this is the negroni response writer because of https://github.com/urfave/negroni/blob/master/negroni.go#L96

https://github.com/urfave/negroni/blob/master/response_writer.go#L31

With that said, to be extra defensive, I capture [the boolean returned from the type assertion](https://tour.golang.org/methods/15) to verify if the `ResponseWriter` is indeed of type `negroini.ResponseWriter`. All of this is so we can add the tag in our span for the http status that was returned: 

image

config/config.go
amir-aslaminejad-hs: Added tests to cover this.

aaslamin: Added unit tests to cover above.

tracing/middleware.go
aeneasr: Ok I see how this works now. This is a negroni middleware so we're getting the correct RW here. I though it worked differently, my bad :)

cmd/server/handler.go
```diff
@@ -130,6 +130,11 @@ func setup(c *config.Config, cmd *cobra.Command, args []string, name string) (ha
 	handler = NewHandler(c, w)
 	handler.registerRoutes(frontend, backend)
 	c.ForceHTTP, _ = cmd.Flags().GetBool("dangerous-force-http")
+	tracer := c.GetTracer()
```
aeneasr: This can be simplified to:

```go
if tracer := c.GetTracer(); tracer.IsLoaded() {
  middlewares = append(middlewares, tracer)
}
```

cmd/server/handler.go
```diff
@@ -208,6 +213,13 @@ func serve(c *config.Config, cmd *cobra.Command, handler http.Handler, address s
 		},
 	})
 
+	srv.RegisterOnShutdown(func() {
+		tracer := c.GetTracer()
+		if tracer.IsLoaded() {
```
aeneasr: Same here, or depending on what `Close()` returns (if it returns nothing) maybe:

```
if tracer := c.GetTracer(); tracer.Loaded() {
  srv.RegisterOnShutdown(tracer.Close)
}
```

tracing/tracer.go
```diff
@@ -0,0 +1,76 @@
...
+				LocalAgentHostPort: t.JaegerConfig.LocalAgentHostPort,
+			},
+		}
+
+		closer, err := jc.InitGlobalTracer(
+			t.ServiceName,
+		)
+
+		if err != nil {
+			t.Logger.Errorf("Could not initialize jaeger tracer: %s", err.Error())
```
aeneasr: Should this be handled "silently", aka no service crash? I'm not sure if it should.

tracing/tracer.go
aeneasr: Yes it should, because it's an obvious mistake so hydra should make it explicit that it's misconfigured. Better to have a service not start than to discover that some feature didn't work all along that you now urgently need to debug something.

aaslamin:
Update:

• Added unit tests for the tracing middleware to ensure the expected number of spans are being created along with the desired tags.

aaslamin:
@aeneasr I am ready for another review! 🙏

cmd/server/handler.go
```diff
@@ -208,6 +213,10 @@ func serve(c *config.Config, cmd *cobra.Command, handler http.Handler, address s
 		},
 	})
 
+	if tracer, _ := c.GetTracer(); tracer.IsLoaded() {
```
aeneasr: Is it possible that there is an error here? If yes, tracer MAY be nil which will cause a panic

aaslamin: At this point, a tracer would have been configured so it would not be possible for it to have an error - recall how we now exit if an error has occurred on boot up.

Although, for safety, incase future refactoring happens by other contributors, it's probably best to do error checking. Will add it in 👍

aaslamin:
@aeneasr addressed latest feedback re: error handling 👍

aeneasr:
Thank you for your hard work!

PR #1030, opened 2018-09-12
propagate context through to the sql store

aaslamin:
With the introduction of https://github.com/ory/hydra/pull/1019 we will need to start propagating the request context down to sql store(s) so we can instrument our database calls. 

Notes:

- Creating the spans will be done in a follow-up PR. 
- Same thing will need to be done for the other handlers.

### DIFF
--- a/client/handler.go
+++ b/client/handler.go
@@ -107,7 +107,7 @@ func (h *Handler) Create(w http.ResponseWriter, r *http.Request, _ httprouter.Pa
 	}
 
 	secret := c.Secret
-	if err := h.Manager.CreateClient(&c); err != nil {
+	if err := h.Manager.CreateClient(r.Context(), &c); err != nil {
 		h.H.WriteError(w, r, err)
 		return
 	}
@@ -159,7 +159,7 @@ func (h *Handler) Update(w http.ResponseWriter, r *http.Request, ps httprouter.P
 		return
 	}
 
-	if err := h.Manager.UpdateClient(&c); err != nil {
+	if err := h.Manager.UpdateClient(r.Context(), &c); err != nil {
 		h.H.WriteError(w, r, err)
 		return
 	}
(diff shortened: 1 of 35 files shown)


aaslamin:
Unit tests pas for me locally.

aeneasr:
Would it make sense to propagate context down to other stores too?

aaslamin:
Yes, it definitely does. Would you like me to do this in this PR @aeneasr?

aaslamin:
ℹ️ Update: 

• Client manager and handler have been updated to pass along context
• All consumers have also been updated accordingly

aaslamin:
@aeneasr at this point I am a bit concerned that the PR is getting too large 😬. Could you review this as is and let me know if you want me to proceed with the Consent and JWK handlers in here or as a separate PR. 

They can be 🚀'ed independently.

aeneasr:
It's definitely a bit larger, but that's fine as there's no functional change so it's very easy to review. So far this looks good to me!

aeneasr:
Do you want to address the rest of the managers in this PR as well? You can also create another one - it doesn't really matter to me :) Maybe it's best to have the context in one commit (the one I'll squash when merging this)

aaslamin:
**Update:** 

• JWK manager and all of it's clients have been updated

**Coming up:**

• Updating the JWTStrategy to pass along context where appropriate. Hopefully this doesn't get ugly 🤞 
• Updating Consent Manager (_this is a biggie_ 😨)

aaslamin:
@aeneasr I believe the [JWTStrategy interface](https://github.com/ory/fosite/blob/master/token/jwt/jwt.go#L38-L45) in fosite needs to be updated to take in a context.

**Why?**

The implementation of this interface in Hydra makes calls to the db to fetch the latest keys:

- https://github.com/ory/hydra/blob/master/jwk/jwt_strategy.go#L75
- https://github.com/ory/hydra/blob/master/jwk/jwt_strategy.go#L83
- https://github.com/ory/hydra/blob/master/jwk/jwt_strategy.go#L91

The oauth handler consumes the Generate method.

Could you please confirm if this makes sense? If so that would be require:

- a PR in fosite to update the interface and any tests that break as a result
- a PR in Hydra to pull in the new version of fosite and update the JWTStrategy interface as necessary

aaslamin:
**Update:**

Consent Manager and all of its clients have been updated to pass along context. All that is left is the JWT strategy which is pending feedback on the comment I left above. 

@aeneasr this PR is ready for review!

aeneasr:
Ah yeah, that makes sense regarding JWT Strategy. It will also improve the API contract in fosite, which is using context almost everywhere anyways (well, except there and in bcrypt).

aeneasr:
Wow this is massive! But looks good to me so far!

aeneasr:

PR #1037, opened 2018-09-18
Allows the generation of a java (Spring RestTemplate) sdk project

pierredavidbelanger:
We talked about this in the chat: when using `swagger-codegen` to generate a Java client, it does not compile probably because of a naming issue (`JSONWebKey` vs `JsonWebKey`).

This PR only adds the lines required to generate the Java client for now so I can show what the problem exactly is.

Then I will do my best to fix it in this same PR.

After applying this patch you can reproduce de problem with those steps:

```sh
./scripts/run-gensdk.sh
cd sdk/java/hydra-client-resttemplate/
mvn compile
```

You should see those errors:

```
[ERROR] [...]/sdk/java/hydra-client-resttemplate/src/main/java/com/github/ory/hydra/model/JSONWebKeySet.java:[30,8] class JsonWebKeySet is public, should be declared in a file named JsonWebKeySet.java
[ERROR] [...]/sdk/java/hydra-client-resttemplate/src/main/java/com/github/ory/hydra/model/JSONWebKey.java:[29,8] class JsonWebKey is public, should be declared in a file named JsonWebKey.java
```


aeneasr:
I see, that's weird. I can't tell right now what's causing this, any ideas?

pierredavidbelanger:
# TL;DR

I think that if you can manage to not use a [the vendored object here](https://github.com/ory/hydra/blob/master/client/client.go#L132), it will be sufficient for swagger to not pull the uppercase version of `JSONWebKey` and `JSONWebKeySet` into the final doc.

# Explication

What I can see right now, is that definitions for `JsonWebKey` and `JsonWebKeySet` ends up being written two times in the `api.swagger.json` file.

Here is the stripped down `api.swagger.json` file that shows the double definitions:

```json
{
  ...
  "definitions": {
    "JSONWebKey": {
      "type": "object",
      "title": "JSONWebKey represents a public or private key in JWK format.",
      "properties": {...},
      "x-go-package": "github.com/ory/hydra/vendor/gopkg.in/square/go-jose.v2"
    },
    "JSONWebKeySet": {
      "type": "object",
      "title": "JSONWebKeySet represents a JWK Set object.",
      "properties": {...},
      "x-go-package": "github.com/ory/hydra/vendor/gopkg.in/square/go-jose.v2"
    },
    ...
    "jsonWebKey": {
      "type": "object",
      "properties": {...},
      "x-go-name": "swaggerJSONWebKey",
      "x-go-package": "github.com/ory/hydra/jwk"
    },
    "jsonWebKeySet": {
      "type": "object",
      "properties": {...},
      "x-go-name": "swaggerJSONWebKeySet",
      "x-go-package": "github.com/ory/hydra/jwk"
    }
    ...
  }
}
```

You can see that the uppercase versions came from the vendored objects, [JSONWebKey for example](https://godoc.org/gopkg.in/square/go-jose.v2#JSONWebKey).

The only place that the uppercase versions are referenced (then pulled in to the doc), is in the `oAuth2Client`'s `jwks` property, this property pulls on `#/definitions/JSONWebKeySet`, then in turn `JSONWebKeySet` pulls on `#/definitions/JSONWebKey`.

Here the stripped down `api.swagger.json` file that shows the `oAuth2Client` definition:

```json
{
  ...
  "definitions": {
    ...
    "oAuth2Client": {
      "type": "object",
      "title": "Client represents an OAuth 2.0 Client.",
      "properties": {
        ...
        "jwks": {
          "$ref": "#/definitions/JSONWebKeySet"
        },
        ...
      },
      "x-go-name": "Client",
      "x-go-package": "github.com/ory/hydra/client"
    },
    ...
  }
}
```

pierredavidbelanger:
Now, having said that, it will not be an easy task to get rid of `jose.JSONWebKeySet`, since it is not the same as your own `swaggerJSONWebKeySet`. Also `jose.JSONWebKeySet` end up marshalled into the DB, so changing the model here will break current installations.

I guess, it may be a bug in swagger-codegen, I see no reason why it was not able to correctly create all those model objects `JSONWebKeySet` AND `JsonWebKeySet` ... alter all they ARE different objects

pierredavidbelanger:
I am sorry, I was thinking out loud here, but no, it is not a bug in swagger-codegen. Because java is supporting OSes with case insensitive file systems, `new File("JSONWebKey.java")` and `new File("JsonWebKey.java")` point to the same file (the first that was created)..

So the explication here is simple: swagger-codegen first generate the model file for the class `JSONWebKey` in a file named `JSONWebKey.java`, then it encounters `JsonWebKey`, so it opens the same old `JSONWebKey.java` file and generate the `JsonWebKey` model class in it.

That explains this:

```
JSONWebKey.java: class JsonWebKey is public, should be declared in a file named JsonWebKey.java
```

So, I guess it is not a good idea to have two model objects with the same name (case insensitive).

pierredavidbelanger:
Ok I am starting to understand, you created `type swaggerJSONWebKey`, to be able to expose all properties of the private jose `type rawJSONWebKey` in the doc... 

I guess there is a way to tell go-swagger that `Client`'s `JSONWebKeys` property must be exposed in the doc as a `swaggerJSONWebKeySet` and not as a `jose.JSONWebKeySet`, but keeping the internal representation/implementation.

aeneasr:
Ah I see, thank you for the traceback! Do you think you could supply a fix for it?

pierredavidbelanger:
I think I have something.

I guess it will need a little cleanup before merging (I dont like having two public struct in `jwk/doc.go` for exemple), obviously only if you want to merge :)

I pushed the fix in two commits, one that touches the code only (to help you find what I have modified), then an other one for everything that changed after running `./scripts/run-gensdk.sh`.

Essentially this fix remove `*jose.JSONWebKeySet` from the `Client` (and thus from the surface API, cleaning at the same time a BUNCH of unused model objects), replacing it with your `swaggerJSONWebKeySet` that I renamed `JSONWebKeySet`.

Every other code changes I made are there to make this happen.

after this patch, `go test ./...` is all `ok` ( if you also apply #1039 ;) )

aeneasr:
Another possibility is to introduce swaggerOAuth2Client which is basically 1:1 the existing client definition but with the jwk structs. This would remove the need for the JSON dance and yield the same result

pierredavidbelanger:
You are absolutely right. I will try this.

pierredavidbelanger:
Oh, it's harder than I thought, `client.Client` is referenced else where in the API, here in the [consent.AuthenticationRequest](https://github.com/ory/hydra/blob/master/consent/types.go#L249) and here in the [consent. ConsentRequest](https://github.com/ory/hydra/blob/master/consent/types.go#L292). 

Those two structs are not for the doc purpose only .. they are use in places hard to refactor, for example here in [consent.GetConsentRequest](https://github.com/ory/hydra/blob/master/consent/manager_sql.go#L218).

I will think about this a little.

aeneasr:
Ah damn, I didn't think about that. The problem I have with this patch is really the JSON marshalling in the get function. I was thinking if we could construct `jose.JSONWebKey` without using marshal but I don't really think so. Maybe we could try to modify the method and use a `byte.Buffer? This would make it a bit more readable:

```go
func (c *Client) GetJSONWebKeys() (*jose.JSONWebKeySet) {
	if c.JSONWebKeys == nil {
		return nil
	}
        b := bytes.NewBuffer()
        if err := json.NewEncoder(b).Encode(c.JSONWebKeys); err != nil {	
		return nil
	}
	var j jose.JSONWebKeySet
	if err := json.NewDecoder(b).Decode(&j) err != nil {
		return nil
	}
	return j
}
```

It's still a long stretch from "readable" but at least it's a bit shorter. I would still like to see a better option here though...

aeneasr:
Another issue I see though is that the meaning of `c.JSONWebKeys` changed. I don't think it's being used somewhere else right now but we'll have to make sure to use `GetJSONWebKeys` at all times when we're expecting the jose version. This is hard...

pierredavidbelanger:
The PR #1045 contains an alternative (more hacky) way of dealing with the same problem.

I will let you decide which one you prefer, or I you want to fix this at all :)

aeneasr:
Closed in favor of #1045

PR #1042, opened 2018-09-20
Bump fosite version and integrate breaking changes

aaslamin:
Bumps fosite version to 0.22.0: [changes between current (0.21.5 and 0.22.0)](https://github.com/ory/fosite/compare/v0.21.5...v0.22.0)

As seen in the diff, the breaking change was to the `JWTStrategy` interface as its methods now accept a context. Handlers have been updated as a result of this to pass along the request context down the call path.

_Review: @aeneasr_

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -351,6 +351,7 @@
     ".",
     "ext",
     "log",
+    "mocktracer",
   ]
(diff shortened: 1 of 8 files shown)


aeneasr:
Thank you!

PR #1044, opened 2018-09-21
client: fix test to pass non-nil context

someone1:


someone1:
When updating my plugin to work with the latest context fixes (thanks @aaslamin !) I was getting panics when running tests as a nil context was being used in parts of the test.

aeneasr:
Thanks!

PR #1045, opened 2018-09-23
Fix swagger

pierredavidbelanger:
Here is an other try at fixing #1037 . , but this time with minimal impact.

This PR has a minimal impact on the code (zero impact in fact), but it uses a hack.
Indeed, it plays on the fact that, in the `swagger generate spec` phase, two model object can have the same name, but only the last one will be written , in this case, its the one we want `jwk.swaggerJSONWebKeySet`.


aeneasr:
Haha, nice find! This will be a tough decision

aeneasr:
After some deliberation, I like this one much more! One thing though, could you document [here](https://github.com/ory/hydra/pull/1045/files#diff-a769d53cc015aef87a3c2ef2cea0fe4aR113) and [here](https://github.com/ory/hydra/pull/1045/files#diff-a769d53cc015aef87a3c2ef2cea0fe4aR123) why it is important that these stay uppercase? This helps preventing regression issues in the future!

PR #1051, opened 2018-09-25
consent: migrate to test helpers [closes #1043]

### LINKED ISSUE #1043: [Feature Request] Update consent tests to match oauth2/client tests
As I develop my backend plugin, its become extremely valuable that I can reuse the same tests hydra uses to validate its own SQL backend implementations.

However, the consent package tests don't use exported test helper functions like the oauth2 and client packages do, so I can't just import and use them to test my plugin's implementation of the corresponding managers.

Would you be open to switching up the consent package tests to match the others? I can try and put a PR together for it but wanted to run it by here first. I'd basically move the contents of `TestManagers` out to a manager_test_helpers.go file and create a `TestManager` function accordingly.

Thanks!

someone1:


someone1:
I had to create a `testhelpers` subpackage to avoid cyclic imports (`consent` tests rely on the `oauth2` package and vice-versa)

aeneasr:
I'm confused, you only changed code in `consent` and did not change any package definitions (except for the new one). What caused the cyclic dependencies? Would it not have been enough to just Upper-case the required methods? Packages like `testhelpers` are an [anti-pattern](https://blog.golang.org/package-names) in go.

someone1:
Yes I realize that, but since the tests found in the package use the package name `consent_test` (black-box testing) I could not break out the tests without introducing cyclic imports unless I placed it in a separate package. I realize I used an anti-pattern here, though I'm not sure of a proper fix which wouldn't introduce code changes all around hydra. I think the only reason you can get away with a different package name for the test files than of the package itself is due to the `_test.go` file name prefix which the compiler ignores when building the package and running tests.

Other ideas:

- Break out each inner `t.Run` as a separate helper function, and leave oauth2 dependent tests separate (seems counter to the goal of the PR)
- Move interface definitions for the various managers to a centralized location (`config` package? also introduce an interface for the FositeStore?) - this would be a lot of work and touch a lot of places of hydra

someone1:
the interface find in the `pkg` package fixes all this! Sorry for the confusion, I'll update the tests to deal without access to the direct object but only access via the interface.

consent/manager_test_helpers.go
```diff
@@ -0,0 +1,513 @@
...
+					ids:     []string{"challengerv1"},
+				},
+				{
+					at: "trva2", rt: "rrva2",
+					subject: "subjectrv2",
+					client:  "clientrv2",
+					ids:     []string{"challengerv2"},
+				},
+			} {
+				t.Run(fmt.Sprintf("case=%d/subject=%s", i, tc.subject), func(t *testing.T) {
```
someone1: This is the only test that was modified. Since we don't have access to the actual fosite memory manager anymore, we rely on the interface access methods to validate the Access Tokens/Refresh Tokens are present or not

aeneasr:
Nice, thanks!

PR #1069, opened 2018-10-03
Add missing indices to resolve #1067

### LINKED ISSUE #1067: Missing database indices
#### `request_id` column in the `hydra_oauth2_access` & `hydra_oauth2_refresh` tables

When hitting the token handler to refresh tokens, the following query is eventually made to both the access token and refresh token table:

```
DELETE FROM hydra_oauth2_<access|refresh> WHERE request_id=???
```

This can result in an outage as there is no index on the `request_id` column. To adhere to the ACID properties, the database needs to lock all the rows that are scanned as a result of this operation. In the absolute worst case, this can result in a full table scan to find the row to delete which will effectively lock the entire table 💥. By adding an index, the db can lock the individual row instead.

#### `requested_at` column in the `hydra_oauth2_access` table

When flushing expired tokens, the following query is made to the access token table:

```
DELETE FROM hydra_oauth2_access WHERE requested_at < ??? AND requested_at < ???
```

This can lead to disaster for the same reasoning as above.

#### Error that occurs when a table is locked and another txn (e.g. delete/update) is attempted
```
Error 1205: Lock wait timeout exceeded; try restarting transaction
```

## Proposal

Create missing indices and add migration guide.

aaslamin:
Resolves #1067 by adding indices to:

• `request_id` column in the hydra_oauth2_access & hydra_oauth2_refresh tables
• `requested_at` column in the hydra_oauth2_access table

_Review: @aeneasr_

### DIFF
--- a/oauth2/fosite_store_sql.go
+++ b/oauth2/fosite_store_sql.go
@@ -87,6 +87,8 @@ func sqlSchemaUp(table string, id string) string {
 	subject 		varchar(255) NOT NULL
 )`,
 		"4": fmt.Sprintf("ALTER TABLE hydra_oauth2_%s ADD active BOOL NOT NULL DEFAULT TRUE", table),
+		"5": fmt.Sprintf("CREATE UNIQUE INDEX hydra_oauth2_%s_request_id_idx ON hydra_oauth2_%s (request_id)", table, table),
+		"6": fmt.Sprintf("CREATE INDEX hydra_oauth2_%s_requested_at_idx ON hydra_oauth2_%s (requested_at)", table, table),
 	}
 
 	return schemas[id]
@@ -98,6 +100,8 @@ func sqlSchemaDown(table string, id string) string {
 		"2": fmt.Sprintf("ALTER TABLE hydra_oauth2_%s DROP COLUMN subject", table),
 		"3": "DROP TABLE hydra_oauth2_pkce",
 		"4": fmt.Sprintf("ALTER TABLE hydra_oauth2_%s DROP COLUMN active", table),
+		"5": fmt.Sprintf("DROP INDEX hydra_oauth2_%s_request_id_idx ON hydra_oauth2_%s", table, table),
+		"6": fmt.Sprintf("DROP INDEX hydra_oauth2_%s_requested_at_idx ON hydra_oauth2_%s", table, table),
 	}
 
 	return schemas[id]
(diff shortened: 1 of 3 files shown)


aaslamin:
I wasn't sure if you had a naming convention for indices. Please let me know if you would like me to update the names.

To test the migrations, I applied them on both postgres and mysql using the supplied docker-compose setup.

oauth2/fosite_store_sql.go
```diff
@@ -87,6 +87,8 @@ func sqlSchemaUp(table string, id string) string {
 	subject 		varchar(255) NOT NULL
 )`,
 		"4": fmt.Sprintf("ALTER TABLE hydra_oauth2_%s ADD active BOOL NOT NULL DEFAULT TRUE", table),
+		"5": fmt.Sprintf("CREATE INDEX hydra_oauth2_%s_request_id_idx ON hydra_oauth2_%s (request_id)", table, table),
```
aeneasr: I think it would make sense to have an unique index. They key should really be unique and it will help with query speed.

oauth2/fosite_store_sql.go
```diff
@@ -169,6 +173,26 @@ var migrations = &migrate.MemoryMigrationSource{
...
+				sqlSchemaUp(sqlTableAccess, "5"),
+				sqlSchemaUp(sqlTableRefresh, "5"),
+			},
+			Down: []string{
+				sqlSchemaDown(sqlTableAccess, "5"),
+				sqlSchemaDown(sqlTableRefresh, "5"),
+			},
+		},
+		{
+			Id: "6",
```
aeneasr: These two can be merged, no?

aeneasr:
Please also add a [test](https://github.com/ory/hydra/blob/master/client/manager_0_sql_migrations_test.go). Since you didn't change the schema it's probably enough to re-enter the previous data again but with a new ID.

oauth2/fosite_store_sql.go
aaslamin: @aeneasr I had to keep them separate because I only wanted to create an index on the `requested_at` column on the `hydra_oauth2_access` table. 

I guess I could add a semicolon and add another statement for migration 5?

oauth2/fosite_store_sql.go
aaslamin: Nice catch 👍

aaslamin:
Thanks for the feedback! Will address in the morning 🙌

oauth2/fosite_store_sql.go
aeneasr: Oh yeah, I see, this is because of the function design. Let's keep it the way you have it

aaslamin:
#### Update:

• Changed index to a unique index on the `request_id` column
• Investigating failing tests as a result of that change now

aaslamin:
#### Update

• Fixed broken test - it was attempting to insert a record with the same `request_id`. The system works 😎

aeneasr:
Nice! I'll review that tomorrow!

aaslamin:
#### Update:

• Added new test coverage to cover the unique constraints placed on the `request_id` column in the `hydra_oauth2_access` & `hydra_oauth2_refresh` tables.

This PR is ready for potentially a final review @aeneasr  🤝

aeneasr:
Thank you!

oauth2/fosite_store_sql.go
aeneasr: lmao 2 years later and I found that this new unique index actually breaks OpenID Connect compliance...welp 🤷

PR #1071, opened 2018-10-04
tracing: use context aware database methods

aaslamin:
This PR continues on our efforts of context propagation by using the associated context aware methods in the sqlx package.

This is necessary for when we [wrap the db driver](https://github.com/golang/go/issues/18080#issuecomment-287867357) with the tracing API, we will start getting spans out of the box 😉.  

Furthermore, using these methods comes with the added benefit that database drivers that support context cancellation will be able to quit early if there is no hope of completing a query/tx.

_Review: @aeneasr_

### DIFF
--- a/consent/manager_sql.go
+++ b/consent/manager_sql.go
@@ -77,7 +77,7 @@ func (m *SQLManager) revokeConsentSession(ctx context.Context, user, client stri
 	}
 
 	var challenges = make([]string, 0)
-	if err := m.db.Select(&challenges, m.db.Rebind(fmt.Sprintf(
+	if err := m.db.SelectContext(ctx, &challenges, m.db.Rebind(fmt.Sprintf(
 		`SELECT r.challenge FROM hydra_oauth2_consent_request_handled as h 
 JOIN hydra_oauth2_consent_request as r ON r.challenge = h.challenge WHERE %s`,
 		part,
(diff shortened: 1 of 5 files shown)


consent/manager_sql.go
```diff
@@ -153,12 +154,13 @@ func (m *SQLManager) RevokeUserAuthenticationSession(ctx context.Context, user s
...
 
 func (m *SQLManager) CreateForcedObfuscatedAuthenticationSession(ctx context.Context, s *ForcedObfuscatedAuthenticationSession) error {
-	tx, err := m.db.Beginx()
+	tx, err := m.db.BeginTxx(ctx, nil)
 	if err != nil {
 		return sqlcon.HandleError(err)
 	}
 
-	if _, err := m.db.Exec(
+	if _, err := tx.ExecContext(
```
aaslamin: This was a bug I believe. I fixed it since I was knee deep in the code at this point already.

` m.db.Exec` is not going to be done in the TX created above.

consent/manager_sql.go
```diff
@@ -169,7 +171,7 @@ func (m *SQLManager) CreateForcedObfuscatedAuthenticationSession(ctx context.Con
 		return sqlcon.HandleError(err)
 	}
 
-	if _, err := m.db.NamedExec(
+	if _, err := tx.NamedExec(
```
aaslamin: Same here 🐛

aeneasr:
Perfect!

PR #1076, opened 2018-10-08
cmd: Fix flaky port finder

### LINKED ISSUE #1054: Fix flaky random-port generator
The port finder fails sometimes because it uses the same port twice for backend and frontend ports.

aeneasr:
Closes #1054


aaslamin:
Nice! Thanks for fixing these @aeneasr! 

💯 🏆

PR #1087, opened 2018-10-11
Add Gopkg for dep

TusharSheth:
Signed-off-by: Tushar Sheth <tushar.sheth13@gmail.com>


aeneasr:
We switched to [go modules](https://github.com/golang/go/wiki/Modules), dep is no longer supported.

Issue #1079, opened 2018-10-09
Support Kubernetes Secrets


keirl:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
Database URLs include username and password as part of URL.  When constructing Kubernetes deployments this encourages plaintext username and passwords and does not support Kubernetes secrets for DB credentials.

**What is the desired behavior?**
Allow DB_USERNAME and DB_PASSWORD environmental variables that can be set.  This would allow you to use Kubernetes deployments that reference Kubernetes secrets.

Such as 
env:
      - name: DB_USERNAME
        valueFrom:
          secretKeyRef:
            name: mysecret
            key: username
      - name: DB_PASSWORD
        valueFrom:
          secretKeyRef:
            name: mysecret
            key: password

Reference: https://kubernetes.io/docs/concepts/configuration/secret/

**Which version of the software is affected?**
1.0.0-beta8

aeneasr:
You could just store the database url in the secret. That's one less secret and env var you have to configure.

keirl:
@aeneasr The entrypoint in the dockerfile does not simply support variable substitution.  We plan look at overriding the entrypoint to do the variable substitution in the container, but it will be a hack and take us further from the default use cases.

fredbi:
IMHO, this request for enhancement unduly puts additional burden on hydra's configuration tasks.

It is rather straightforward to configure Postgres and avoid any disclosure of credentials in the connection string by using password files. You are then free to use K8 secrets (or in my case, just docker secrets) to store these files.

Example:
```yaml
    - PGPASSFILE=/var/certs/pgpass
    - DATABASE_URL=postgres://dbuser@postgres:5432/accesscontroldb?sslmode=verify-full&sslrootcert=/var/certs/ca.crt
```  

And your pgpass file contains the credentials for your user.

fredbi:
> The entrypoint in the dockerfile does not simply support variable substitution. We plan look at overriding the entrypoint to do the variable substitution in the container, but it will be a hack and take us further from the default use cases.

Yes you can substitute variables at build time: `docker build -e DB_USER=dbuser ...` 
And in `Dockerfile`:

```
ENV DATABASE_URL=postgres://$DB_USER:$DB_PASSWORD@$DB_HOST:$DB_PORT
```

keirl:
[...]

PR #1089, opened 2018-10-11
cmd: configuration for secrets

fredbi:
* Added SYSTEM_SECRET_PATH and COOKIE_SECRET_PATH env variables to
  retrieve secrets from a file rather than env
* Enforced COOKIE_SECRET to be at least 256 bits long (symmetric to SYSTEM_SECRET)

Signed-off-by: Frederic BIDON <frederic@oneconcern.com>

**NOTE**: this PR allows to configure hydra secrets without exposing initialization secrets in environment and/or configuration files.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -103,6 +103,16 @@ before finalizing the upgrade process.
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
+## master
+
+### Configuration changes
+
+New environment variables `SYSTEM_SECRET_PATH` and `COOKIE_SECRET_PATH`
+allow to retrieve secrets from files, and avoid exposing secrets in environment variables.
+
+These new variables supplement `SYSTEM_SECRET` and `COOKIE_SECRET`
+respectively, and do not affect existing configurations.
+
 ## 1.0.0-beta.10
 
 ### Schema Changes
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -22,8 +22,10 @@ package cmd
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strings"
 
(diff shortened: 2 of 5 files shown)


aeneasr:
Why would you store secrets in an unencrypted format on disk? Platforms are getting better at hiding env vars marked as secrets from the shell. I don't really see the upside to this.

fredbi:
Same as TLS certs, keys: all crypto material should be configurable either directly from env of from ext file (which may then be a secret on your favorite container platform).

This offers more possibilities to build configuration.

Finally, even if platforms are getting better, it is still a lot of env hack to do for every such parameter.

fredbi:
BTW, the CI failure is really strange, given the limited scope of my changes.
I suspect another failure:
```
#!/bin/bash -eo pipefail
./scripts/test-sdk.sh

/go/src/github.com/ory/hydra/cmd/cli/handler_migrate.go:26:2: could not import github.com/ory/x/cmdx (cannot find package "github.com/ory/x/cmdx" in any of:
	/go/src/github.com/ory/hydra/vendor/github.com/ory/x/cmdx (vendor tree)
	/usr/local/go/src/github.com/ory/x/cmdx (from $GOROOT)
	/go/src/github.com/ory/x/cmdx (from $GOPATH))
couldn't load packages due to errors: github.com/ory/hydra/cmd/cli
Exited with code 1
```

fredbi:
The difference with #1079  is that in this particular issue, Postgres provides everything to build your
configuration the way you want. In this PR, I am addressing the same kind of issue, but for hydra-specific parameters.

aeneasr:
I'm not sure about this. Every platform supports secret management with environment variables, is there a specific issue you're addressing with this, like - say - secret management doesn't work with env vars on AWS ECS?

Regarding HTTPS certificates I agree, it is sort of have to allow it for one thing but not for the other. But then again, TLS Certificates are lengthy and a bit tricky to add as env vars (because of newlines) whereas those secrets are just 32 byte strings.

aeneasr:
Actually, I stand more firmly by my point. If we were to indeed support config file configuration like you're suggesting it, we must put more effort into it and let everything be configurable using the filesystem. I don't really want to go there at the moment because it opens another bag of questions and problems (like reloading config, config from db). Adding more and more env config does not make this product more usable but harder to set up. We should try to reduce the amount of things you can use and change and update and have one coherent, streamlined approach to configuration. I'm generally open to file configuration because I think it works great with CI, but it needs to be properly planned and all the things we want in the future should be supported.

fredbi:
Here is my source of inspiration. This is a Postgres init script for docker: 

https://github.com/docker-library/postgres/blob/eed67ed0f33435bffd1e78d27b389e0492d30599/9.6/docker-entrypoint.sh#L7

aeneasr:
That's of course another option, writing a bash script/entrypoints that load env vars from files.

aeneasr:
Closing in favor of #1108

PR #1105, opened 2018-10-22
Fix missing LoginChallenge and LoginSessionID from GetConsentRequest

jcxplorer:
I believe this was missing from #1013. Responses from `GET /oauth2/auth/requests/consent/:challenge` have the `login_challenge` and `login_session_id` fields, but they're always empty strings, because they were not being converted from the database response.

This PR adds the missing fields to the relevant SQL converter test and implementation.

cc @aeneasr

### DIFF
--- a/consent/manager_test_helpers.go
+++ b/consent/manager_test_helpers.go
@@ -508,6 +508,7 @@ func compareAuthenticationRequest(t *testing.T, a, b *AuthenticationRequest) {
 	assert.EqualValues(t, a.RequestURL, b.RequestURL)
 	assert.EqualValues(t, a.CSRF, b.CSRF)
 	assert.EqualValues(t, a.Skip, b.Skip)
+	assert.EqualValues(t, a.SessionID, b.SessionID)
 }
 
 func compareConsentRequest(t *testing.T, a, b *ConsentRequest) {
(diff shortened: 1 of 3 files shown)


aeneasr:
Yup, that looks definitely like a bug! Can you please add a test for this as well? I think one place to do that is [here](https://github.com/ory/hydra/blob/master/consent/manager_test_helpers.go#L512) and another one [in this file](https://github.com/ory/hydra/blob/master/consent/sql_helper_test.go). Thank you!

aeneasr:
Oh sorry, I missed that you added it to the latter file already. Adding it to the first one *should* assure that the behaviour is correct across all managers.

aeneasr:
Because another PR was merged, there's a small conflict now which has to be addressed as well.

jcxplorer:
Thanks for looking at this so quickly, and for the pointers!

I've rebased, and also added the LoginChallenge assertion to compareConsentRequest. However, adding an assertion for LoginSessionID causes a bunch of test failures. I know that a login session ID is only provided in certain cases, but unsure of how to add it to the test. Any idea how to proceed?

aeneasr:
Hm, that's interesting, the session ID should [always be set](https://github.com/ory/hydra/blob/master/consent/manager_test_helpers.go#L103). Maybe it hints to another error?

aeneasr:
Yeah awesome, good thing the other test was added too :) Thank you!

jcxplorer:
And thank you @aeneasr for handling this so quickly. This feature makes Hydra really flexible for us. :+1:

PR #1115, opened 2018-10-24
tracing: add support for tracing db interactions

aaslamin:
This PR adds support for creating spans around DB interactions by leveraging our effort in https://github.com/ory/x/pull/3 (_which utilizes the library created by @luna-duclos to decorate the interfaces exposed by the `database/sql` package with the opentracing API_).

#### Things of interest:

- An instrumented driver will only be used IFF tracing has been configured
  - Perhaps we should add a config specifically for creating spans around DB interactions incase someone does not care about such spans? 
- ~Configuration added to omit SQL args from being added to spans~
- New test coverage to exercise new tracing options

#### Example trace from making a bad call to `POST /clients`

image
---

_Review: @aeneasr_


cmd/root.go
```diff
@@ -216,6 +216,9 @@ func initConfig() {
 	viper.BindEnv("TRACING_SERVICE_NAME")
 	viper.SetDefault("TRACING_SERVICE_NAME", "Ory Hydra")
 
+	viper.BindEnv("TRACING_OMIT_SQL_ARGUMENTS_FROM_SPANS")
```
aeneasr: Please document this around here: https://github.com/ory/hydra/blob/master/cmd/serve.go#L182

And while you're add it, please document the other tracing env vars there too :)

config/backend_connector.go
```diff
@@ -41,7 +41,7 @@ var (
 )
 
 type BackendConnector interface {
-	Init(url string, l logrus.FieldLogger) error
+	Init(url string, l logrus.FieldLogger, opts ...ConnectorOptions) error
```
aeneasr: ping @someone1 fyi

config/connector_options.go
```diff
@@ -0,0 +1,38 @@
+package config
+
+type ConnectorOptions func(*options)
+
+type options struct {
```
aeneasr: would it make sense to reuse those from `sqlcon`?

aeneasr: That makes sense to me

aaslamin:
Thanks for the feedback @aeneasr @someone1, looking into them now 🙌

config/connector_options.go
aaslamin: Re: `OmitSQLArgsFromTracingSpans` I was following some of the concerns that Google identified in [Dapper](https://ai.google/research/pubs/pub36356) for not releasing potentially sensitive details in traces. Perhaps in some organizations, the engineers that have access to look at traces should not be able to view such data for compliance/privacy reasons.

I can remove this feature until we get a feature request, maybe?

someone1: My apologies for not being clear on that - I meant that traces should always omit the data. For general hydra use-cases, I don't think there's a time this is needed outside of debugging/testing.

aaslamin: Sweet, that sounds good to me. I will leave this option on by default and remove the config and if people ask for it be included, they can create an issue so we can re-introduce it.

config/backend_connector.go
aaslamin: Should be exported now. I was completely unaware of the plugin use case, exported options should be generic enough for any type of connector, not just SQL (e.g. redis, mongo, etc).

config/config.go
```diff
@@ -317,22 +327,20 @@ func (c *Config) Context() *Context {
 	}
 
 	if backend, ok := backends[scheme]; ok {
-		if err := backend.Init(c.DatabaseURL, c.GetLogger()); err != nil {
+		options := []ConnectorOptions{}
+
+		if tracingEnabled {
+			options = append(options, WithTracing(), withOmitSQLArgsFromSpans())
```
aaslamin: I could have done something like `os.Getenv("OMIT_SQL_ARGS_FROM_TRACES")` to give _advanced users_ the option, but I felt dirty doing it? ¯\_(ツ)_/¯

config/connector_options.go
```diff
@@ -0,0 +1,40 @@
+package config
+
+type ConnectorOptions func(*Options)
+
+type Options struct {
```
aaslamin: 👀

config/connector_options.go
```diff
@@ -0,0 +1,40 @@
...
+	UseTracing bool
+
+	// unexported as these are specifically used by Hydra for writing tests
```
aaslamin: 👀

config/config.go
aeneasr: Yeah, let's just keep it as is. If people complain we can add it or they have to re-build the binary which hopefully makes them incredibly aware of removing that once they're done.

config/connector_options.go
aeneasr: This makes a ton of sense to me, all of these options are only used internally for testing and other stuff. Good job!

aaslamin: Yes, that was completely my bad. I was not aware of the plugin use case. It makes sense to me 👍

config/config.go
```diff
@@ -294,6 +294,16 @@ func (c *Config) Context() *Context {
...
 	var hasher fosite.Hasher
+	var tracingEnabled bool
+
+	hasher = &fosite.BCrypt{
+		WorkFactor: c.BCryptWorkFactor,
+	}
+
+	if tracer, err := c.GetTracer(); err == nil && tracer.IsLoaded() {
+		tracingEnabled = true
+		hasher = &tracing.TracedBCrypt{c.BCryptWorkFactor}
```
aeneasr: Please use keyed fields here

aeneasr: I also noticed that this is kinda duplicated with [here](https://github.com/ory/hydra/blob/master/cmd/server/handler_oauth2_factory.go#L120-L122). I also noticed that [`hasher`](https://github.com/ory/hydra/blob/master/cmd/server/handler_oauth2_factory.go#L56) might be null if [tracing is disabled](https://github.com/ory/hydra/blob/master/cmd/server/handler_oauth2_factory.go#L120-L122). So instead this should use the bcrypt from the config (here)

aeneasr: The nil thing is actually default behavior and did not cause any issues. `BCryptWorkFactor` was also set properly.

aeneasr: I think for clarity it should we should use the hasher from the config [here](https://github.com/ory/hydra/blob/master/cmd/server/handler_oauth2_factory.go#L120-L122).

aaslamin: > I also noticed that this is kinda duplicated with [here](https://github.com/ory/hydra/blob/master/cmd/server/handler_oauth2_factory.go#L120-L122). I also noticed that [`hasher`](https://github.com/ory/hydra/blob/master/cmd/server/handler_oauth2_factory.go#L56) might be null if [tracing is disabled](https://github.com/ory/hydra/blob/master/cmd/server/handler_oauth2_factory.go#L120-L122). So instead this should use the bcrypt from the config (here)

If Hasher is `nil` it gets set by fosite when Compose is called https://github.com/ory/fosite/blob/master/compose/compose.go#L56

aaslamin: 👍 on DRYing up the check to see if tracing has been loaded tho. I will clean that up.

someone1:
LGTM

Great job!

tracing/tracer.go
```diff
@@ -76,3 +76,43 @@ func (t *Tracer) Close() {
...
+- TRACING_PROVIDER_JAEGER_SAMPLING_VALUE: The value passed to the sampler type that has been configured.
+
+	Supported values: This is dependant on the sampling strategy used:
+		- const: 0 or 1 (all or nothing)
+		- rateLimiting: a constant rate (e.g. setting this to 3 will sample requests with the rate of 3 traces per second)
+		- probabilistic: a value between 0..1
+
+	Example: TRACING_PROVIDER_JAEGER_SAMPLING_VALUE=1
+
+- TRACING_PROVIDER_JAEGER_LOCAL_AGENT_HOST_PORT: The address of the jaeger-agent where spans should be sent to
```
aeneasr: Let's call this `TRACING_PROVIDER_JAEGER_LOCAL_AGENT_ADDRESS`

tracing/tracer.go
```diff
@@ -76,3 +76,43 @@ func (t *Tracer) Close() {
...
+
+	Example: TRACING_PROVIDER_JAEGER_LOCAL_AGENT_HOST_PORT=127.0.0.1:6831
+
+- TRACING_SERVICE_NAME: Specifies the service name to use on the tracer.
+
+	Default: Ory Hydra
```
aeneasr: `Ory` should be uppercase: `ORY`

aeneasr:
Thank you for your hard work!

PR #1136, opened 2018-10-29
Minor bug fix in JWK sql migrations test case

### LINKED ISSUE #1135: Copy-paste error in manager_0_sql_migrations_test.go
**Describe the bug**
Not a very big deal, but there is an error in "4-data" set: inserting row with id "4-sid" and deleting another, with id "3-sid". I suppose this can fail the test if run twice in a row, because a corresponding key from the previous run would still be there.

**Expected behavior**
Delete the same row.

**Screenshots**
https://github.com/ory/hydra/blob/master/jwk/manager_0_sql_migrations_test.go#L71
```	{
		Id: "4-data",
		Up: []string{
			`INSERT INTO hydra_jwk (sid, kid, version, keydata, created_at) VALUES ('4-sid', '4-kid', 0, 'some-key', NOW())`,
		},
		Down: []string{
			`DELETE FROM hydra_jwk WHERE sid='3-sid'`,
		},
	},
```

**Version:**
 - master

jacor84:
Fixes #1135

### DIFF
--- a/jwk/manager_0_sql_migrations_test.go
+++ b/jwk/manager_0_sql_migrations_test.go
@@ -68,7 +68,7 @@ var createJWKMigrations = []*migrate.Migration{
 			`INSERT INTO hydra_jwk (sid, kid, version, keydata, created_at) VALUES ('4-sid', '4-kid', 0, 'some-key', NOW())`,
 		},
 		Down: []string{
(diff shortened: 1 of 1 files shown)


aeneasr:
thank you!

PR #1158, opened 2018-11-05
check length of array before get[0], split replaced by splitx

decoomanj:
## Related issue

As discussed in the Discord channel, the array SubjectTypes is queried without checking for the length. When it is empty, the values [0] is selected which leads to a panic.

In the same vein: when the SubjectTypesSupported is empty, the split returns a length of 1 which is incorrect.

## Proposed changes

Check the length of the array before selecting element [0].

Switch to stringsx.Splitx() for splitting the array.


aeneasr:
Could you please add tests for this, otherwise regressions may occur.

decoomanj:
will do

aeneasr:
ping :)

decoomanj:
hang on :-)

aeneasr:
Superseded by #1314

PR #1239, opened 2018-12-15
fixes #1238 bump go to 1.11.3 (CVE-2018-16875)

### LINKED ISSUE #1238: Security: bump Golang version to 1.11.3 (CVE-2018-16875)
https://nvd.nist.gov/vuln/detail/CVE-2018-16875

https://groups.google.com/forum/#!msg/golang-announce/Kw31K8G7Fi0/z2olKn-QCAAJ

aaslamin:
FYI: **this version of go introduces a known regression** which will be addressed in the next patch release. I guess they were moving fast to address the CVE.

> We are aware of a functionality regression in "go get" when executed in GOPATH mode on an import path pattern containing "..." (e.g., "go get github.com/golang/pkg/..."), when downloading packages not already present in the GOPATH workspace. This is issue golang.org/issue/29241. It will be resolved in the next minor patch releases, Go 1.11.4 and Go 1.10.7, which we plan to release soon. We apologize for any disruption.
https://groups.google.com/forum/#!msg/golang-announce/Kw31K8G7Fi0/z2olKn-QCAAJ

👀 @aeneasr

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.11-alpine
+FROM golang:1.11.3-alpine
 
 ARG git_tag
 ARG git_commit
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,20 +5,20 @@ version: 2
 jobs:
   format:
     docker:
-      - image: circleci/golang:1.11
+      - image: circleci/golang:1.11.3
     working_directory: /go/src/github.com/ory/hydra
     steps:
 # This is disabled for now because goimports is really slow when go modules are used, see
 # https://github.com/golang/go/issues/27287
 #
 #      - run:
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.11-alpine
+FROM golang:1.11.3-alpine
```
aeneasr: I was under the impression that `go1.11` refers to the latest go 1.11.x tag. Actually I just double checked and that is indeed the case. The question being that, if `go1.11` is cached in docker already, it is updated or not. In that case, making the dependency explicit would be an option. Alternatively we could run `git pull golang:1.11` (or whatever version) to force an update to the latest tag.

aeneasr:
Closed in favor of #1243

Issue #1247, opened 2018-12-23
Support fosite delegated transactions in SQL storage


aaslamin:
**Is your feature request related to a problem? Please describe.**

Certain oauth flows require transactional semantics (_see issue linked below_). Thanks to Fosite v0.29.0, Fosite will now attempt to initiate a transaction if the underlying storage provider has implemented the [`Transactional` interface](https://github.com/ory/fosite/blob/master/storage/transactional.go#L17).

**Describe the solution you'd like**

- Bump fosite to at least `v0.29.0`
- Extend `FositeSQLStore` so that it implements `Transactional`
- Modify all sql storage methods to first lookup for a transaction in the context (_if one does not exist, its business as usual_).
- Test coverage of new behaviour

**Additional context**

https://github.com/ory/fosite/issues/309
https://github.com/ory/fosite/pull/343

aeneasr:
👍

aaslamin:
My colleague, @michaelwagler, has decided to take this issue on for his first contribution to Hydra 🏆

michaelwagler:
Yup 👍

PR #1277, opened 2019-02-05
addresses #1247

michaelwagler:
https://github.com/ory/hydra/issues/1247


michaelwagler:
@aeneasr  👀when you get a chance

oauth2/fosite_store_sql.go
```diff
@@ -84,6 +91,10 @@ var Migrations = map[string]*dbal.PackrMigrationSource{
 	}, true),
 }
 
+type transactionKey struct{}
+
+var txKey = transactionKey{}
```
aeneasr: This is fine! However, a more idiomatic would be by using int (optionally with iota)

```
type transactionKey int
const txKey transactionKey = iota
```

oauth2/fosite_store_sql.go
```diff
@@ -213,17 +225,27 @@ func (s *FositeSQLStore) createSession(ctx context.Context, signature string, re
...
+	if _, err := dbAccess.NamedExecContext(ctx, query, data); err != nil {
 		return sqlcon.HandleError(err)
 	}
 	return nil
 }
 
+func (s *FositeSQLStore) getDbAccess(ctx context.Context) dbAccess {
+	if tx, ok := ctx.Value(txKey).(*sqlx.Tx); ok {
+		return tx
+
```
aeneasr: remove empty line

oauth2/fosite_store_sql.go
```diff
@@ -200,6 +211,7 @@ func (s *FositeSQLStore) hashSignature(signature, table string) string {
 }
 
 func (s *FositeSQLStore) createSession(ctx context.Context, signature string, requester fosite.Requester, table string) error {
+	dbAccess := s.getDbAccess(ctx)
```
aeneasr: dbAccess is both a type and a value here. Maybe just rename this to `db := s.db()`?

oauth2/fosite_store_sql.go
```diff
@@ -50,6 +50,13 @@ type FositeSQLStore struct {
 	HashSignature       bool
 }
 
+type dbAccess interface {
```
aeneasr: Rename this to `sqlxDB` which represents the type that sqlx.DB represents?

oauth2/fosite_store_helpers.go
```diff
@@ -385,3 +389,73 @@ func testHelperFlushTokens(x ManagerTestSetup, lifespan time.Duration) func(t *t
...
+			txnStore, ok := m.F.(storage.Transactional)
+			require.True(t, ok)
+
+			ctx := context.Background()
+			ctx, err := txnStore.BeginTX(ctx)
+
+			require.NoError(t, err)
+
+			signature := uuid.New()
+			err = m.F.CreateAccessTokenSession(ctx, signature, createTestRequest(signature))
```
aeneasr: Would it make sense to test this against all methods?

oauth2/fosite_store_test.go
```diff
@@ -134,7 +134,6 @@ func TestManagers(t *testing.T) {
 	for k, store := range fositeStores {
 		TestHelperRunner(t, store, k)
 	}
-
```
aeneasr: Add whitespace :)

oauth2/fosite_store_helpers.go
michaelwagler: @aeneasr sounds good. I just added a whole bunch of tests, let me know if you have suggestions.

There's a lot of repetition so I wrote helpers that take the methods being tested as parameters. I think it mostly works, but there are a couple things that are a bit ugly:

-For the Access token and Refresh token tests, I call `doTestCommit` and do `doTestRollback` twice each, to account for them having both a revoke and delete method

-The `OpenIdConnectSession` tests couldn't use the helper functions, since `GetOpenIDConnectSession` has a different signature than all the other getter methods.

aeneasr: Nice!

aeneasr:
Thank you, Michael!

Issue #1282, opened 2019-02-13
Support multi proxies between TLS termination proxy and hydra


sawadashota:
**Is your feature request related to a problem? Please describe.**

Hydra seems to support only following architecture.

```
Client -> TLS termination proxy -> hydra
```

Then, set TLS termination proxy's CIRDER address as `HTTPS_ALLOW_TERMINATION_FROM`.

But how about following case?

```
Client -> TLS termination proxy -> proxy -> hydra
```
Set TLS termination proxy's CIRDER address then it doesn't work.
Set proxy's CIRDER address then it works but is it best way?

**Describe the solution you'd like**

I think it's the best setting TLS termination proxy's CIDER address as `HTTPS_ALLOW_TERMINATION_FROM`  even if there proxies between TLS termination proxy and hydra.

aeneasr:
Why wouldn't you just set `HTTPS_ALLOW_TERMINATION_FROM` to the IP of `proxy`?

sawadashota:
Yes, it works but I think this is not better in the sense of `HTTPS_ALLOW_TERMINATION_FROM` option.
This option seems to mean "set TLS termination proxy's IP then trust it and check X-Forwarded-Proto header".
Therefore, I think it better to ensure TLS termination proxy's IP.

https://github.com/ory/hydra/blob/aa6ab26908ea5fc856c67c2650c2124d3331e184/cmd/serve.go#L186-L190

In addition my case, using AWS, TLS termination proxy and proxy are different subnet.
When I set `HTTPS_ALLOW_TERMINATION_FROM` to the IP of nginx, it works but I want to change to ALB IP.

```
Client -> ALB (TLS termination) -> nginx on ECS -> hydra on ECS
```

PR #1283, opened 2019-02-13
config: Support multi proxies between TLS termination proxy and hydra

sawadashota:
## Related issue

#1282 

## Proposed changes

Check not only remote IP bu also `X-Forwarded-For` header when enabled TLS termination option.

## Checklist

force push.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -153,22 +153,30 @@ func (c *Config) GetScopeStrategy() fosite.ScopeStrategy {
 }
 
 func matchesRange(r *http.Request, ranges []string) error {
-	ip, _, err := net.SplitHostPort(r.RemoteAddr)
+	forwardedFor := r.Header.Get("X-Forwarded-For")
+	ips := strings.Split(forwardedFor, ", ")
+
+	remoteIp, _, err := net.SplitHostPort(r.RemoteAddr)
 	if err != nil {
 		return errors.WithStack(err)
 	}
+
+	ips = append(ips, remoteIp)
 
 	for _, rn := range ranges {
 		_, cidr, err := net.ParseCIDR(rn)
 		if err != nil {
 			return errors.WithStack(err)
 		}
-		addr := net.ParseIP(ip)
-		if cidr.Contains(addr) {
-			return nil
+
+		for _, ip := range ips {
+			addr := net.ParseIP(ip)
+			if cidr.Contains(addr) {
+				return nil
+			}
 		}
 	}
-	return errors.Errorf("Remote address %s does not match cidr ranges %v", ip, ranges)
+	return errors.Errorf("X-Forwarded-For %s does not match cidr ranges %v", forwardedFor, ranges)
 }
 
 func newLogger(c *Config) *logrus.Logger {
--- a/config/config.go
+++ b/config/config.go
@@ -36,12 +36,13 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
-	"gopkg.in/yaml.v1"
+	yaml "gopkg.in/yaml.v1"
 
 	"github.com/ory/fosite"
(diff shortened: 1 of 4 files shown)


aeneasr:
`X-Forwarded-For` is not being used correctly here. This header is reserved for the IP of the **actual client** (browser, mobile app, ...), not the proxy chain in your network stack.

aeneasr:
Sorry, my bad, I overlooked that you're splitting it. In that case, the header is being used correctly and I think this PR makes sense!

config/config_test.go
```diff
@@ -69,6 +77,16 @@ func TestDoesRequestSatisfyTermination(t *testing.T) {
 		URL:        &url.URL{Path: "/health"},
 	}
 	assert.NoError(t, c.DoesRequestSatisfyTermination(r))
+
+	r = &http.Request{
+		RemoteAddr: "127.0.0.1:123",
+		Header: http.Header{
+			"X-Forwarded-Proto": []string{"https"},
+			"X-Forwarded-For":   []string{"227.0.0.1", "127.0.0.1", "227.0.0.2"},
```
aeneasr: This test doesn't actually do anything, it's just the same as the one above (you're checking for the remote address which is 127.0.0.1 and thus satisfied, `X-Forwarded-For` is not being touched). Please also note that `"X-Forwarded-For":   []string{"227.0.0.1", "127.0.0.1", "227.0.0.2"},` should (most likely) be `"X-Forwarded-For":   []string{"227.0.0.1, 127.0.0.1, 227.0.0.2"},`.

config/config_test.go
```diff
@@ -49,6 +49,14 @@ func TestDoesRequestSatisfyTermination(t *testing.T) {
 	r = &http.Request{Header: http.Header{"X-Forwarded-Proto": []string{"http"}}, URL: new(url.URL)}
 	assert.Error(t, c.DoesRequestSatisfyTermination(r))
 
+	r = &http.Request{
+		Header: http.Header{
```
aeneasr: This test could also fail because no remoteaddress is given (thus SplitHostPort returns an error)

config/config.go
```diff
@@ -153,22 +153,30 @@ func (c *Config) GetScopeStrategy() fosite.ScopeStrategy {
 }
 
 func matchesRange(r *http.Request, ranges []string) error {
-	ip, _, err := net.SplitHostPort(r.RemoteAddr)
+	forwardedFor := r.Header.Get("X-Forwarded-For")
```
aeneasr: This has some issues, one of them is that `strings.Split` always returns at least one element (even if the string is empty) and that we shouldn't expect a whitespace after `,`. Let's rewrite this a bit:

```
remoteIP, _, err := net.SplitHostPort(r.RemoteAddr)
if err != nil {
	return errors.WithStack(err)
}

check := []string{remoteIP}
for _, fwd := stringsx.Split(r.Header.Get("X-Forwarded-For"), ",") {
    check = append(check, strings.Trim(fwd))
}
```

config/config.go
```diff
@@ -153,22 +153,30 @@ func (c *Config) GetScopeStrategy() fosite.ScopeStrategy {
...
+
+		for _, ip := range ips {
+			addr := net.ParseIP(ip)
+			if cidr.Contains(addr) {
+				return nil
+			}
 		}
 	}
-	return errors.Errorf("Remote address %s does not match cidr ranges %v", ip, ranges)
+	return errors.Errorf("X-Forwarded-For %s does not match cidr ranges %v", forwardedFor, ranges)
```
aeneasr: This should probably be `neither remote address nor any x-forwarded-for values match CIDR ranges %v: %v, ranges, check)`

config/config_test.go
sawadashota: Sorry, this was meaningless test...

aeneasr:
Awesome! Thank you :)

Issue #1285, opened 2019-02-14
Consent /reject without error data will always return an invalid_request error


LaylConway:
**Describe the bug**
When rejecting a consent request, but not including "error", "error_description" or "error_hint", the values that aren't given will default to a generic invalid_request error. This is completely inaccurate to what happened and very confusing while developing.

**To Reproduce**
1. Go through the OAuth2 authorization code flow until consent
2. Reject consent, but do not set any value for "error", "error_description" or "error_hint"
3. Read the resulting error in the callback

**Expected behavior**
The right behavior should probably be either passing an error that indicates the consent request was rejected, or a generic error. The invalid_request error sends the developer to the wrong place for debugging.

**Version:**
 - Environment: Docker
 - Version oryd/hydra:v1.0.0-rc.6_oryOS.10

aeneasr:
Thank you! Feel free to PR :)

PR #1297, opened 2019-03-05
Fix 1285

kminehart:
## Related issue

#1285 

## Proposed changes

Summary: Changed the defaults for `Name`, `Description`, and `Hint` from `consent/types.go`.

Changes:

1. Changed the `ToRFCError()` function to only provide defaults if all 3 (Name, Description, and Hint) are left blank.
    * I think it's safe to assume that if a developer provides any combination of "Name", "Description", or "Hint", injecting any defaults will cause confusion.
2. Changed the default values of `Name`, `Description`, and `Hint` to be more generic, and clear that they were left blank.
    * Name: "Consent request denied"
    * Description: "The request was denied and no further description was provided"
    * Given no context at the time of generation, no valuable hint could be provided, so it will remain empty.

## Checklist

force push.


aeneasr:
I don't get why crypto is failing all the time. You can fix this with `go mod tidy` though!

consent/types.go
```diff
@@ -27,6 +27,11 @@ import (
 	"github.com/ory/hydra/client"
 )
 
+const (
+	requestDeniedErrorName        = "Consent request denied"
```
aeneasr: Errors names [should not start with a capital letter](https://github.com/golang/go/wiki/Errors) :)

consent/types.go
```diff
@@ -53,18 +58,14 @@ type RequestDeniedError struct {
...
 		e.Code = fosite.ErrInvalidRequest.Code
 	}
-	if e.Description == "" {
-		e.Description = fosite.ErrInvalidRequest.Description
-	}
-	if e.Hint == "" {
-		e.Hint = fosite.ErrInvalidRequest.Hint
-	}
 	if e.Debug == "" {
 		e.Debug = fosite.ErrInvalidRequest.Debug
```
aeneasr: Probably makes sense to remove this too?

consent/types.go
kminehart: I like capitalization too much.

Fixed.

consent/types.go
kminehart: Yeah I think so.

Should the default error code stay as `fosite.ErrInvalidRequest` (400)?

aeneasr:
Alternatively `go get -u  golang.org/x/crypto` with go modules enabled.

kminehart:
My bad, go modules are a new thing for me. I ran `go mod tidy` and committed the `go.mod` and `go.sum` files. That should suffice, right?

aeneasr:
Perfect, last thing would be to add a test so this doesn't break in a future patch by accident :)

kminehart:
Cool. I wasn't sure what the right place was to put the unit tests for this change, or if they were necessary given the nature of the tests in `hydra/consent`.

I also am not sure if there needs to be any documentation changes. I couldn't find anything relevant but I could have missed something.

consent/types.go
```diff
@@ -27,6 +27,11 @@ import (
 	"github.com/ory/hydra/client"
 )
 
+const (
+	requestDeniedErrorName        = "consent request denied"
+	requestDeniedErrorDescription = "the request was denied and no further description was provided"
```
aeneasr: The description can be upper case :)

consent/types_test.go
```diff
@@ -0,0 +1,79 @@
...
+	for k, tc := range []struct {
+		input  *RequestDeniedError
+		expect *fosite.RFC6749Error
+	}{
+		{
+			input: &RequestDeniedError{
+				Description: "not empty",
+			},
+			expect: &fosite.RFC6749Error{
+				Name:        "",
```
aeneasr: Wouldn't it make sense for the name and description to always be set to the default values? Then you can override e.g. just the description or only the name without having to set both?

consent/types.go
kminehart: ahhhh!

kminehart:
@aeneasr I think this one is ready.

Pretty basic, if `Name` is not provided, it'll use the default. `Description` and `Hint` no longer have defaults.  All of the commits were me just getting used to the way everything is structured. 😅

aeneasr:
Thank you!

Issue #1296, opened 2019-03-04
Invalid expiration time during introspection the token refresh


pr0head:
**Describe the bug**
Introspection of refresh token returns incorrect expiration time. The expiration time is the same as the access token expiration time.

**To Reproduce**
1. Download latest version from GitHub
2. Add to the docker-compose.yml environment variable: `- REFRESH_TOKEN_LIFESPAN=720h`
3. Run `docker-compose -p hydra -f docker-compose.yml up --build`
4. Log in and get the code to exchange tokens
5. Exchange code for tokens
6. Send a refresh token to the endpoint of introspection
7. Received token has an expiration time of 1 hour (default value for access tokens)

**Expected behavior**
The expiration time should be 720 hours, as specified by the configuration parameter.

**Version:**
 - Environment: Docker Desktop 2.0.0.3 (31259), Kubernetes 1.10.11
 - Version: master

aeneasr:
What does `git rev-parse HEAD` say?

pr0head:
1295663ada908dd431d7ffc9927feb6e2606b724

aeneasr:
Your master is way behind, please try with a recent version...

pr0head:
I updated to the latest version of the wizard - c174f96e6e8ab31aa362c7a5d32e5637984aab5b, but the problem has not been fixed

aeneasr:
Could you provide appropriate CURL commands for easier debugging? Also, does this affect only the response of token introspection?

pr0head:
> Also, does this affect only the response of token introspection?
Yes, the expiration time for the access token is returned correctly. When the ACCESS_TOKEN_LIFESPAN parameter is changed, the expiration time also changes correctly.

```
curl -X POST \
  http://192.168.5.12:4445/oauth2/introspect \
  -H 'Postman-Token: 50e3be92-4917-4fc7-ae68-d789e82f17fc' \
  -H 'cache-control: no-cache' \
  -d 'token=0KuBIL84xrL0cy72OcuWP_sVu8e3NzJ4447-ziwEadk.eUGZskTbVlWl_U_NQjjP00ScOz_K_ehgrn3PTvOHQTs&undefined='
{"active":true,"scope":"openid offline","client_id":"5c7d930a02429c10c0864511","sub":"5c7d934002429c10c0864513","exp":1551744293,"iat":1551733492,"iss":"http://192.168.5.12:4444/","token_type":"refresh_token"}
```

aaslamin:
[...]

PR #1300, opened 2019-03-05
Return the expiration time of the token, depending on its type, on the endpoint of introspection.

pr0head:
## Related issue
https://github.com/ory/hydra/issues/1296

## Proposed changes
The change allows to get the correct expiration time of the token, depending on its type.

## Checklist
- [x] I confirm that this pull request does not address a security vulnerability. If this pull request addresses a security vulnerability, I confirm that I got green light (please contact [hi@ory.sh](mailto:hi@ory.sh)) from the maintainers to push the changes.
by signing my commit(s). You can amend your signature to the most recent commit by using `git commit --amend -s`. If you amend the commit, you might need to force push using `git push --force HEAD:<branch>`. Please be very careful when using force push.


pr0head:
Sorry, new PR created

pr0head:
Couldn't do the tests, I'll try later.

oauth2/handler.go
```diff
@@ -430,9 +430,13 @@ func (h *Handler) IntrospectHandler(w http.ResponseWriter, r *http.Request, _ ht
...
-	exp := resp.GetAccessRequester().GetSession().GetExpiresAt(fosite.AccessToken)
+	exp := resp.GetAccessRequester().GetSession().GetExpiresAt(tt)
 	if exp.IsZero() {
-		exp = resp.GetAccessRequester().GetRequestedAt().Add(h.AccessTokenLifespan)
+		if tt == fosite.RefreshToken {
+			exp = resp.GetAccessRequester().GetRequestedAt().Add(h.RefreshTokenLifespan)
+		} else {
+			exp = resp.GetAccessRequester().GetRequestedAt().Add(h.AccessTokenLifespan)
+		}
 	}
```
aaslamin: What about the lifetime of authorization codes and id_tokens? Perhaps fetching the expiry can be extracted to a non-exported helper somewhere within this package or file.

aaslamin: Also, it is possible for refresh tokens to not expire if the value of the lifespan has been set to -1. In that case, perhaps it is best to omit the expiry field in the response payload.

@aeneasr thoughts?

aaslamin:
Thanks for taking this on @pr0head! 🏆

oauth2/handler.go
pr0head: > What about the lifetime of authorization codes and id_tokens? Perhaps fetching the expiry can be extracted to a non-exported helper somewhere within this package or file.

The introspection point serves access or update tokens. Other tokens are outside the specification. 
https://tools.ietf.org/html/rfc7662#section-2.1

aaslamin: Ah right! 💯

aeneasr:
Why is `golang.org/x/...` so f* annoying. Use `go mod tidy` to update... :)

pr0head:
> Why is `golang.org/x/...` so f* annoying. Use `go mod tidy` to update... :)

But I did not make any new dependencies, did not change the go.mod and go.sum. Maybe I do not understand something, could you explain to me?

oauth2/handler.go
aeneasr: > Also, it is possible for refresh tokens to not expire if the value of the lifespan has been set to -1. In that case, perhaps it is best to omit the expiry field in the response payload.

Damn, you know this system better by me. I'm almost 💯 that's true. In that case it should be omitted.

> The introspection point serves access or update tokens. Other tokens are outside the specification.
https://tools.ietf.org/html/rfc7662#section-2.1

Probably just a typo, introspection serves **access** and **refresh** tokens. So @aaslamin point is right!

aeneasr:
> For this behavioural change we will want tests to avoid future regressions.
> 
> Some test scenarios:
> 
> * Introspecting a refresh token that have been configured to never expire
> * Introspecting a refresh token that has been configured to expire (_use a different expiry than that of the access token_)
> * Introspecting an access token
> * Introspecting an authorization code?
> * Introspecting an id_token?

Only access and refresh tokens are relevant :) All the other cases are super valid though!

aeneasr:
> > Why is `golang.org/x/...` so f* annoying. Use `go mod tidy` to update... :)
> 
> But I did not make any new dependencies, did not change the go.mod and go.sum. Maybe I do not understand something, could you explain to me?

I would love to but I can't. For some reason `golang.org/x`is special and breaks constantly. If you find out why I'd invite you for a night out or something because it's so f* annoying.

aeneasr:
Hi there! Sorry for the silence here - can you rebase on master? The broken test will go away by magic then. :)

aeneasr:
Nice, thanks! :)

Issue #1256, opened 2019-01-14
Call to consent accept/reject for a second time gives error


condemil:
**Describe the bug**
When I call hydra API endpoint `PUT /oauth2/auth/requests/consent/{id]/reject` or `/accept` for a second time with the same id it gives the error back:

```
An error occurred while handling a request" code=409 details="map[]" error="Error 1062: Duplicate entry '{id}' for key 'PRIMARY': Unable to insert or update resource because a resource with that value exists already
```

**To Reproduce**
1. Call `PUT /oauth2/auth/requests/consent/{id]/reject` twice
2. See error in Hydra logs

**Expected behavior**
I expect that this behavior is handled correctly, I think the best approach is to validate the existing primary key in db and return same response as for the first call.

**Screenshots**
N/A

**Version:**
 - Environment: Hydra Alpine docker image
 - Version: v1.0.0-rc.6_oryOS.1

**Additional context**
N/A

aeneasr:
The error is handled correctly though with a 409. What problem are you facing when the error is returned? Is it that you need to check for the 409 error state explicitly?

condemil:
I see error in Hydra logs, I expect errors to be something that should be fixed in code or do Hydra app use some different approach?

```
level=error msg="An error occurred while handling a request" code=409
```

condemil:
Another problem I am facing is that when I receive the 409 error the `redirect_to` is not available in the response and I cannot redirect user back properly. Imagine the situation when the first request was handled and connection broke before consent provider received the response and consent provider is re-sending it once again.

condemil:
Actually I just checked and see the same error for `PUT /oauth2/auth/requests/consent/{id}/accept` call too.

kewde:
I ran into the same/similar issue but with *DIFFERENT CHALLENGE IDs*.

This occurs when consent `remember` is set to true and consent has been granted.
A second new request will have the consent information with `skip`true.
I was unable to accept  the consent and redirect the user.

kewde:
Thanks, that solved it indeed!

xiven:
I'm running into an issue where when I make the login request I notice that the login_challenge is not getting set on the response.  I'm thinking this may be why I always have skip set to false on the accept consent response when I log in again even though I set the remember flag to true from the consent form.  I'm not sure if that is related to #1165.
Why would the login_challenge not be getting set on login?

PR #1301, opened 2019-03-06
Make HandleConsentRequest idempotent

kminehart:
## Related issue

#1256 

## Proposed changes

* Added an error handler `sqlcon.ErrUniqueValidation` on `SQLManager.HandleConsentRequest` in `hydra/consent`, making `HandleConsentRequest` idempotent.

## Checklist

force push.

## Further comments

The original issue was about misleading error logs and ending up in a difficult place after receiving an unexpected 409.

The documentation also does not display a 409 as a possibility for most of these endpoints.
* https://github.com/ory/hydra/blob/master/consent/handler.go#L529
* https://github.com/ory/hydra/blob/master/consent/handler.go#L458

It took me a while to wrap my head around a proper solution for this for some reason; mostly with the layout of the various tests. I had a hard time finding a good place for the repeated `HandleConsentRequest` call 😅


consent/manager_sql.go
```diff
@@ -287,7 +287,11 @@ func (m *SQLManager) HandleConsentRequest(ctx context.Context, challenge string,
 		strings.Join(sqlParamsConsentRequestHandled, ", "),
 		":"+strings.Join(sqlParamsConsentRequestHandled, ", :"),
 	), d); err != nil {
-		return nil, sqlcon.HandleError(err)
+		err = sqlcon.HandleError(err)
+		if errors.Cause(err) == sqlcon.ErrUniqueViolation {
```
aeneasr: In this case we wouldn't override the existing value. However, I think that's what we should do here as long as `was_used` is false.

consent/manager_sql.go
```diff
@@ -287,7 +287,38 @@ func (m *SQLManager) HandleConsentRequest(ctx context.Context, challenge string,
...
+			return m.ReplaceUnusedConsentRequest(ctx, challenge, r)
+		}
+		return nil, err
+	}
+
+	return m.GetConsentRequest(ctx, challenge)
+}
+
+func (m *SQLManager) ReplaceUnusedConsentRequest(ctx context.Context, challenge string, r *HandledConsentRequest) (*ConsentRequest, error) {
+	d, err := newSQLHandledConsentRequest(r)
```
aeneasr: This method can be shortened, `d` is already defined by it's outer method `HandleConsentRequest` and processing it another time is thus not necessary. `ReplaceUnusedConsentRequest` should also not be exported since it's only being used internally.

consent/manager_sql.go
```diff
@@ -287,7 +287,38 @@ func (m *SQLManager) HandleConsentRequest(ctx context.Context, challenge string,
...
+	if err != nil {
+		return nil, err
+	}
+
+	if _, err := m.DB.NamedQueryContext(ctx, fmt.Sprintf(`
```
aeneasr: Query implements a scanner for things like `SELECT`, not an executor, see [query](https://golang.org/pkg/database/sql/#DB.Query) versus [exec](https://golang.org/pkg/database/sql/#DB.Exec) in godoc. If you update this to `NamedExecContext` be aware that it won't return `sqlcon.ErrNoRows` if no rows are affected. Instead, you need to check the first argument, probably something like `results.ReowsAffected()`. If it's 0, the update failed.

consent/sql_helper.go
```diff
@@ -90,6 +90,20 @@ var sqlParamsConsentRequestHandled = []string{
 	"was_used",
 }
 
+var sqlParamsConsentRequestHandledUpdate = []string{
+	"challenge=:challenge",
```
aeneasr: Let's avoid duplication and construct this from `sqlParamsConsentRequestHandled` instead.

consent/manager_sql.go
kminehart: That makes the logic a lot cleaner, actually. I return `m.GetConsentRequest` regardless of how many rows were affected. I just needed to make sure that if an error was returned, then take the appropriate action. (I thought 0 rows affected would apply here)

aeneasr:
Nice! This LGTM - run `make format` and the format error will go away too :)

kminehart:
hmm. I ran `make format` but that didn't seem to resolve the missing dependencies. I see that it's listed in `go.mod`; `go mod tidy` didn't seem to help.

Any thoughts? 🤔

aeneasr:
That's working fine, there's a merge conflict with go.sum which can be resolved I think easily. The tests fail due to flakiness and also some code errors (see the format task).

test/mock-client/main.go
```diff
@@ -34,7 +34,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/dgrijalva/jwt-go"
```
aeneasr: that's probably not intentional?

kminehart: Not sure what changed that, or why I `git add`ed it. Definitely not intentional.

kminehart:
I completely borked the history on this one. I forgot to sign a commit somewhere and I'm still not sure how to correct that if that happens.

if needed I can recreate the changes and open a separate PR that doesn't screw around with the history.

aeneasr:
Yeah this has only one change now?

aeneasr:
@kminehart will you be opening a new PR or trying to fix this one? If you need help with git let me know.

kminehart:
Yeah I'll open a new PR for this one.

I ran the `rebase` commit that the DCO bot suggested; that on top of the merge I did from upstream/master tricked Github. 😅  I'll do that now.

PR #1302, opened 2019-03-06
circleci: disable modules support temporarily when fetching a tool

aaslamin:
@aeneasr 🙏


aeneasr:
I love you.

aaslamin:
FYI: https://github.com/golang/go/wiki/Modules#why-does-installing-a-tool-via-go-get-fail-with-error-cannot-find-main-module

Issue #1321, opened 2019-03-20
Update docker-compose to v3


lopezator:
Update to docker-compose version 3, as it is the standard and recommended right now:

See:

https://docs.docker.com/compose/compose-file/compose-versioning/

Issue #1196, opened 2018-12-01
DX: Easily support different workflows by sharing compose configurations


aaslamin:
**Describe the solution you'd like**

The current docker compose setup in Hydra makes it cumbersome to switch between various workflows (e.g. switching between mysql/postgres, setting up tracing, etc). Currently, one has to [uncomment the services/configurations they are after](https://github.com/ory/hydra/blob/master/docker-compose.yml#L40).

Instead, we can use multiple compose files to customize the application to support a completely different workflow. This is supported by compose - https://docs.docker.com/compose/extends/#multiple-compose-files

I propose we have a default configuration with postgres, with extensions/overrides for mysql and setting up tracing via jaeger:

• docker-compose.with-mysql.yml
• docker-compose.with-jaeger.yml

For example, if I want to setup Hydra to use mysql instead AND configure tracing via Jaeger:

`docker-compose -f docker-compose.yml -f docker-compose.with-mysql.yml -f docker-compose.with-jaeger.yml up`

This makes it easy for people to add new workflows in the future (e.g. a setup for Prometheus).

aeneasr:
Oh yeah, super nice!

PR #1323, opened 2019-03-20
docker: update docker-compose to v3 + docker-compose files refactor

lopezator:
## Related issues

#1321 #1196

## Proposed changes

* Update docker-compose to v3.
* Reorganize/refactor docker-compose config into multiple files for cleanness and readability.

## About the second commit, reorganize/refactor...

I've detected six usecases of the docker-compose files.

1. postgres 
2. mysql
3. postgres + tracing
4. mysql + tracing
5. twoc-postgres
6. twoc-mysql

Instead of forcing the user to edit the `docker-compose.yml` and `docker-compose-twoc.yml` files by hand, the usage now would be:

1. docker-compose up --build
2. docker-compose -f docker-compose.yml -f docker-compose-mysql.yml up --build
3. docker-compose -f docker-compose.yml -f docker-compose-tracing.yml up --build
4. docker-compose -f docker-compose.yml -f docker-compose-mysql.yml -f docker-compose-tracing.yml up --build
5. docker-compose -f docker-compose-twoc.yml up --build
6. docker-compose -f docker-compose-twoc.yml -f docker-compose-twoc-mysql.yml up --build

## How does it work:

docker-compose does a merge between the files you pass  in via `-f`. So idea is to just override/modify the config as needed from base `docker-compose.yml`.

#### Advantages

* Cleaner to read.
* User doesn't have to touch files by hand to change the database or add tracing.

#### Disadvantages

* Commands are longer to type if you are not on standard configuration.
* Requires better docker-compose understanding if you want to tweak the files.

## Notes

* An edit of the docs would be required if this gets approved.
* I've tested all the described scenarios using `docker-compose config` and diff-ing the files with the produced output by the original config to ensure the behavior is the same on both scenarios.

CC \ @aeneasr

### DIFF
--- a/docker-compose-mysql.yml
+++ b/docker-compose-mysql.yml
@@ -1,0 +1,29 @@
+###########################################################################
+#######             FOR DEMONSTRATION PURPOSES ONLY                 #######
+###########################################################################
+#                                                                         #
+# If you have not yet read the tutorial, do so now:                       #
+#  https://www.ory.sh/docs/hydra/5min-tutorial                            #
+#                                                                         #
+# This set up is only for demonstration purposes. The login               #
+# endpoint can only be used if you follow the steps in the tutorial.      #
+#                                                                         #
+###########################################################################
+
+version: '3'
+
+services:
+
+  hydra-migrate:
+    environment:
+      - DATABASE_URL=mysql://root:secret@tcp(mysqld:3306)/mysql?parseTime=true
+
+  hydra:
+    environment:
+      - DATABASE_URL=mysql://root:secret@tcp(mysqld:3306)/mysql?parseTime=true
+    restart: unless-stopped
+
+  mysqld:
+    image: mysql:5.7
+    environment:
+      - MYSQL_ROOT_PASSWORD=secret
--- a/docker-compose-tracing.yml
+++ b/docker-compose-tracing.yml
@@ -1,0 +1,36 @@
+###########################################################################
+#######             FOR DEMONSTRATION PURPOSES ONLY                 #######
+###########################################################################
+#                                                                         #
+# If you have not yet read the tutorial, do so now:                       #
+#  https://www.ory.sh/docs/hydra/5min-tutorial                            #
+#                                                                         #
+# This set up is only for demonstration purposes. The login               #
+# endpoint can only be used if you follow the steps in the tutorial.      #
+#                                                                         #
+###########################################################################
+
+version: '3'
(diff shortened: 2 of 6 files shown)


aeneasr:
That's dope! We will have to [update the docs](https://github.com/ory/docs/blob/master/docs/hydra/5min-tutorial.md) to reflect those changes though! Would you be up to do that?

It might also be a good idea to include this

```
docker-compose up --build
docker-compose up -f docker-compose.yml -f docker-compose-mysql.yml --build
docker-compose up -f docker-compose.yml -f docker-compose-tracing.yml --build
docker-compose up -f docker-compose.yml -f docker-compose-mysql.yml -f docker-compose-tracing.yml --build
docker-compose up -f docker-compose-twoc.yml --build
docker-compose up -f docker-compose-twoc.yml -f docker-compose-twoc-mysql.yml --build
```

somewhere in the docs. Any suggestions?

milesrichardson:
This is how I use docker-compose with my projects too. The commands can be pretty tedious to write (or realistically, copy/paste), though. Maybe worth wrapping them in a script?

lopezator:
That's a great idea @milesrichardson, what do you think @aeneasr? 

And since english is not my native language, I also would appreciate help on where and how reflect in the docs the changes made here.

Maybe you could help in this matter @milesrichardson ?

aaslamin:
Also related: https://github.com/ory/hydra/issues/1196

Nice work! As far as a script, perhaps a `Makefile` would make sense?

Something like `make run-with-mysql`, etc

aeneasr:
Yeah, having that in a `Makefile` would be great!

lopezator:
> Also related: #1196

Nice work! As far as a script, perhaps a Makefile would make sense?

Something like make run-with-mysql, etc

Cool. I think that closing this would close the linked issue also. This approach is slightly different, because extends is not supported on docker-compose v3, but equivalent.

lopezator:
I ended up doing a bash script as it was cleaner IMHO. PTAL @aeneasr

aeneasr:
The problem is that that won't work on windows (yeah those people exist...) :/ Do you have an idea how we could make it work for them too? I think especially for the 5 minutes guide that would be important. Alternatively, we'll do the 5 minutes tutorial with those docker commands and have the other stuff as a script or in the Makefile.

aeneasr:
I'll come up with a solution - thank you for your hard work so far!

PR #1324, opened 2019-03-25
Remove opencollective from package.json

DASPRiD:
This PR removes opencollective from the package.json, as it causes problems when npm installing this package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,21 +4,10 @@
   "description": "The official JavaScript / NodeJS SDK for ORY Hydra.",
   "license": "Apache-2.0",
   "main": "sdk/js/swagger/src/index.js",
-  "scripts": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -4,21 +4,10 @@
...
   "main": "sdk/js/swagger/src/index.js",
-  "scripts": {
-    "postinstall": "opencollective postinstall"
-  },
   "browser": {
     "fs": false
   },
   "dependencies": {
     "superagent": "3.7.0"
-  },
```
aeneasr: Could you move the open collective stuff into a different package.json, e.g. `package.oc.json`? It's relevant for generating the badges in the readme. Thank you! :)

DASPRiD: Sure, updated :)

Issue #1165, opened 2018-11-08
consent: Remember logic confuses developers


aeneasr:
Right now, the `remember` payload has to be `false` or `false || !request.skip` for both login and consent requests. This is confusing and not very clear, and does not serve a real purpose. Let's think about why it is there and if it's non-sensical to keep this check, remove it.

aeneasr:
I think the idea here was to educate people about the consent flow and user interaction model, but that doesn't really happen - all it does is make this flow more confusing.

kminehart:
I think reading https://www.ory.sh/docs/hydra/oauth2 makes the `remember` process makes sense.

What alternative would you suggest to imply that a login UI should not be presented?

Maybe to resolve confusion it would worth renaming "skip" to "remembered". Just a guess, but most people might expect the data in a response to be descriptive of the data rather than the action the client should take.

aeneasr:
So `skip` would still indicate that the UI should not be presented. I think having `remember` forcibly true/false is just a bit too much? Since `skip: true` indicates that the process has been remembered already, we can probably just discard `remember`? Does that make sense?

aeneasr:
So what I want to say is, `skip` will only be `true`, if that specific session has had `remember` set to `true`. Since `true` > `false` in terms of authorization (remembering consent/login usually requires extra consent like clicking a checkbox), there will be no negative implication (read privilege escalation) if `remember` is set to `true` if `skip` is `true` because `remember` had to be `true` before anyways. If `remember` is set to `false` it won't have an impact (imo) because, well, the session was remembered before and the user is actually not presented the user interface so the value `remember` can be ignored. Not sure that makes sense how I wrote it, but it does make sense in my head :D

kminehart:
Would removing `remember` from responses cause problems with existing clients? Or would it be easier to just set it to `false` everywhere?

aeneasr:
Setting it to false / ignoring it will be the safest route IMO. Since it didn't have an effect before, it won't change that now either.

kminehart:
A PR for this will hopefully be ready tonight. :)

PR #1325, opened 2019-03-26
Default `Remember` to false in payloads with `skip`

kminehart:
Signed-off-by: Kevin Minehart <kmineh0151@gmail.com>

## Related issue

#1165 

## Proposed changes

~Basically, `toHandledConsentRequest` and `toHandledAuthenticationRequest`, being only used in a 2 places (and ultimately as a response), now return `False` for `Remember`~

In the handlers, instead of checking for `Remember` and `Skip`, and returning an error if both are true, just set `Remember` to false (as a skipped request can not be remembered) when `Skip` is true.

## Checklist

force push.
- [-] I have added tests that prove my fix is effective or that my feature works

## Further comments


consent/sql_helper.go
```diff
@@ -408,7 +408,7 @@ func (s *sqlHandledAuthenticationRequest) toHandledAuthenticationRequest(a *Auth
 	return &HandledAuthenticationRequest{
 		ForceSubjectIdentifier: s.ForceSubjectIdentifier,
 		RememberFor:            s.RememberFor,
-		Remember:               s.Remember,
+		Remember:               false,
```
aeneasr: That does not seem right. Why would this always be false? And why only for SQL? This should be done in the handler instead for requests for `Skip` is false. This change needs to be reverted.

consent/sql_helper_test.go
```diff
@@ -62,7 +62,7 @@ func TestSQLAuthenticationConverter(t *testing.T) {
 	b := &HandledAuthenticationRequest{
 		AuthenticationRequest: a,
 		RememberFor:           120,
-		Remember:              true,
+		Remember:              false,
```
aeneasr: I think this needs to be reverted.

Issue #990, opened 2018-08-17
Feature request: CockroachDB support


glerchundi:
I'm evaluating Hydra v1.0.0-beta.8 and its compatibility with CockroachDB and I can confirm that except for two issues, things are working properly.

These two issues are:

- [ ] `ALTER COLUMN ... SET NOT NULL` is not supported by CockroachDB which is required by [this migration](https://github.com/ory/hydra/blob/v1.0.0-beta.8/client/manager_sql.go#L144-L148). They've opened an issue for tracking this feature request (cockroachdb/cockroach#28751) and for your interest this probably was triggered by the issue I raised some days ago cockroachdb/cockroach#28415.
- [x] Updates without a `WHERE` statement are not supported although there is a work-around by calling `SET sql_safe_updates = false` explicitly in the current connection or transaction. Another one could be to modify current migration to do an `UPDATE .. WHERE (SELECT *)` or something like that. *Nothing to do here, for details take a look here*: https://github.com/cockroachdb/cockroach/issues/28826#issuecomment-414236373.

WDYT, are you open to include CockroachDB support?

arekkas:
I really want to avoid having another migration table like [here](https://github.com/ory/ladon/blob/master/manager/sql/databases.go). I also suspect that the two issues you mentioned wont be the only ones. I'm not sure how well CockroachDB works with SQL and specifically more complex join and lock statements. As most cloud providers support manages postgres and mysql at very low cost (but not cockroach), adding cockroach is not a priority for this project.

glerchundi:
> I really want to avoid having another migration table like here.

The idea is to have the same migration table as the postgres one, I'm aligned with you here.

> I also suspect that the two issues you mentioned wont be the only ones.

CockroachDB is working hard making everything pg-compat so if something doesn't work it should be fixed in the upstream and not here.

> I'm not sure how well CockroachDB works with SQL and specifically more complex join and lock statements.

Can you point me to concrete SQL sentences, the more complex one you can think of?

> As most cloud providers support manages postgres and mysql at very low cost (but not cockroach)

Yeah, but CockroachDB is open-source (at least most of its features) and with a growing user base. Great PostgreSQL compatibility, automatic sharding (world-wide capable), NoSQL capabilities, Geo-Partitioning and cloud-native philosophy (no need for an operator to run inside a k8s cluster, for example) makes this a great database to start with.

> adding cockroach is not a priority for this project.

Would you be open to do some reviews of further PRs at least?

Thanks for your time.

[...]

PR #1326, opened 2019-03-26
all: add support for cockroachdb

lopezator:
add cockroachdb support

## Related issue

https://github.com/ory/hydra/issues/990


lopezator:
PTAL just to start discussing about this @aeneasr

aeneasr:
Thank you for this PR! There's currently a major effort going on which will restructure how internals are instantiated: https://github.com/ory/hydra/pull/1314

Basically, there is a central "registry" (sort of like a ServiceLocator) which is responsible for setting up everything properly. This refactoring will also touch cockroachdb and might even make things a bit easier. I don't think there's a ton that needs to change in this PR for that but it would make sense to rebase on #1314 before merging this. Plan is to have the PR done either by today or by tomorrow, it's fairly far in terms of progress.

lopezator:
Thanks @aeneasr! If that is so advanced I think I can put this on hold for a couple of days and rebase from master once merged.

aeneasr:
I think that would be the best course of action!

PR #1366, opened 2019-04-18
Allow localhost to skip consent for public clients

### LINKED ISSUE #1364: Unable to test silent refresh in local development
This might be related to #1068, getting the same error as those users. I did some digging around, and if I understand @aeneasr post on [this issue here](https://github.com/openid-certification/oidctest/issues/97) correctly, this is basically caused by the silent renew requiring additional user consent, unless the callback URL has an HTTPS scheme.

This is clearly fine for normal deployment, but it breaks in local development. Would it be possible to whitelist "localhost" return URLs for that case, possibly as a server config variable? Browsers actually follow the same pattern, e.g. with service workers, which normally require https, except when the host is "localhost".

DASPRiD:
This PR **should** fix #1364.

I'm not a go developer, and I didn't manage to get the tests working (unrelated to this code change). Basically I was getting the following error during testing, which did not result from my code changes:

```
client/handler.go:209: not enough arguments in call to pagination.Header
	have (*url.URL, int, int, int)
	want (http.ResponseWriter, *url.URL, int, int, int)
```

### DIFF
--- a/consent/strategy_default.go
+++ b/consent/strategy_default.go
@@ -416,11 +416,12 @@ func (s *DefaultStrategy) requestConsent(w http.ResponseWriter, r *http.Request,
 	// accepted, as appropriate.
 	if ar.GetClient().IsPublic() {
 		// The OpenID Connect Test Tool fails if this returns `consent_required` when `prompt=none` is used.
-		// According to the quote above, it should be ok to allow https to skip consent.
+		// According to the quote above, it should be ok to allow https to skip consent. Additionally any localhost
+		// hostname is allowed to skip consent.
 		//
 		// This is tracked as issue: https://github.com/ory/hydra/issues/866
 		// This is also tracked as upstream issue: https://github.com/openid-certification/oidctest/issues/97
-		if ar.GetRedirectURI().Scheme != "https" {
+		if ar.GetRedirectURI().Scheme != "https" && !(fosite.IsLocalhost(ar.GetRedirectURI() && ar.GetRedirectURI().Scheme == "http")) {
 			return s.forwardConsentRequest(w, r, ar, authenticationSession, nil)
 		}
 	}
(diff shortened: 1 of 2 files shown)


consent/strategy_default.go
```diff
@@ -416,11 +416,12 @@ func (s *DefaultStrategy) requestConsent(w http.ResponseWriter, r *http.Request,
...
 	if ar.GetClient().IsPublic() {
 		// The OpenID Connect Test Tool fails if this returns `consent_required` when `prompt=none` is used.
-		// According to the quote above, it should be ok to allow https to skip consent.
+		// According to the quote above, it should be ok to allow https to skip consent. Additionally any localhost
+		// hostname is allowed to skip consent.
 		//
 		// This is tracked as issue: https://github.com/ory/hydra/issues/866
 		// This is also tracked as upstream issue: https://github.com/openid-certification/oidctest/issues/97
-		if ar.GetRedirectURI().Scheme != "https" {
+		if ar.GetRedirectURI().Scheme != "https" && !(fosite.IsLocalhost(ar.GetRedirectURI() && ar.GetRedirectURI().Scheme == "http")) {
```
aeneasr: Seems like there's an issue with the parenthesis here:

```
!(fosite.IsLocalhost(ar.GetRedirectURI() && ar.GetRedirectURI().Scheme == "http"))
```

should probably be:

```
!(fosite.IsLocalhost(ar.GetRedirectURI()) && ar.GetRedirectURI().Scheme == "http"))
```

DASPRiD:
> Thank you :) So that error probably comes from you not using go modules. The best way to do that is (after git clone, you need Go 1.11+):

Oh, good point, Ubuntu 18.04 still comes with Go 1.10, I gonna update and try again over the weekend.

PR #1374, opened 2019-04-19
oauth2: Resolve memory leak in gorilla/sessions

### LINKED ISSUE #1363: Memory leak with jaeger tracing enabled
**Describe the bug**
Memory leak with Jaeger tracing enabled.

**To Reproduce**
1. Run serve all with TRACING_PROVIDER="jaeger", TRACING_PROVIDERS_JAEGER_SAMPLING_TYPE="const",  TRACING_PROVIDERS_JAEGER_SAMPLING_VALUE="1"
2. run test plan with jmetter
3. hydra's memory usage grows

**Screenshots**
memory

**Version:**
 - Environment: CentOS Linux release 7.5.1804 (Core), Docker version 18.09.2, build 6247962, image oryd/hydra:v1.0.0-rc.8_oryOS.10, Postgres 10.7
 - Version v1.0.0-rc8

Mem profiles with and without tracing:
[mem-trc-on.zip](https://github.com/ory/hydra/files/3085142/mem-trc-on.zip)
[mem-trc-off.zip](https://github.com/ory/hydra/files/3085143/mem-trc-off.zip)

aeneasr:
Closes #1363

### DIFF
--- a/cmd/server/handler.go
+++ b/cmd/server/handler.go
@@ -35,9 +35,8 @@ import (
 	"github.com/ory/x/flagx"
 	"github.com/ory/x/logrusx"
 
-	"github.com/gorilla/context"
 	"github.com/julienschmidt/httprouter"
-	negronilogrus "github.com/meatballhat/negroni-logrus"
+	"github.com/meatballhat/negroni-logrus"
 	"github.com/rs/cors"
 	"github.com/spf13/cobra"
 	"github.com/urfave/negroni"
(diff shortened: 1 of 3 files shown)


aaslamin:
Would be nice if we had nightly (or weekly if it is $$$$$) load tests for common workflows running in CI so we can catch major regressions easier.

Note: not offering any solutions to 🔝yet, just something to think about!

aeneasr:
So we do have a benchmark thingy running but it doesn't check for memory leaks and it also didn't run against the OAuth2 Auth endpoint. Maybe we can enhance that script to include that?

PR #1376, opened 2019-04-21
Implement OpenID Connect Front-/Backchannel logout

aeneasr:


haf:
Why do you commit with `u` as your commit message?

Issue #1176, opened 2018-11-19
metrics: Add prometheus metrics


aeneasr:
* https://github.com/zbindenren/negroni-prometheus
* https://github.com/slok/go-prometheus-middleware

kminehart:
I started typing up a comment and then accidentally closed my browser. :(

I'll be working on this one over the next couple weeks.

I'll be using https://github.com/slok/go-http-metrics to instrument basic HTTP metrics like request count, duration, response statuses. That should be sufficient for measuring Hydra's performance and reliability.

I'll be using the normal prometheus go client to instrument some hydra specific metrics.

I'll need some input as to what metrics should be measured.

Some ideas (after discussing with @aeneasr ):

* # of clients
* # of JWTs
* # of access tokens
* # of active refresh tokens
* # of refresh tokens
* # of authorize codes
* # of id tokens

---

Implementation:

* We need to be careful [not to overuse labels](https://prometheus.io/docs/practices/instrumentation/#do-not-overuse-labels).
* Querying for these metrics in the databbase might cause slow responses, and high scrape intervals could cause data to not be returned at all.
  * Keeping up with this data as it's created / deleted instead of querying the databbase could cause problems with synchronization, especially if Hydra is restarted.

Some good articles by one of the Prometheus authors:

* https://www.robustperception.io/why-predeclare-metrics
* https://www.robustperception.io/existential-issues-with-metrics

Personally, after doing some thinking, as long as we're smart about how the metrics are retrieved from the databbase, then I see no reason why querying the database would be a huge issue. :)

aeneasr:
Sounds great! I think there are a couple of metrics we'd be able to extract:

- Client:
  - Number of clients in the datastore
- JWK/Keys:
  - Number of JWKs in the datastore
- User login & consent:
  - Number of unique users that performed the flow in the last X m/h/d/w/y
- OAuth 2.0:
  - Number of total active access tokens
  - Number of total refresh tokens (a refresh token can not be inactive as it will be removed from the datastore once used)
  - Sum (increase/counter) of access tokens issued overall
  - Sum (increase/counter) of refresh tokens issued overall
  - Sum (increase/counter) of ID tokens issued overall
  - Number of token revokations for access tokens and refresh tokens
  - Number of grant types performed (authorize code, implicit, hybrid, client credentials)
  - Which scope(s) have been requested and which scope(s) have been authorized

aeneasr:
[...]

PR #1381, opened 2019-04-23
Prometheus metrics

kminehart:
## Related issue

#1176

## Proposed changes

Add custom Prometheus metrics and HTTP middlewares for request count, duration, and size.

## Checklist

force push.

## Further comments

This is still a work in progress. Wanted to make sure that I was heading in the right direction code quality-wise. :)


consent/handler.go
```diff
@@ -248,6 +249,7 @@ func (h *Handler) GetLoginRequest(w http.ResponseWriter, r *http.Request, ps htt
 
 	request, err := h.r.ConsentManager().GetLoginRequest(r.Context(), challenge)
 	if err != nil {
+		log.Println(err)
```
aeneasr: :)

consent/handler.go
```diff
@@ -298,13 +300,21 @@ func (h *Handler) AcceptLoginRequest(w http.ResponseWriter, r *http.Request, ps
...
+	)
 	d := json.NewDecoder(r.Body)
 	d.DisallowUnknownFields()
-	if err := d.Decode(&p); err != nil {
+	if err = d.Decode(&p); err != nil {
 		h.r.Writer().WriteErrorCode(w, r, http.StatusBadRequest, errors.WithStack(err))
 		return
 	}
+
+	defer func(p *HandledLoginRequest, err error) {
```
aeneasr: A particular reason why this is a defer?

consent/metrics.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	for _, v := range cr.GrantedAudience {
+		m.PrometheusMetrics.ConsentRequestAudiences.With(prom.Labels{
+			"audience": v,
+			"type":     "granted",
+		}).Inc()
+	}
+
+	if cr.ConsentRequest != nil {
```
aeneasr: Can we send something when this is nil? Under what circumstances would this be nil?

metrics/prometheus/metrics.go
```diff
@@ -1,42 +1,120 @@
...
+	pm := &Metrics{
+		HTTPRequestCount: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "http_request_count",
+			Help: "",
+		}, []string{"code", "method"}),
+		HTTPRequestDuration: prometheus.NewHistogramVec(
+			prometheus.HistogramOpts{
+				Name:    "http_request_duration_seconds",
+				Help:    "A histogram of latencies for requests",
+				Buckets: []float64{.05, .25, .5, 1, 2.5, 5, 10},
```
aeneasr: Is it possible to create auto buckets here? So that prometheus chooses the right bucket size and median itself?

metrics/prometheus/metrics.go
```diff
@@ -1,42 +1,120 @@
...
+			},
+			[]string{"method"},
+		),
+		HTTPResponseSizeBytes: prometheus.NewHistogramVec(
+			prometheus.HistogramOpts{
+				Name:    "http_response_size_bytes",
+				Help:    "A histogram of response sizes for requests",
+				Buckets: []float64{200, 400, 800, 1600},
```
aeneasr: Same question here

metrics/prometheus/metrics.go
```diff
@@ -1,42 +1,120 @@
...
+			},
+			[]string{},
+		),
+		ConsentRequests: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Namespace: "hydra",
+			Subsystem: "consent",
+			Name:      "requests",
+			Help:      "incremented when a request is sent to consent.AcceptConsentRequest",
```
aeneasr: What about `reject`?

consent/handler.go
kminehart: :grimacing:

metrics/prometheus/metrics.go
kminehart: I created a separate metric for `RejectConsentRequest` [here](https://github.com/ory/hydra/pull/1381/files/b0b492c3823b70296b90eca69d2872746595b60e#diff-01645a5b90df989c7ec7d047661882dfR90)

metrics/prometheus/metrics.go
kminehart: No unfortunately. It actually can become a problem with visualization if you have a sample that changes around its bucket sizes over time, so it's important that we at least have these values set the way we want them the first time around.

consent/metrics.go
kminehart: I actually am not entirely sure what would cause it to be nil, but when I was doing some basic oauth2 flows using [this](https://www.ory.sh/docs/deprecated/latest/hydra/1-tutorial/), it ended up being `nil` for me and causing a panic.

metrics/prometheus/metrics.go
kminehart: Also worth noting, the lower bucket count, the better. Querying histograms, especially over a large time range can be pretty taxing on the Prometheus server. We run into this issue at work now where a histogram will time out if queried over a 30 day span because there are too many buckets.

The histogram will include `count` and `sum`, so no data will really be missing.

consent/handler.go
kminehart: The alternative is to call `recordAcceptLoginRequest(h.r.PrometheusManager(), p, err)` before every return. I thought the defer would work nicely here.

metrics/prometheus/metrics.go
kminehart: Unless you are referring to this: :sweat_smile: 

https://godoc.org/github.com/prometheus/client_golang/prometheus#LinearBuckets

https://godoc.org/github.com/prometheus/client_golang/prometheus#ExponentialBuckets

consent/metrics.go
kminehart: I found out that this field is just never set in the handler, even though the ConsentRequest is retrieved from the database.

consent/handler.go
aeneasr: Ah I see, since you're passing the error down to prometheus you want that in there right?

metrics/prometheus/metrics.go
aeneasr: Hm yeah I think neither linear nor exponential make sense here.

metrics/prometheus/metrics.go
aeneasr: What are those numbers based on?

consent/handler.go
kminehart: exactly! I still need to make sure that works as I expect it to; apparently the way that defer is evaluated is a little weird, which is why its an anonymous function and not just directly calling `recordConsentRequest`.

kminehart:
So this will be a fun PR to merge with all them conflicts. Hopefully we can just force the merge when it's since I forgot to sign off on a merge commit.

Highlights:

* First class Prometheus support.
  * Rather than implement a metrics interface or some kind of abstraction layer for metrics, I propose first class Prometheus support.
    * The Prometheus format is [preparing to reach [an IETF RFC standard](https://github.com/OpenObservability/OpenMetrics), which means it's a suitable format for most services.
    * The Prometheus format is supported by Sysdig, Datadog, and InfluxDB (As a remote-write target).
  * The Prometheus client library is an excellent tool for receiving metrics ad-hoc (in realtime, like `number of requests`) and for watching for metrics at intervals (things like "# of `x` in the database), and comes bundled with good HTTP middleware and instrumentation.

* The `Manager` interfaces now also implement the [prometheus.Collector](https://github.com/supplypike/metrics).
  * Once registered with the Registry ([here](https://github.com/kminehart/hydra/blob/prometheus-metrics/driver/registry_base.go#L417)), the client will continuously retrieve metrics for that manager. This is where SQL queries for "COUNT" and such will happen.
  * For metrics not retrieved by `Collect`, they are incremented as-need / in real-time, and then sent to the Registry in `Collect`.

* A new interface, `metrics.Bridge` was created. This interface is for pushing metrics to external metric services that may not yet accept the Prometheus format.

* The `metrics.BridgeManager` was created to manage these individual `Bridge` instances, repeatedly pushing metrics to each external service provided. It is initialized [here](https://github.com/kminehart/hydra/blob/prometheus-metrics/cmd/server/handler.go#L234)

* The Segment stuff in `ory/x/metricsx` was moved (mostly) to `ory/hydra/metrics` as the [`SegmentBridge`](https://github.com/kminehart/hydra/blob/prometheus-metrics/metrics/bridge_segment.go)

**Todo**:

* [ ] Include path in HTTP metrics middleware (use ory/x/metricsx code)
* [ ] Implement the SQL queries for metrics in the managers
* [ ] Clean up & comment code. Add package-level comments and comments for exported functions.

aeneasr:
Nice! What should I look out for when reviewing the WIP changes?

Regarding Segment, it is very important that the existing fields, keys, and events stay unchanged. I also want to avoid to move stuff from `x/metrics` back here as `x/metrics` is standardized across all repos and we do not want to diverge per project.

kminehart:
Gotya. That is actually the kind of information I was looking for! I'm looking for general ideas more than anything, like maybe, "this stuff would be better in ory/x", or "this isn't suitable for the manager" kind of stuff.

I'll make sure that the fields, keys, and events stay the same.

aeneasr:
Ok, perfect!

Would it make sense to have the bridge in x?

kminehart:
> Would it make sense to have the bridge in x?

I was actually thinking the same thing!

I went ahead and did that.

In order for this to pass, it'll need this: https://github.com/ory/x/pull/49

client/metrics.go
```diff
@@ -0,0 +1,14 @@
+package client
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
+	metricClients = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: "hydra",
```
aeneasr: Would it make sense to have the namespace fixed as a const somewhere? We should also decide if it's `ory-hydra`, `ory_hydra`, `ORY Hydra`, `hydra` or any other combination.

cmd/server/handler.go
```diff
@@ -77,6 +82,7 @@ func RunServeAdmin(version, build, date string) func(cmd *cobra.Command, args []
 	return func(cmd *cobra.Command, args []string) {
 		d := driver.NewDefaultDriver(
 			logrusx.New(),
+			prometheus.NewRegistry(),
```
aeneasr: If this is always `prometheus.NewRegistry()`, wouldn't it make more sense to create that within the `registry_base`?

cmd/server/handler.go
```diff
@@ -174,10 +182,35 @@ func setup(d driver.Driver, cmd *cobra.Command) (admin *x.RouterAdmin, public *x
...
-		cmd,
-		d.Registry().Logger(),
-		&metricsx.Options{
+	var (
+		bridges = []metricsx.Bridge{}
+		l       = d.Registry().Logger()
+		ctx     = context.Background()
+	)
+
+	optOut, err := cmd.Flags().GetBool("sqa-opt-out")
```
aeneasr: Do we need all of this here or could we hide that in `metricsx` as it was previously?

cmd/server/handler.go
```diff
@@ -226,14 +258,27 @@ func setup(d driver.Driver, cmd *cobra.Command) (admin *x.RouterAdmin, public *x
...
 				"/",
 			},
+			WriteKey:     "h8dRH3kVCWKkIFWydBmWsyYHR4M0u0vr",
 			BuildVersion: d.Registry().BuildVersion(),
 			BuildTime:    d.Registry().BuildDate(),
 			BuildHash:    d.Registry().BuildHash(),
-		},
-	)
+		}
+		b, err := metricsx.NewFormattedSegmentBridge(ctx, s, l, d.Registry().PrometheusManager().Registry)
```
aeneasr: Ok I see - so segment actually works (mostly) independent of Prometheus except for e.g. memory stats?

client/manager_sql.go
```diff
@@ -351,9 +352,56 @@ func (m *SQLManager) GetClients(ctx context.Context, limit, offset int) (clients
 func (m *SQLManager) CountClients(ctx context.Context) (int, error) {
 	var n int
 	if err := m.DB.QueryRow("SELECT count(*) FROM hydra_client").Scan(&n); err != nil {
-		fmt.Println(err.Error())
```
aeneasr: 😅

consent/handler.go
```diff
@@ -313,13 +313,22 @@ func (h *Handler) AcceptLoginRequest(w http.ResponseWriter, r *http.Request, ps
...
 
-	var p HandledLoginRequest
+	var (
+		p   HandledLoginRequest
+		err error
+	)
+
+	defer func(p *HandledLoginRequest, err error) {
+		recordAcceptLoginRequest(p, err)
+	}(&p, err)
```
aeneasr: Relying on a global error is very flaky - there are several places where err will be overshadowed, e.g. [here](https://github.com/ory/hydra/pull/1381/files#diff-ac4ce4ab0ca6c69035745a50d3108c65R337) and [here](https://github.com/ory/hydra/pull/1381/files#diff-ac4ce4ab0ca6c69035745a50d3108c65R354) and [here](https://github.com/ory/hydra/pull/1381/files#diff-ac4ce4ab0ca6c69035745a50d3108c65R360). I'm not sure about other projects but I feel that multiple return values usually re-define/overshadow the original err scope:

```
package main

import (
	"fmt"
)

func main() {
	a := "a"
	fmt.Println(a) // "a"
	
	b, a := "b", "c"
	fmt.Println(a) // "c"
	fmt.Println(b) // "b"
}
```

kminehart:
I'm gonna close this and re-open a new one when its ready.

PR #1407, opened 2019-05-01
ttl is a top-level config value

MDrollette:
Viper is looking for ttl at the top-level. Update these docs to reflect reality.


aeneasr:
Thank you!

Issue #1447, opened 2019-05-20
Support Jaeger Tracing with Istio and Kubernetes


ptescher:
**Is your feature request related to a problem? Please describe.**

The Opentracing integration is nice but it does not support a few features needed to integrate with Istio (at least not with the standard Istio setup). Istio uses Zipkin b3 headers which are supported by jaeger-client-go but not by default. A standard Istio/Kubernetes setup will also run jaeger-agent as a daemon set and in order to get the IP address you usually do `fieldPath: status.hostIP` but that only sets the host, not the port.

**Describe the solution you'd like**

- [x] Support for separate JAEGER_AGENT_HOST env var (https://github.com/jaegertracing/jaeger-kubernetes/blob/master/README.md#jaeger-components).
- [ ] Support Zipkin B3 headers (https://github.com/jaegertracing/jaeger-client-go/blob/master/zipkin/README.md#NewZipkinB3HTTPHeaderPropagator) potentially via the `JAEGER_PROPAGATION` env var.

**Describe alternatives you've considered**

Using a custom bash script to initialize hydra.

**Additional context**

The Jaeger Go client already supports a bunch of env vars, it would be nice to support all of them: https://www.jaegertracing.io/docs/1.12/client-features/

aeneasr:
Related #1442

aeneasr:
Did the related issue resolve the Zipkin B3 issue too? If not, would you be open for a PR?

ptescher:
@aeneasr it did not, the go framework doesn't want to expose the B3 via an env var because it would require that the configuration module be linked to other modules. Should be an easy PR though.

aeneasr:
Would you be up for a PR?

ptescher:
Yeah I can take a shot at it.

PR #1456, opened 2019-06-01
tracing: Add support for B3 headers via the JAEGER_PROPAGATION env var

ptescher:
This will provide compatibility with istio.

Fixes https://github.com/ory/hydra/issues/1447

### DIFF
--- a/tracing/tracer.go
+++ b/tracing/tracer.go
@@ -8,6 +8,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	jeagerConf "github.com/uber/jaeger-client-go/config"
+	"github.com/uber/jaeger-client-go/zipkin"
 )
 
 type Tracer struct {
@@ -25,6 +26,7 @@ type JaegerConfig struct {
 	SamplerType        string
 	SamplerValue       float64
 	SamplerServerUrl   string
+	Propagation        string
 }
 
 func (t *Tracer) Setup() error {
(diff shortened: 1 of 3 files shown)


tracing/tracer.go
```diff
@@ -52,8 +54,21 @@ func (t *Tracer) Setup() error {
 			jc.Reporter.LocalAgentHostPort = t.JaegerConfig.LocalAgentHostPort
 		}
 
+		var configs []jeagerConf.Option
+
+		// This works in other jaeger clients, but is not part of jaeger-client-go
+		if os.Getenv("JAEGER_PROPAGATION") == "b3" {
```
aeneasr: Please, instead:

1. Add `Propagation string` to [`JaegerConfig`](https://github.com/ory/hydra/pull/1456/files#diff-6aa771dfea3dd559e0b2c9263ebf1ccdR25)
2. Use the `Propagation` value here (`t.JaegerConfig`)
3. Set the value [here](https://github.com/ory/hydra/blob/master/driver/configuration/provider_viper.go#L243)
4. Document the new value [around here](https://github.com/ory/hydra/blob/master/docs/config.yaml#L399)
To stay compatible it probably makes sense to do something like (assuming `JAEGER_PROPAGATION` is a de-facto standard):

```
Propagation: stringsx.Coalesce(
  viper.GetString("JAEGER_PROPAGATION")
  viperx.GetString(v.l, "tracing.providers.jaeger.propagation", ""),
),
```

ptescher: Pushed some new changes, let me know if that works. I also added `TRACING_PROVIDER_JAEGER_PROPAGATION` just for consistency.

aeneasr:
Your changes to ory/x are now released as v0.0.61

aeneasr:
ps: If you rebase on master, the module changes should be up to 0.0.61

ptescher:
@aeneasr all done, tests are passing now.

aeneasr:
Awesome, thank you!

PR #1480, opened 2019-06-23
Fix namespace in composer.json (again?)

DASPRiD:
PR https://github.com/ory/hydra/pull/1468 changed the namespace within the PHP files from `Hydra\SDK` to `HydraSDK`, but it did not update the autoload section in the `composer.json` file.

This seems to happen quite often recently, I wonder if there should be a test in place to avoid having a mismatch there **ever**.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,6 @@
     "ext-mbstring": "*"
   },
   "autoload": {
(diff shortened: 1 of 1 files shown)


aeneasr:
It seems like that particular PR did not run `make sdk`, causing the composer.json file to not be updated. Running that command `make sdk` would be the right course of action for this PR imo.

PR #1495, opened 2019-07-19
 cmd: Improve config loader logic

aeneasr:


aeneasr:
This PR is no longer relevant as we have a new approach to loading, validating config.

Issue #1557, opened 2019-09-08
Potential bug in remember logic for login when login is skipped


doubliez:
**Describe the bug**

When login is skipped and I call "accept login request" with only a subject (without remember params), the `oauth2_authentication_session` cookie is reset to be a session cookie instead of maintaining the same expiry date it had on first login (based on `remember_for` value).

**Reproducing the bug**

Should be reproducible with the sample login-consent app provided by Hydra, which is what I based my implementation on.

In the sample code here: https://github.com/ory/hydra-login-consent-node/blob/f4605748c2500f113813bc87b21c4875fc04694d/routes/login.js#L30

Only the subject is passed when accepting the login request (in case it was skipped). This for me causes the `oauth2_authentication_session` cookie to be reset to a session cookie. If I specify the remember params, the lifetime of the cookie is set accordingly. However for skipped login I don't want to have to re-specify remember params, it should not touch the cookie at all in my opinion.

**Expected behavior**

When accepting a skipped login request, the cookie used to remember the user should not be altered, and maintain its expiry date if it was set on first login through `remember_for` param.

**Environment**

* Version: oryd/hydra:v1.0.0 (Docker image)
* Environment: Docker

aeneasr:
This is intended behavior. If a user is requested to sign in again, and remember for is not set, then the cookie has to be overwritten as the user now chose not to remember the login. Not handling according to this logic would keep the cookie around even though the user chose not to stay logged in.

doubliez:
@aeneasr In this case the user is not explicitly requested to sign in again, there is no user interaction. I'm using a `prompt=none` silent refresh within a hidden iframe in my SPA, and this behavior means that even though on the first (fresh) login the user chose to remember its session, when the first silent refresh occurs this will override the decision of the user...

For me that's unintended behavior. When login is skipped, Hydra should not override the initial decision of the user to remember the login. This defeats the purpose of the "remember me" feature when using silent refresh.

But maybe I'm doing something wrong?

aeneasr:
It’s not possible for hydra to know that. You can set remember to !skip || request.form.remember (pseudocode) to achieve the behavior you want. I believe this is documented but if not it definiteöy shouöd be!

doubliez:
To be clear, on first login, the session is remembered for 30 days using:
```
hydra.acceptLoginRequest(challenge, {
    subject: user.id,
    remember: true,
    remember_for: 2592000
});
```

(when the user checks the "Remember me" checkbox)

[...]

PR #1564, opened 2019-09-18
Don't touch authentication cookie on skipped logins

doubliez:
## Related issue

This will fix https://github.com/ory/hydra/issues/1557

## Proposed changes

When login was skipped, return the session and don't touch the existing authentication cookie (which would alter its expiry date).

## Checklist


aeneasr:
Oh and please sign the CLA :)

aeneasr:
Ping :)

doubliez:
I signed the CLA. I'll try to add a test this week. I assume there was no test for this specific case then? ;)

aeneasr:
Perfect - yes exactly ;)

doubliez:
@aeneasr I updated the docs regarding RememberFor value for login.

Regarding the tests I was thinking of retrieving the auth cookie from the cookie jar and making sure the `Expires` value is set correctly, but it seems the `net/http/cookiejar` package doesn't allow inspecting those values. When retrieving cookies with `cookiejar.Cookies(u)` it returns only the `Name` and `Value` attributes for the cookies: https://github.com/golang/go/blob/master/src/net/http/cookiejar/jar.go#L219
```golang
for _, e := range selected {
	cookies = append(cookies, &http.Cookie{Name: e.Name, Value: e.Value})
}
```

So I'm not sure how to properly test this.

aeneasr:
Ah, I see - I think that's why the tests didn't catch it, because the Cookie Jar doesn't really concern itself with sessions like the browser does.

I think the only way to get the cookie value is to somehow use the gorilla session decoder to get it (I've tried before but failed but it should in theory be possible IMO) or write a small endpoint and add it to the test server that simply echoes the cookie values as e.g. a json string.

doubliez:
@aeneasr Since I'm not too familiar with Go and don't have much time to work on this, can you take it from there?

aeneasr:
SGTM - thank you for your contribution!

Issue #1576, opened 2019-09-23
Fix homebrew release pipeline


aeneasr:
Upstream: https://github.com/goreleaser/goreleaser/issues/1146

aeneasr:
I'm pretty sure this is a bug with goreleaser itself - please make sure to revert https://github.com/ory/hydra/commit/b6216940dc932469e678e30a13ddaa9c8bd889c8

aeneasr:
This has been resolved upstream

PR #1582, opened 2019-09-26
ci: Remove ARM binaries

aeneasr:
This is a workaround for #1576

ARM is still supported but must be built from source.


aeneasr:
This has been resolved and is no longer required

Issue #1588, opened 2019-10-01
Additional Claims


ihippik:
**Is your feature request related to a problem? Please describe.**

We need some fields that are not part of the UserInfo response, but they can be added according to the OpenID specification.
https://openid.net/specs/openid-connect-core-1_0.html#AdditionalClaims

**Describe the solution you'd like**

I want to add ExtraVars (map [string] intreface {}) field to UserInfo response

aeneasr:
See comment https://github.com/ory/hydra/pull/1589#issuecomment-537006111

aeneasr:
Upstream: https://github.com/go-swagger/go-swagger/issues/2073

aeneasr:
This apparently can be solved with the `additionalProperties` directive

vinckr:
Hey @ihippik does the `additionalProperties` directive solve this for you? 
If so we can close this issue?

ihippik:
Hi! Yes you can close, thanks!  We decided a little differently.

vinckr:
close this @aeneasr  ?

aeneasr:
No, this still needs to be fixed in the code.

vinckr:
Ok, but what needs to be fixed?
Sorry, I might have missed something.
Is `additionalProperties` that needs to be fixed?

PR #1589, opened 2019-10-01
Additional claims

ihippik:
Add ExtraVars into userInfo response.

## Related issue

https://github.com/ory/hydra/issues/1588

## Proposed changes

add ExtraVars (map [string] intreface {}) field to UserInfo response

## Checklist

### DIFF
--- a/docs/api.swagger.json
+++ b/docs/api.swagger.json
@@ -2805,6 +2805,14 @@
           "type": "boolean",
           "x-go-name": "EmailVerified"
         },
+        "extra_vars": {
+          "description": "Addition claims (https://openid.net/specs/openid-connect-core-1_0.html#AdditionalClaims).",
+          "type": "object",
+          "additionalProperties": {
+            "type": "object"
+          },
+          "x-go-name": "ExtraVars"
+        },
         "family_name": {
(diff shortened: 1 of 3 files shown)


aeneasr:
So this is already possible (from Hydra, not the SDK). The way this is formatted in swagger in this PR is however not the way this actually works. In fact, the swagger response should contain some well-defined fields (like e.g. `sub`) but additionally contain other values as well. **These values should not be encoded in an arbitrarily defined field like `extra_envs`.**. Instead, this should probably be solved with the appropriate swagger annotation that would allow bespoke responses with a huge variety of fields.

However, we're very explicit on our SDK not being a good substitute for performing OAuth2 and OpenID Connect-related flows. The SDK works very well with ORY Hydra's Admin APIs, it however does not work well with the public APIs (such as well known, userinfo, /oauth2/token, /oauth2/auth). 

The SDK is usually used by the OpenID or OAuth2 provider, while the actual endpoints like userinfo are being used by the OAuth2 and OpenID Consumer, which are third parties, and do not know what Hydra is or does let alone use the SDKs from us.

ihippik:
@aeneasr thanks for the answer. And how do I use fields that are not explicitly specified in the sdk client? Use your client?

aeneasr:
https://openid.net/developers/certified/

Issue #1585, opened 2019-09-28
DSN credential is shown in log


oligo:
**Describe the bug**

When hydra starts, it prints database connection information in the log. These log records have DSN info in them which keep the username and password unmasked.

**Reproducing the bug**

1. Run `hydra serve all` with a MySQL DSN configured
2. Check the server startup log.

*Server logs*

```
time="2019-09-28T02:42:37Z" level=info msg="Establishing connection with SQL database backend" dsn="username:password@tcp(dbip:3306)/hydradb?parseTime=true://*:*@tcp(dbip:3306)/hydradb?parseTime=true"
time="2019-09-28T02:42:37Z" level=info msg="Successfully connected to SQL database backend" dsn="username:password@tcp(dbip:3306)/hydradb?parseTime=true://*:*@tcp(dbip:3306)/hydradb?parseTime=true"
```

**Expected behavior**

The DSN username and password should not be shown in plaintext.

**Environment**

* Version: v1.0.1+oryOS.12, 937cb2e473c525d4e546bf34c5be1dd8ffcade28
* Environment: Docker

aeneasr:
Thank you for reporting, this has been resolved in 1.0.2+

PR #1593, opened 2019-10-03
driver: don't log DSN

MDrollette:
Removes an errant logging of the DSN. The DSN can include credentials which should not be logged.

related to https://github.com/ory/hydra/issues/1585 and https://github.com/ory/k8s/issues/55


aeneasr:
Thank you!

Issue #1615, opened 2019-10-22
CORS allowed_origins not working as expected


davideferre:
After many retries I found that _Origin_ header is requested to get correct CORS headers.
In one of this retry in my configuration file I've specified _allowed_origins_ as _http://*_ and I've created a client with _allowed_cors_origins: ["http://*"]_. When I call (from postman) _./well-known/jwks.json_ url specifying _Origin: http://127.0.0.1:4200_, CORS headers are missing, but when I specified _Origin: http://*_ it worked.

**Reproducing the bug**
I've tried with 1.0.2 version and 1.0.8 version but the bug is still present. See attached configuration.

*Server configuration*

cors:
      enabled: true
      allowed_origins:
        - http://*
      allowed_methods:
        - POST
        - GET
        - PUT
        - PATCH
        - DELETE
      allowed_headers:
        - Authorization
        - Content-Type

**Environment**

* Version: v1.0.8
* Environment: Ubuntu, Docker

**Additional context**

For get it working I've to specify _allowed_origins: http://localhost:4200_ in server configuration and _"allowed_cors_origins": ["http://127.0.0.1:4200"]_ into the client.

aeneasr:
Yes, that's how CORS works: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS

Without `Origin` in the HTTP header, the whole CORS idea is baseless. If you use Postman to emulate a browser, you need to behave like a browser - which, when using CORS, consists of (among others) setting the right `Origin` header.

davideferre:
I'm not sure if hydra is working properly because if I specify **allowed_origins** as **http://*** I would expect that with **Origin:http://127.0.0.1:4200** headers the CORS are working fine but this didn't happen.

aeneasr:
Okay, I think I misundestood what you said! So it appears that `http://*` is parsed as a literal string as opposed to a wildcard when using `allowed_cors_origins` in the client?

davideferre:
Correct!

aeneasr:
Got it, thanks! Triaging as bug/client.

What I don't really understand is, we actually have a test case covering your bug report:

https://github.com/ory/hydra/blob/95a51deb3100034db5c6d98bbd7838a3b43249ce/driver/cors_test.go#L92-L100

https://github.com/ory/hydra/blob/95a51deb3100034db5c6d98bbd7838a3b43249ce/driver/cors_test.go#L92-L100

[...]

PR #1623, opened 2019-10-31
possible #1615 solution: partial wildcard allowed-origins, with just …

Aterocana:
…protocol specified (for example "http://*") could use glob ** pattern in order to match any sequence of characters (even the separator).

Signed-off-by: Aterocana <dominicimaurizio<at>gmail.com>

## Related issue
`@aeneasr` as discussed in #1615 I'm proposing this fix.

## Proposed changes

Basically I think the issue here is the protocol is specified, but no domain is: for example https://*.
Internally glob.Glob is compiled (with glob.Compile) with the provided string and `'.'` rune separator. Since `g:=glob.Compile("https://*", '.')` doesn't match `g.Match("http://foobar.com")` due to `.` separator, we could use `g:=glob.Compile("https://**", '.') to ignore `.` separator in this specific case.
In order to do so I simply tried to match if "://" substring is into an allowed-origin string.

## Checklist

- [x] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md)
- [x] I have read the [security policy](../security/policy)
- [x] I confirm that this pull request does not address a security vulnerability. If this pull request addresses a security
vulnerability, I confirm that I got green light (please contact [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push the changes.
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation within the code base (if appropriate)
- [ ] I have documented my changes in the [developer guide](https://github.com/ory/docs) (if appropriate)

PR #1624, opened 2019-11-01
Using glob ** pattern in order to match partial wildcard with protoco…

Aterocana:
…protocol specified (for example "http://*") could use glob ** pattern in order to match any sequence of characters (even the separator).

## Related issue
`@aeneasr` as discussed in #1615 I'm proposing this fix.

## Proposed changes

Basically I think the issue here is the protocol is specified, but no domain is: for example https://*.
Internally glob.Glob is compiled (with glob.Compile) with the provided string and `'.'` rune separator. Since `g:=glob.Compile("https://*", '.')` doesn't match `g.Match("http://foobar.com")` due to `.` separator, we could use `g:=glob.Compile("https://**", '.') to ignore `.` separator in this specific case.
In order to do so I simply tried to match if "://" substring is into an allowed-origin string.

## Checklist

### DIFF
--- a/driver/cors.go
+++ b/driver/cors.go
@@ -22,6 +22,7 @@ package driver
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


Aterocana:
I'm sorry, but I couldn't (or I wasn't able to :) ) modified previous license/cla, so I created another PR with proper git user. It is the same as https://github.com/ory/hydra/pull/1623

aeneasr:
Awesome, thank you!

Issue #1574, opened 2019-09-23
Need way to housekeep records in DB


22vincetsang:
There are lots of records in Hydra DB that get accumulated over time and are never cleaned up.
The flush API (as I tested with 1.0) can only clean up access tokens as stated in documentation, but not records in at least these tables:

- [ ] hydra_oauth2_authentication_request
- [ ] hydra_oauth2_authentication_request_handled
- [ ] hydra_oauth2_authentication_session
- [ ] hydra_oauth2_code
- [ ] hydra_oauth2_consent_request
- [ ] hydra_oauth2_consent_request_handled
- [ ] hydra_oauth2_pkce
- [ ] hydra_oauth2_refresh

We need at minimum an API to flush all obsolete records, if not an automatic housekeeping mechanism which I understand isn't in place today for access tokens neither.

aeneasr:
Yes, so I checked and flush currently only clears up access tokens. Since this has not ever become an issue even in large deployments so far, I'm pushing the urgency back a bit.

aeneasr:
PKCE is removed when it's being used, but it's still possible that old, dead entries are around if the authorize code is not used.

22vincetsang:
Thanks, that's helpful - and I just verified the normal path won't create a record there. So it should be safe to clean up these pkce records then (still don't yet understand where they came from as the auth code should always be consumed immediately by our mobile app/spa... maybe there's some error case I'm not aware of...).

Looking forward to this enhancement so other DB tables can be handled too. It just seems not that trivial by looking at the other tables to identify which records are safe to delete so I dare not do so... 

Thanks again!

pereslava:
Hello. Any progress with it? 
I was in milestone v1.3.0, I'm running v1.4.6 now and still no API.

As described [here](https://github.com/ory/hydra/pull/1639#issuecomment-557050034) it is safe to cleanup the tables manually. But PR was not merged

aeneasr:
Contributions are welcomed, I think the threads and comments and existing PRs shed light on what needs to be done!

pereslava:
@aeneasr 
Lets discuss what do you mean [here](https://github.com/ory/hydra/pull/1639#issuecomment-557050034):

[...]

PR #1639, opened 2019-11-20
[oauth2] Flush refresh token/ auth code/ id token in FlushHandler 

genchilu:
## Proposed changes
Current FlushHandler in oauth2 only flush access token, but I think it should flush other tokens also.
ref: https://github.com/ory/hydra/issues/1574

## Checklist

### DIFF
--- a/oauth2/fosite_store_memory.go
+++ b/oauth2/fosite_store_memory.go
@@ -145,6 +145,11 @@ func (s *FositeMemoryStore) DeleteOpenIDConnectSession(_ context.Context, author
 	return nil
 }
 
+func (s *FositeMemoryStore) deleteOpenIDConnectSession(_ context.Context, authorizeCode string) error {
+	delete(s.IDSessions, authorizeCode)
+	return nil
+}
+
 func (s *FositeMemoryStore) CreateAuthorizeCodeSession(_ context.Context, code string, req fosite.Requester) error {
 	s.Lock()
 	defer s.Unlock()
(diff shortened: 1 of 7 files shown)


aeneasr:
Thank you! Unfortunately, this can not be merged as is. Removing refresh tokens will cause the refresh token chain to fail. It's important to keep track of all issued refresh tokens and access codes to be able to properly revoke them and their children later on. I believe this is true for OIDC as well.

We remove, iirc, used refresh tokens from the store anyways.

genchilu:
It's fine.
DBA tell me that the size of tables is growing too fast, and all I can do now is delete data through pure SQL, that's not reasonable. That is why I submit this PR.

My original plan was to implement another method to clean expired data in hydra_oauth2_xxx_request  after this PR  merged. But I think that maybe not a good idea now, since it may cause the refresh token chain to fail, too.

Is there any plan to implement such kind of method to do housekeeping job? Or what may I help to make it happen?

aeneasr:
Thanks for getting back to me. Which tables are growing too fast exactly? Maybe there is a missing cleanup job.

genchilu:
hydra_oauth2_authentication_request is the table growing up the most.
The ratio of numbers of raw data in `hydra_oauth2_authentication_request`, `hydra_oauth2_authentication_session` and `other session/request/token table` is about to 4:2:1.

aeneasr:
Ah ok I see - that makes sense - I think we can clean up those tables (and they are not properly being cleaned up at the moment), and it would not actually do a lot if we cleaned up the others instead.

I have to check if we're safe deleting those requests but I do believe so. The problem is that everytime someone calls `/oauth2/auth`, we create a new session and request, even if the user aborts the flow. That's why these are growing and why it's safe to remove those if they are simply "dead" requests. Would you be open to giving this a shot?

genchilu:
Great, I will try it.
And please let me know if you make sure it's 100% safe deleting those requests, I really want to clean up DB through API, but not SQL in another system.

Thx a lot.

aeneasr:
So from what I can tell right now, you can delete:

* All `hydra_oauth2_(authentication|consent)_request` that do not have a result for `hydra_oauth2_(authentication|consent)_request_handled.challenge = hydra_oauth2_(authentication|consent)_request.challenge` and where `hydra_oauth2_(authentication|consent)_request.requested_at` is older than a certain threshold.
* All `hydra_oauth2_logout_request` where `accepted=false`, `rejected=false`, `was_used=false` and (**this field needs to be added**) `requested_at` is older than a certain threshold.
* All `hydra_oauth2_(authentication|consent)_request` where one result exists for `hydra_oauth2_(authentication|consent)_request_handled.challenge = hydra_oauth2_(authentication|consent)_request.challenge` and where `hydra_oauth2_(authentication|consent)_request_handled.was_used=false` and `hydra_oauth2_(authentication|consent)_request.requested_at` is older than a certain threshold.

Issue #1660, opened 2019-12-09
Add option to restrict front-/back-channel logout to RPs associated with the logged out session


tyaps:
**Describe the bug**

If user was previously logged in multiple RPs during some OIDC sessions, when user does logout from current OIDC session, it gets frontChannel logout script with all of RP's URLs, even it was logged to only one RP during current OIDC session.

While doing session logout, when preparing frontChannelUris, sql query uses only subject, but doesn't use sid. 

func (s *DefaultStrategy) generateFrontChannelLogoutURLs(ctx context.Context, subject, sid string) ([]string, error) {
	clients, err := s.r.ConsentManager().ListUserAuthenticatedClientsWithFrontChannelLogout(ctx, subject)

**Reproducing the bug**

Login to several RP's (RP1, RP2) during one OIDC session (remember=true).
Logout this OIDC session.
Login to one RP1 during new OIDC session (remember=true).
Logout this OIDC session.
frontChannel logout script will contain frontChannelLogout URLs list for RP1 and RP2

**Expected behavior**

 frontChannel logout script should contain frontChannelLogout URLs only for RP1

aeneasr:
I believe this is intentional, although I'm not sure. Do you have a link/resource to a spec or implementation guideline where this behavior is defined?

tyaps:
I'll try to find it.
But from common sense I think, that if you were logged in during one OIDC session to RP1, then if you do logout, only that RP1 should be notified. Why some other RPs should know that you logged out...
Ok, I'll search some evidences ))

aeneasr:
I believe that we send the `sid` with the request, that way, the RP can decide if they wish to log the user out based on that sid or not.

tyaps:
Hmm... by the way about sid. I thought that RP doesn't know anything about sid. Sid (OIDC sessionID)  never arrives on RP: when you change code on token there is no any Sid parameters in queryString. RP can't get it somehow.

Sid (that is represented with oauth2_authentication_session cookie) is only between User and OIDC Provider to remember successful login occurrence. Or do I miss something.

(haven't got time to search guidelines, but I'll find some time to investigate soon)

aeneasr:
https://stackoverflow.com/questions/45370324/what-is-sid-claim-in-logout-token-in-openid-connect-back-channel-logout

tyaps:
Ok, thank you. I can get sid from id_token.

But here is still my considerations.
[...]

PR #1673, opened 2020-01-05
Restrict Front/Back Channel logout PRs associated with logged out session.

obasajujoshua31:
## Related issue

Issue [#1660 ](https://github.com/ory/hydra/issues/1660)
## Checklist

I added logging_session_id with the db query script to search for clients from the hydra_oauth2_consent_request table.

- [x ] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md)
- [x ] I have read the [security policy](../security/policy)
- [x ] I confirm that this pull request does not address a security vulnerability. If this pull request addresses a security
- [ x] I have added tests that prove my fix is effective or that my feature works


aeneasr:
Let me know when this is good to review!

obasajujoshua31:
Hi @aeneasr it is ready for review now.

consent/manager_memory.go
```diff
@@ -473,7 +473,7 @@ func (m *MemoryManager) ListUserAuthenticatedClientsWithFrontChannelLogout(ctx c
 	return rs, nil
 }
 
-func (m *MemoryManager) ListUserAuthenticatedClientsWithBackChannelLogout(ctx context.Context, subject string) ([]client.Client, error) {
+func (m *MemoryManager) ListUserAuthenticatedClientsWithBackChannelLogout(ctx context.Context, subject, sid string) ([]client.Client, error) {
```
aeneasr: You have added the `sid` parameter here, but are not actually using it. It should do the same thing that you're doing in the SQL adapter!

consent/manager_test_helpers.go
```diff
@@ -576,21 +576,29 @@ func ManagerTests(m Manager, clientManager client.Manager, fositeManager x.Fosit
...
-					c.Subject = "subjectLUACWFCL"
+
+					fakeLoginSession := LoginSession{
+						ID:              fmt.Sprintf("fk-login-session-LUACWFCL-%d", i),
+						AuthenticatedAt: time.Now(),
+						Subject:         fmt.Sprintf("subjectLUACWFCL-%d", i),
+					}
+					err = m.CreateLoginSession(context.TODO(), &fakeLoginSession)
+					require.NoError(t, err)
+					c.LoginSessionID = fakeLoginSession.ID // otherwise we had to create the login session as well..
```
aeneasr: You can remove the comment because it is no longer accurate

consent/manager_test_helpers.go
```diff
@@ -576,21 +576,29 @@ func ManagerTests(m Manager, clientManager client.Manager, fositeManager x.Fosit
...
+					lc.SessionID = fakeLoginSession.ID
 					require.NoError(t, m.CreateLoginRequest(context.TODO(), lc))
 
 					require.NoError(t, m.CreateConsentRequest(context.TODO(), c))
 					_, err = m.HandleConsentRequest(context.TODO(), c.Challenge, h)
 					require.NoError(t, err)
 				}
 
-				clients, err := m.ListUserAuthenticatedClientsWithFrontChannelLogout(context.TODO(), "subjectLUACWFCL")
+				clients, err := m.ListUserAuthenticatedClientsWithFrontChannelLogout(context.TODO(), "subjectLUACWFCL-5", "fk-login-session-LUACWFCL-5")
```
aeneasr: The test is misleading in my opinion. So what you're doing is you are creating 5 sessions for 5 different identities and assign that to the same client.

This will always return the one client you expect - unless you call for an invalid session (e.g. `I-AM-NOT-fk-login-session-LUACWFCL-5`), which is why all the tests pass here.

What you should do instead is create two identities with two sessions each, and assign each of these sessions to two clients each. So for example:

- identity 1 + session 1 + client id 1
- identity 1 + session 2 + client id 2
- identity 2 + session 3 + client id 3
- identity 2 + session 4 + client id 4

Now you will be checking against those combinations, so basically:

- Does identity 1 and session 1 give client 1?
- Does identity 1 and session 2 give client 2?
- ...

And you will also need a negative test, so:

- Does identity 1 and session 1 give client 2? -> Expected result is no!
- Does identity 1 and session 2 give client 1? -> Expected result is no!

aeneasr: This also applies to the other modified tests!

consent/manager_test_helpers.go
aeneasr: This also applies to the other modified tests!

consent/manager_memory.go
obasajujoshua31: If I remove it, go will say that the consent manager   incorrectly implements `ListUserAuthenticatedWithBackChannelLogout`

consent/manager_test_helpers.go
obasajujoshua31: okay

consent/manager_memory.go
aeneasr: Do not remove it, you need to add the same check as you did in the SQL manager.

obasajujoshua31: you want me to check with the `sid` parameter too right

aeneasr:
Please run `go mod tidy` and re-publish your changes, it seems like there is a checksum error

aeneasr:
Closing in favor of #1691

Issue #1680, opened 2020-01-08
Make Logging Traceable


sawadashota:
**Is your feature request related to a problem? Please describe.**

Jaeger provides a great distributed tracing system but it's not small cost for transition period of monolith to micro service: really few micro services. So now I need traceable logging until I install jaeger.

**Describe the solution you'd like**

Add logging configuration option like following

```yaml
serve:
  admin:
    ...
    additional_logging:
      header_tracing_id: ToTraceID
      header_span_id: ToSpanID
```

Then, reading `ToTraceID` and `ToSpanID` are read from HTTP headers and logging appears like this

```
 {"level":"info",...,"tracing_id": "xxxxxxxxxxxxxxx", "span_id", "xxxxxxxxxxxxxxx"}
```

aeneasr:
I think that's a good idea. Are there any standards we could follow here? I believe we're working with `X-Request-ID` as the HTTP header at the moment.

sawadashota:
Yes, I think `X-Request-ID` is a standard as span id but I didn't come up with counterpart trace id.

To cover both, my example is following OpenTracing specification.
https://github.com/opentracing/specification/blob/584c52365fe0700ba25effa79dce5effb845ced8/rfc/trace_identifiers.md

aeneasr:
This is resolved now :)

PR #1685, opened 2020-01-09
Make logging traceable

sawadashota:
## Related issue

#1680

## Proposed changes

Add Tracing logging option.

```yaml
log:
  tracing:
    public: false
    admin: true
```

Then, hydra logs like this.

```
$ curl -k https://localhost:4445/health/alive -H 'Traceparent: 00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01'
```

```
{"level":"info","method":"GET","msg":"started handling request","remote":"[::1]:56308","request":"/health/alive","time":"2020-01-11T11:09:46+09:00","trace_id":"00f067aa0ba902b7"}
{"level":"info","measure#hydra/admin: https://localhost:4444/.latency":218780,"method":"GET","msg":"completed handling request","remote":"[::1]:56308","request":"/health/alive","status":200,"text_status":"OK","time":"2020-01-11T11:09:46+09:00","took":218780,"trace_id":"00f067aa0ba902b7"}
```

If request doesn't have `Traceparent` header, log doesn't appear `trace_id` and `span_id`.

```
$ curl -k https://localhost:4445/health/alive 
```

```
{"level":"info","method":"GET","msg":"started handling request","remote":"[::1]:56396","request":"/health/alive","time":"2020-01-11T11:27:45+09:00"}
{"level":"info","measure#hydra/admin: https://localhost:4444/.latency":99643,"method":"GET","msg":"completed handling request","remote":"[::1]:56396","request":"/health/alive","status":200,"text_status":"OK","time":"2020-01-11T11:27:45+09:00","took":9964
```

## Checklist


cmd/server/handler.go
```diff
@@ -154,6 +154,29 @@ func RunServeAll(version, build, date string) func(cmd *cobra.Command, args []st
...
+	}
+
+	if len(headers) == 0 {
+		return
+	}
+
+	logger.Before = func(entry *logrus.Entry, r *http.Request, _ string) *logrus.Entry {
+		fields := make(map[string]interface{})
+		for key, headerKey := range headers {
+			fields[key] = r.Header.Get(headerKey)
```
aeneasr: This does not be to be compliant with the open tracing specification: https://github.com/opentracing/specification/blob/584c52365fe0700ba25effa79dce5effb845ced8/rfc/trace_identifiers.md

I would expect that this parses the HTTP Header `traceparent`. I'm pretty sure that opentracing has a library to do that, so maybe we can just use that here?

sawadashota: You're right. And I found library to extract tracing information from header!
https://github.com/open-telemetry/opentelemetry-go

cmd/server/handler.go
```diff
@@ -154,6 +155,21 @@ func RunServeAll(version, build, date string) func(cmd *cobra.Command, args []st
 	}
 }
 
+func setTracingLogger(d driver.Driver, logger *negronilogrus.Middleware) {
+	logger.Before = func(entry *logrus.Entry, r *http.Request, remoteAddr string) *logrus.Entry {
```
aeneasr: This overwrites the existing logger.Before, right? I think we have a custom Before somewhere - Maybe you could check if `logger.Before` is set and if it is, execute it and then execute this logic here, but only adding the `trace_id` and `span_id`?

aeneasr: I see, it appears that we're not actually using that here. Here's the actual function I was referring to: https://github.com/ory/x/blob/master/reqlog/middleware.go

I think it would make sense to use that function and add the trace_id in that middleware. We should also add it to both before, and after!

driver/configuration/provider_viper.go
```diff
@@ -413,3 +415,11 @@ func (v *ViperProvider) ShareOAuth2Debug() bool {
 func (v *ViperProvider) PKCEEnforced() bool {
 	return viperx.GetBool(v.l, ViperKeyPKCEEnforced, false, "OAUTH2_PKCE_ENFORCED")
 }
+
+func (v *ViperProvider) LogTracingPublic() bool {
```
aeneasr: I would be fine with actually having this enabled always no config required

cmd/server/handler.go
aeneasr: Here's an example: https://github.com/ory/oathkeeper/blob/master/cmd/server/server.go#L44

sawadashota: I fixed it!
It seemed `logger.Before` was set default(`request`, `method` and `remote`). I realized it's not safety to consider whether `logger.Before` or not, I fixed to add field.

> We should also add it to both before, and after!

It's seems it's enough to add only before. Actually logs appear `span_id` and `trace_id` at both stating and completed handling request.

```
{"level":"info","method":"GET","msg":"started handling request","remote":"[::1]:56308","request":"/health/alive","time":"2020-01-11T11:09:46+09:00","trace_id":"00f067aa0ba902b7"}
{"level":"info","measure#hydra/admin: https://localhost:4444/.latency":218780,"method":"GET","msg":"completed handling request","remote":"[::1]:56308","request":"/health/alive","status":200,"text_status":"OK","time":"2020-01-11T11:09:46+09:00","took":218780,"trace_id":"00f067aa0ba902b7"}
```

driver/configuration/provider_viper.go
sawadashota: I fixed it!

aeneasr:
Thank you! :)

Issue #1684, opened 2020-01-08
Add authenticated_at to type HandledConsentRequest


marcohutzsch1234:
**Is your feature request related to a problem? Please describe.**

When we ask hydra for all accepted consents for a subject we want to show the subject the date, when the latest consent was accepted. This value exists in the db but is not accessible via go-sdk.

**Describe the solution you'd like**

We had the idea of adding the value to the mentioned struct. Would you accept a PR for that?

**Describe alternatives you've considered**

No idea.

aeneasr:
I think `authenticated_at` is currently not part of the consent itself but only of the authentication session. Do you need the time of the authentication itself, or the time where the consent request was granted?

marcohutzsch1234:
Closing Issue because discussion moved to PR: https://github.com/ory/hydra/pull/1689

PR #1689, opened 2020-01-13
feat: consent: add field for handled at to consent request type and database

DennisPattmann5012:
Co-authored-by: Marco Hutzsch <marco.hutzsch@otto.de>

## Related issue

#1684
## Proposed changes

When a consent is handled hydra stores the result (accepted or rejected). With this change it is possible to access the date and time when a consent was handled.

## Checklist


consent/migrations/sql/shared/14.sql
```diff
@@ -0,0 +1,7 @@
+-- +migrate Up
+ALTER TABLE hydra_oauth2_consent_request ADD handled_at timestamp NULL;
```
aeneasr: What do you think about `created_at` and `updated_at`?

consent/types.go
```diff
@@ -367,6 +367,9 @@ type ConsentRequest struct {
 	// Context contains arbitrary information set by the login endpoint or is empty if not set.
 	Context map[string]interface{} `json:"context,omitempty"`
 
+	// HandledAt contains the timestamp the consent request was handled.
+	HandledAt time.Time `json:"handled_at,omitempty"`
```
aeneasr: This value still needs to be set somewhere, otherwise it will always be nil.

consent/migrations/sql/shared/14.sql
DennisPattmann5012: We decided to name this field `handled_at` because:
- `created_at` could also be the date when the database entry was created
- `updated_at` could also be a change of any other parameter in the request

`handled_at` is more explicit, therefore we named the new field `handled_at`

consent/types.go
DennisPattmann5012: Yes, we missed this and added this to the PR.

consent/migrations/sql/shared/14.sql
aeneasr: That makes sense!

consent/handler.go
```diff
@@ -553,6 +553,7 @@ func (h *Handler) AcceptConsentRequest(w http.ResponseWriter, r *http.Request, p
 
 	p.Challenge = challenge
 	p.RequestedAt = cr.RequestedAt
+	p.HandledAt = time.Now()
```
aeneasr:
```suggestion
	p.HandledAt = time.Now().UTC()
```

consent/handler.go
```diff
@@ -553,6 +553,7 @@ func (h *Handler) AcceptConsentRequest(w http.ResponseWriter, r *http.Request, p
 
 	p.Challenge = challenge
 	p.RequestedAt = cr.RequestedAt
+	p.HandledAt = time.Now().UTC()
```
aeneasr: What about rejection?

aeneasr:
> Do you prefer to extend handler_test.go (https://github.com/ory/hydra/blob/master/consent/handler_test.go) or adding more cypress test (https://github.com/ory/hydra/blob/master/cypress/integration/oauth2/consent.js) for the new field? At the moment the accepting of consent is not tested anywhere, right?

In three parts:

* Since you've updated the DBAL, please add a test in the manager: https://github.com/ory/hydra/blob/master/consent/manager_test_helpers.go
* Also, as you touched the handler, please add a handler test. It would first "accept" and/or "reject" the consent request using the HTTP API, and would then test if the storage has the appropriate time set. You can use `time.Round()` to round to the day so that you don't have clock synchro issues.
* I would love to see some e2e tests. It would need a new test suite however, because we don't have tests for that (checking past consents) at the moment.
* However, another test area is here: https://github.com/ory/hydra/blob/master/oauth2/oauth2_auth_code_test.go

aeneasr:
Thank you for the updates, I've added this to the [project board](https://github.com/orgs/ory/projects/8) and will review it as soon as I can

cypress/integration/oauth2/consent.js
```diff
@@ -36,6 +36,34 @@ describe('OAuth 2.0 End-User Authorization', () => {
       .then(body => {
         expect(body.length).to.be.greaterThan(0);
         expect(hasConsent(client, body)).to.be.true;
+        body.forEach(consent => {
```
aeneasr: `handled_at` should probably already been set - even without this code, right? `sessions/consent` should only show "handled" consent requests - which implies that they need the `handled_at` to be set.

DennisPattmann5012: You're right. We have updated the test.

DennisPattmann5012:
Are you sure to enforce only standard semantic pull requests? Do you plan to add a `.github/semantic.yaml` with types fitting to hydra? If you want to we can add the prefix `feat:` to the PR.

aeneasr:
Yes, it will help with auto-generation of changelogs as that's still something that's really annoying and manual work (but could be automated). I don't think we'll be using schemantic release.

Don't worry about the commit messages, I'll just squash merge them with appropriate wording :)

aeneasr:
Awesome, thank you for your hard work!

aeneasr:
> Are you sure to enforce only standard semantic pull requests? Do you plan to add a .github/semantic.yaml with types fitting to hydra? If you want to we can add the prefix feat: to the PR.

@DennisPattmann5012 do you have negative experience with semantic commits? To me it made sense and I'm also kind of inconsistently tagging commits. Enforcing scopes e.g. `feat(sql|consent|...)` will hopefully make this easier.

I'll definitely add `semantic.yml` with a scope soon, the default `schematic` has things like `feat` `docs` `ci` etc

PR #1733, opened 2020-02-18
feat: add a config.schema.json and validate the configuration against it

### LINKED ISSUE #1729: Validate configuration using JSON Schema and support config hot reloading

Currently, ORY Hydra validates the configuration in a very crude way. We should fix that and support JSON Schema as we already do in Oathkeeper and Kratos.

To see how kratos does that (validation and reloading), check here: https://github.com/ory/kratos/blob/master/cmd/serve.go#L53

Hydra already implements the architecture of kratos/oathkeeper that supports hot reloading of config values. Some config values (such as the server bind address) can not be reloaded and should thus cause an error (see how kratos/oathkeeper is handling that). We need to create a list of variables that can not be reloaded. I believe that the fosite's config doesn't support reloading at this time because many config values are read only once: https://github.com/ory/hydra/blob/74d8e3174c0adb33a263cb8d935d9c6c653fa81e/driver/registry_base.go#L247

zepatrik:
## Related issue
closes #1729 

## Checklist

- [ ] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md)
- [ ] I have read the [security policy](../security/policy)
- [ ] I confirm that this pull request does not address a security vulnerability. If this pull request addresses a security


docs/config.schema.json
```diff
@@ -0,0 +1,816 @@
...
+      }
+    },
+    "ttl": {
+      "type": "object",
+      "description": "Configures time to live.",
+      "properties": {
+        "login_consent_request": {
+          "description": "Configures how long a user login and consent flow may take.",
+          "default": "1h",
+          "allOf": [
```
aeneasr: Again, not sure if we need allOf?

docs/config.schema.json
```diff
@@ -0,0 +1,816 @@
...
+          "properties": {
+            "bcrypt": {
+              "type": "object",
+              "description": "Configures the BCrypt hashing algorithm used for hashing Client Secrets.",
+              "properties": {
+                "cost": {
+                  "type": "integer",
+                  "description": "Sets the BCrypt cost. The higher the value, the more CPU time is being used to generate hashes.",
+                  "default": 10,
+                  "minimum": 4
```
aeneasr: Minimum is 8

docs/config.schema.json
```diff
@@ -0,0 +1,816 @@
...
+        }
+      }
+    },
+    "profiling": {
+      "type": "string",
+      "description": "Enables profiling if set. For more details on profiling, head over to: https://blog.golang.org/profiling-go-programs",
+      "enum": [
+        "cpu",
+        "mem",
+        ""
```
aeneasr: I think this can be removed as empty is just leaving it out of the definition right?

docs/config.schema.json
```diff
@@ -0,0 +1,816 @@
...
+      "properties": {
+        "enabled": {
+          "type": "boolean",
+          "description": "Sets whether CORS is enabled.",
+          "default": false
+        },
+        "allowed_origins": {
+          "type": "array",
+          "description": "A list of origins a cross-domain request can be executed from. If the special * value is present in the list, all origins will be allowed. An origin may contain a wildcard (*) to replace 0 or more characters (i.e.: http://*.domain.com). Only one wildcard can be used per origin.",
+          "items": {
```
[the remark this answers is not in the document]
zepatrik: examples of an array should be
``` json
{
  "examples": [[
    "https://example.com",
    "https://*.example.com",
    "https://*.foo.example.com"
  ]]
}
```
right?

docs/config.schema.json
```diff
@@ -0,0 +1,816 @@
...
+      "type": "object",
+      "description": "Controls the configuration for the http(s) daemon(s).",
+      "properties": {
+        "public": {
+          "type": "object",
+          "description": "Controls the public daemon serving public API endpoints like /oauth2/auth, /oauth2/token, /.well-known/jwks.json",
+          "properties": {
+            "port": {
+              "default": 4444,
+              "allOf": [
```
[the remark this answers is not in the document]
zepatrik: there are no keys allowed next to `$ref` https://stackoverflow.com/questions/33565090/json-schema-property-description-and-ref-usage#answer-33663032

docs/config.schema.json
```diff
@@ -0,0 +1,816 @@
...
+          "format": "uri",
+          "examples": [
+            "https://my-example.app/logout-successful"
+          ]
+        }
+      }
+    },
+    "strategies": {
+      "type": "object",
+      "description": "TODO",
```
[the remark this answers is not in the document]
zepatrik: haha yes felt like at this point a description is needed

docs/config.schema.json
zepatrik: sure? found this in `config.yaml`
> Minimum value is 4 and default value is 10.

docs/config.schema.json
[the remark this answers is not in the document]
aeneasr: yeah

docs/config.schema.json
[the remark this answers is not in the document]
aeneasr: ok, then forget the similar comments :)

zepatrik:
As I found out it is not possible with jsonschema to have `nil` as a default value. It's the same problem as with `"const": null` but there it is solved by [wrapping the value with a slice](https://github.com/ory/jsonschema/blob/master/schema.go#L30). So we will have to set some defaults in `config.schema.json` that are currently set in the code (e.g. [like this one](https://github.com/ory/hydra/pull/1733/commits/b86f99f485ff0967b23be402336a7a5d015bbc1d#diff-944f3a9b461996536a28cb631807fe5bR661)). `viperx` will automatically set the default viper value to either the [default value from the schema or the default value for the type](https://github.com/ory/x/blob/master/viperx/bind_env.go#L42:L48). This means an array type with `"minItems"` that is not set will not pass validation. @aeneasr

aeneasr:
That is not a good idea, especially with secrets. Let’s nit do that please and figure out another way to tackle this. If no secret is set, then it should not pass validation, right? If it should pass, then minItems should not be 1

docs/config.schema.json
```diff
@@ -656,28 +656,16 @@
...
-            [
-              "this-is-the-primary-secret",
-              "this-is-an-old-secret",
-              "this-is-another-old-secret"
-            ]
-          ]
-        },
-        "cookie": {
+          "default": [
+            "I AM NOT SECRET PLEASE REPLACE ME!!!"
```
aeneasr: Please remove this immediately. This will cause any malformed config or inexperienced user to have this as their secret which is really bad. Never set hardcoded default secrets ever anywhere!

docs/config.schema.json
```diff
@@ -656,28 +656,16 @@
...
+          ],
           "type": "array",
-          "description": "A secret that is used to encrypt cookie sessions. Defaults to secrets.system. It is recommended to use\ta separate secret in production. The first item in the list is used for signing and encryption. The whole list is used for verifying signatures and decryption.",
           "items": {
             "type": "string",
             "minLength": 16
           },
+          "minItems": 1,
```
aeneasr: I think hydra auto-generates a secret when none is set. This has caused some confusion where people forgot to set the system secret, restarted the service, and then wondered why the db connectivity did not longer work.

I am actually not sure if we still generste the secret per default if none is set or not (but I think we do). In that (we do) just set minItems to 0 and say in the description that it is generated per default - or change (and document) the behavior and require a system secret to be set.

zepatrik: yes it is generated and there is also log output

aeneasr: Ok, in that case `minItems` should be removed and the default as well

aeneasr:
@zepatrik what's the status of this? Will you continue working on it?

zepatrik:
latest schema changes needed as workaround for https://github.com/ory/x/issues/131

cmd/server/handler.go
```diff
@@ -75,10 +85,67 @@ func isDSNAllowed(d driver.Driver) {
 	}
 }
 
+var schemas = packr.New("schemas", "../../docs")
+
+// LoggerWithValidationErrorFields adds all validation errors as fields to the logger.
+func LoggerWithValidationErrorFields(l logrus.FieldLogger, err error) logrus.FieldLogger {
```
aeneasr: Is this not provided by the X package?

aeneasr: ping @zepatrik

zepatrik: Done

Issue #1719, opened 2020-02-06
Duplicate key value violates unique constraint


mpanne:
**Describe the bug**

For our project we are using Hydra and a PostgreSQL database. Sometimes when trying to get an access token calling the /oauth2/token endpoint, Hydra returns an 500 Internal Server Error and logs the following error message: 

`level=error msg="An error occurred" debug="pq: duplicate key value violates unique constraint \"hydra_oauth2_access_request_id_idx\": Unable to insert or update resource because a resource with that value exists already" description="The authorization server encountered an unexpected condition that prevented it from fulfilling the request" error=server_error`

Hydra seems to try to add an access token request to the database with a request id which already exists. What we could observe is that every time this happened multiple (2-3) token requests have been received in a very short time period (<100ms). 

Did you experience this at some point or do you have any hint what could be the root cause for it?
We are using an older version (v1.0.0) but I couldn't find anything regarding this in the changelog.

**Reproducing the bug**

We could not reproduce the bug not even when sending ~500 token requests (almost) simultaneously.  

**Environment**

* Version: v1.0.0
* Environment: Docker, PostgreSQL DB

aeneasr:
Never seen this before, could you set `LOG_LEVEL=trace` and print out the trace? That would help a lot! Please also include logs surrounding the error (so we can see what requests are happening etc).

It would also be interesting to see if this happens with the latest version.

mpanne:
We will set the log level as requested and provide additional logs as soon as it happens again, thank you! 
We are also about to update to a newer Hydra version soon.

mpanne:
Is log level `trace` even supported? As stated [here](https://github.com/ory/hydra/blob/master/docs/config.yaml#L78), only `debug` is supported. And even when setting the log level to `debug`, we cannot observe any debug logs.

aeneasr:
Yeah, that's a docs issue. `trace` should work but you can also use `debug` if you want. The error messages, if some occur, should be enhanced with stack traces.

mpanne:
It happened again today. Following the logs enhanced with debug log. The `request_id` is randomly generated and added to the request headers by our nginx ingress controller.

```
time="2020-02-10T09:42:18Z" level=info msg="started handling request" method=POST remote=172.20.6.158 request=/oauth2/token request_id=16ebd03dde4752f4d6b551c05e885e06
[...]

PR #1734, opened 2020-02-19
test: add test to reproduce https://github.com/ory/hydra/issues/1719

aaslamin:
### Related issue

https://github.com/ory/hydra/issues/1719

### What's going on here...

This PR adds a test that consistently reproduces the bug identified in the linked issue. To run locally, navigate to the `oauth2` directory and run:

```
go test -run=TestCreateRefreshTokenSessionBug -count=1 -v
```

⚠️ _Note: this is an integration test and requires you to have Docker running locally as it will bring up a Postgres container_

#### 

### Checklist

### DIFF
--- a/oauth2/oauth2_refresh_token_test.go
+++ b/oauth2/oauth2_refresh_token_test.go
@@ -1,0 +1,128 @@
+package oauth2_test
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/ory/fosite"
+	hc "github.com/ory/hydra/client"
+	"github.com/ory/hydra/internal"
+	"github.com/ory/hydra/oauth2"
+	"github.com/ory/hydra/x"
+	"github.com/pborman/uuid"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestCreateRefreshTokenSessionBug reproduces the bug raised in https://github.com/ory/hydra/issues/1719
+// Once this bug is fixed, this test should start failing. It currently only deals with Postgres as that was
+// what the issue was based on due to the default isolation level used by the storage engine.
+func TestCreateRefreshTokenSessionBug(t *testing.T) {
+	if testing.Short() {
+		return
+	}
+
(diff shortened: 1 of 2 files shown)


oauth2/oauth2_refresh_token_test.go
```diff
@@ -0,0 +1,128 @@
...
+					return
+				default:
+					switch err := errors.Cause(err).(type) {
+					case *fosite.RFC6749Error:
+
+						if strings.Contains(err.Debug, "pq: duplicate key value violates unique constraint") {
+							stopWorkers()
+							successCh <- struct{}{}
+							return
+						}
```
aaslamin: This is where you want to end up in! _(to reproduce the 🐛)_ 

image

oauth2/oauth2_refresh_token_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+						if strings.Contains(err.Debug, "not_found") {
+							// too late, a goroutine finished the transaction before any other goroutine got passed the first read
+							// at this point, the test will be unfruitful so let's try to add another refresh session so they can
+							// race again!
+							postgresRegistry.OAuth2Storage().CreateRefreshTokenSession(ctx, tokenSignature, request)
+							stopWorkers()
+						}
```
aaslamin: 😭 try again!

image

aeneasr:
Yo, this is a sweet test case! If you'd be up for fixing the issue as well I think this can be merged right away!

aaslamin:
> Yo, this is a sweet test case! If you'd be up for fixing the issue as well I think this can be merged right away!

@aeneasr I'll take a look at this later in the week when I get a free cycle. However, I think the solution lies in `fosite` as the errors in the refresh flow are encapsulated as a [server error](https://github.com/ory/fosite/blob/master/errors.go#L74) which is not what we want.

We may need to handle this on a DB by DB basis as the error copy returned by Postgres, MySQL and CockroachDB could be different when an outdated read is done in a transaction utilizing the repeatable read iso level. Alternatively, we could return the raw error from fosite and let the Hydra deal with it via the sqlx package.

aeneasr:
You're the best man!

I think we can generally enforce one transaction type for all databases, and release a beta to see if it has negative performance impacts for people. What do you think?

aaslamin:
Closing in favour of https://github.com/ory/hydra/pull/1766 which has a better implementation.

PR #1736, opened 2020-02-20
refactor: switch from lib/pq to jackc/pgx

### LINKED ISSUE #1599: Hydra write to database: broken pipe
**Describe the bug**

Requests to hydra (running with at least postgres) can sometimes hit an "idle" (but closed) connection to the DB. In this scenario the caller gets back:

```
error_hint=write tcp SRC:37738->DST:5432: write: broken pipe
```

I've run into similar errors in other apps, and the issue is upstream in lib/pq as well (https://github.com/lib/pq/issues/870).

I was able to work around this issue by setting `max_conn_lifetime=10s` in the DSN -- as it avoids the connections being held open but actually closed on the remote.

**Reproducing the bug**

This has been annoying to reproduce on-demand. I've noticed this when tokens go to get refreshed from the IDP -- specifically after a long time (8hrs in this case) such that the connection can close out.

**Expected behavior**

I would expect that L4 errors such as this would be transparently retried instead of bubbling all the way to the client.

glerchundi:
Closes #1599


glerchundi:
Do not merge this one until ory/x#118 is merged so that I could remove the `replace` thing from `go.mod`.

glerchundi:
I think this is failing due to gorp not identifying `pgx` driver. Perhaps it would be interesting to keep the same driver by wrapping it? `WrapDriver("postgres", "pgx")`.

WDYT?

glerchundi:
@aeneasr PTAL.

aeneasr:
Thank you! Released as a new version

glerchundi:
Do not forget to remove the replace in the go.mod:

replace github.com/ory/x => github.com/glerchundi/x
v0.0.97-0.20200221090043-7f3ebe7b46ec

PR #1739, opened 2020-02-23
refactor: switch from lib/pq to jackc/pgx (#1738)

### LINKED ISSUE #1599: Hydra write to database: broken pipe

aeneasr:
Closes #1599


go.mod
```diff
@@ -64,4 +69,6 @@ replace git.apache.org/thrift.git => github.com/apache/thrift v0.0.0-20180902110
 // Fix for https://github.com/golang/lint/issues/436
 replace github.com/golang/lint => github.com/golang/lint v0.0.0-20190227174305-8f45f776aaf1
 
+replace github.com/ory/x => github.com/glerchundi/x v0.0.97-0.20200221090043-7f3ebe7b46ec
```
glerchundi: Drop this line and update to use latest.

glerchundi:
LGTM.

go.mod
aeneasr: Thanks, I thought I did that but apparently not :D

Issue #1735, opened 2020-02-20
Token retrieval failed with error code 500


inigohu:
**Describe the bug**

Sometimes, I get error code 500 when clients makes a request to the token endpoint

**Reproducing the bug**

1. Make API Request to with `https://mydomain/oauth2/token`
2. Request fails with response: `level=error msg="An error occurred" debug=": not_found" description="The authorization server encountered an unexpected condition that prevented it from fulfilling the request" error=server_error`

*Server logs*

```
Stack trace: 
github.com/ory/fosite/handler/oauth2.(*RefreshTokenGrantHandler).PopulateTokenEndpointResponse
    /go/pkg/mod/github.com/ory/fosite@v0.30.0/handler/oauth2/flow_refresh.go:145
github.com/ory/fosite.(*Fosite).NewAccessResponse
    /go/pkg/mod/github.com/ory/fosite@v0.30.0/access_response_writer.go:36
github.com/ory/hydra/oauth2.(*Handler).TokenHandler
    /go/src/github.com/ory/hydra/oauth2/handler.go:585
net/http.HandlerFunc.ServeHTTP
    /usr/local/go/src/net/http/server.go:2007
github.com/julienschmidt/httprouter.(*Router).Handler.func1
    /go/pkg/mod/github.com/julienschmidt/httprouter@v1.2.0/params_go17.go:26
github.com/julienschmidt/httprouter.(*Router).ServeHTTP
    /go/pkg/mod/github.com/julienschmidt/httprouter@v1.2.0/router.go:334
github.com/urfave/negroni.Wrap.func1
    /go/pkg/mod/github.com/urfave/negroni@v1.0.0/negroni.go:46
github.com/urfave/negroni.HandlerFunc.ServeHTTP
    /go/pkg/mod/github.com/urfave/negroni@v1.0.0/negroni.go:29
github.com/urfave/negroni.middleware.ServeHTTP
    /go/pkg/mod/github.com/urfave/negroni@v1.0.0/negroni.go:38
net/http.HandlerFunc.ServeHTTP
    /usr/local/go/src/net/http/server.go:2007
github.com/ory/hydra/x.RejectInsecureRequests.func1
    /go/src/github.com/ory/hydra/x/tls_termination.go:55
github.com/urfave/negroni.HandlerFunc.ServeHTTP
    /go/pkg/mod/github.com/urfave/negroni@v1.0.0/negroni.go:29
github.com/urfave/negroni.middleware.ServeHTTP
[...]

PR #1766, opened 2020-03-22
fix: force transaction isolation level to `LevelRepeatableRead`

aaslamin:
## Related issues

- https://github.com/ory/hydra/issues/1719
- https://github.com/ory/hydra/issues/1735

## ⚠️ Dependent PRs

- [x] https://github.com/ory/fosite/pull/402
- [x] https://github.com/ory/x/pull/129

_(Each of the packages above will need a new tag release once/if the PRs have been merged)_ 🙏 

## Proposed changes

To improve consistency in certain authorization flows that utilize transactions, this PR forces the SQL storage transaction isolation level to `LevelRepeatableRead`. This will ensure that we avoid the phenomena of non-repeatable reads which occur when a transaction re-reads data it has previously read and then finds out that another transaction has since modified that data and committed. As a result, setting this isolation level fixes a flaw where one could use a given refresh token more than once. See the test added.

In the event that multiple concurrent transactions are competing under a given refresh token workflow, the underlying database engine will eventually return an error when one of the transactions successfully commits. For example, in such a scenario, postgres will rollback the transaction with:

```
could not serialize access due to concurrent update (SQLSTATE 40001)
```

## TODO 

- [x] This PR needs to accompany changes in `fosite` and likely `x` for error handling logic in the event the underlying database engine returns an error due to concurrent transaction activity on the same rows
- The error types and codes returned under this condition will likely be different for each database driver implementation

I am thinking the best way to do this is to add a new error type to represent a "transaction (serialization) error" which can be caught & returned by the [`HandleError`](https://github.com/ory/x/blob/master/sqlcon/error.go#L33) API exposed by the `sqlcon` package. This API is already consumed by relevant methods of Hydra's SQL based implementation of the `fosite` storage interfaces.

Finally, inside `fosite`'s [refresh token grant handler](https://github.com/ory/fosite/blob/master/handler/oauth2/flow_refresh.go#L118) we could check for this new transaction error type and return an appropriate `RFC6749Error` error. I am thinking this particular error should result in a 503 based on the [postgres documentation](https://www.postgresql.org/docs/9.1/transaction-iso.html#MVCC-ISOLEVEL-TABLE
):

> When an application receives this error message, it should abort the current transaction and retry the whole transaction from the beginning.

In our case, we will not be retrying the transaction because it will be unfruitful. The next time the user tries to call the token endpoint with the same refresh token, it will result in an invalid request as expected. It seems strange to deduce an error as a result of transaction concurrency with a concrete 4xx error, no?

```
To summarize: 

(multiple txn's competing to use the same refresh token)

• fosite refresh handler calls out to storage after initing txn earlier w/ new iso level
• storage calls out to actual database to modify some row
• db returns an error due to concurrent access, eg:
   • `could not serialize access due to concurrent update (SQLSTATE 40001)`
• we use the `sqlcon` package to capture this error and wrap it in some generic "txn serialization error" type and return it up the callstack
• eventually back into fosite code which can check for this error and create a RFC6749Error
```

@aeneasr what do you think of this approach?

- [x] Expand test coverage for all SQL backed stores, currently I am only dealing with postgres

- [x] Investigate how `MySQL` & `CockroachDB` behave under this isolation level (_the test I added should hopefully make this straightforward 🤞_)

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,7 +26,7 @@ require (
 	github.com/oleiade/reflections v1.0.0
 	github.com/olekukonko/tablewriter v0.0.1
 	github.com/opentracing/opentracing-go v1.1.1-0.20190913142402-a7454ce5950e
-	github.com/ory/fosite v0.30.5
+	github.com/ory/fosite v0.30.6
 	github.com/ory/go-acc v0.2.1
 	github.com/ory/graceful v0.1.1
 	github.com/ory/herodot v0.7.0
--- a/oauth2/fosite_store_sql.go
+++ b/oauth2/fosite_store_sql.go
@@ -31,6 +31,7 @@ import (
 	"time"
 
 	"github.com/jmoiron/sqlx"
+	"github.com/ory/herodot"
 	"github.com/pkg/errors"
 	migrate "github.com/rubenv/sql-migrate"
 	"github.com/sirupsen/logrus"
@@ -289,7 +290,14 @@ func (s *FositeSQLStore) createSession(ctx context.Context, signature string, re
 		":"+strings.Join(sqlParams, ", :"),
 	)
 	if _, err := db.NamedExecContext(ctx, query, data); err != nil {
-		return sqlcon.HandleError(err)
+		err = sqlcon.HandleError(err)
+		switch errCause := errors.Cause(err).(type) {
+		case *herodot.DefaultError:
+			if errCause == sqlcon.ErrConcurrentUpdate {
+				return errors.Wrap(fosite.ErrSerializationFailure, err.Error())
+			}
+		}
+		return err
 	}
 	return nil
 }
@@ -430,8 +438,14 @@ func (s *FositeSQLStore) revokeSession(ctx context.Context, id string, table tab
 	/* #nosec G201 - table is a const enum */
 	if _, err := db.ExecContext(ctx, db.Rebind(fmt.Sprintf("DELETE FROM hydra_oauth2_%s WHERE request_id=?", table)), id); err == sql.ErrNoRows {
 		return errors.Wrap(fosite.ErrNotFound, "")
-	} else if err != nil {
-		return sqlcon.HandleError(err)
+	} else if err := sqlcon.HandleError(err); err != nil {
+		switch errCause := errors.Cause(err).(type) {
+		case *herodot.DefaultError:
+			if errCause == sqlcon.ErrConcurrentUpdate {
+				return errors.Wrap(fosite.ErrSerializationFailure, err.Error())
+			}
+		}
+		return err
 	}
 	return nil
 }
--- a/oauth2/fosite_store_sql.go
+++ b/oauth2/fosite_store_sql.go
@@ -447,7 +447,10 @@ func (s *FositeSQLStore) FlushInactiveAccessTokens(ctx context.Context, notAfter
 }
 
 func (s *FositeSQLStore) BeginTX(ctx context.Context) (context.Context, error) {
-	if tx, err := s.DB.BeginTxx(ctx, nil); err != nil {
+	if tx, err := s.DB.BeginTxx(ctx, &sql.TxOptions{
+		Isolation: sql.LevelRepeatableRead,
+		ReadOnly:  false,
+	}); err != nil {
 		return ctx, err
 	} else {
 		return context.WithValue(ctx, txKey, tx), nil
--- a/oauth2/fosite_store_sql.go
+++ b/oauth2/fosite_store_sql.go
@@ -475,7 +475,16 @@ func (s *FositeSQLStore) Commit(ctx context.Context) error {
 	if tx, ok := ctx.Value(txKey).(*sqlx.Tx); !ok {
 		return errors.Wrap(fosite.ErrServerError, "commit failed: no transaction stored in context")
 	} else {
-		return tx.Commit()
+		if err := sqlcon.HandleError(tx.Commit()); err != nil {
+			switch errCause := errors.Cause(err).(type) {
+			case *herodot.DefaultError:
+				if errCause == sqlcon.ErrConcurrentUpdate {
+					return errors.Wrap(fosite.ErrSerializationFailure, err.Error())
+				}
+			}
+			return err
+		}
+		return nil
 	}
 }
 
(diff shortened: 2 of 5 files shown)


oauth2/fosite_store_sql.go
```diff
@@ -447,7 +447,18 @@ func (s *FositeSQLStore) FlushInactiveAccessTokens(ctx context.Context, notAfter
...
+	// TL;DR make it a no-op if the transaction has already been initiated:
+	// some 'fosite' APIs may attempt to initiate a transaction and they may not be aware that
+	// one has already been initiated by the caller so we make this a no-op in those cases where
+	// the caller (hydra) has more knowledge of the transaction boundary.
+	if _, ok := ctx.Value(txKey).(*sqlx.Tx); ok {
+		return ctx, nil
+	}
+
+	if tx, err := s.DB.BeginTxx(ctx, &sql.TxOptions{
+		Isolation: sql.LevelRepeatableRead,
```
aaslamin: If you change this to `sql.LevelReadCommitted` and run the test scenario added, you will be able to successfully use the same refresh token [more than once](https://github.com/ory/hydra/pull/1766/files#diff-8a4c55f53fc1d3a445cf9930b67cfafaR149) 😨

oauth2/oauth2_refresh_token_test.go
```diff
@@ -0,0 +1,160 @@
...
+				switch err := errors.Cause(err).(type) {
+				case *fosite.RFC6749Error:
+
+					// TODO: ok this is the tricky part, we need to add error handling logic somewhere such that we are
+					// able to catch the following transaction error(s):
+					//
+					//   • ERROR: could not serialize access due to concurrent update (SQLSTATE 40001)
+					//
+					// this logic likely belongs somewhere in fosite as the refresh flow is currently wrapping any error
+					// it gets back into a `fosite.ErrServerError`.
```
aaslamin: FYI @aeneasr

oauth2/fosite_store_sql.go
aeneasr: Nice find man!

oauth2/oauth2_refresh_token_test.go
aeneasr: If it's not possible yet we could definitely allow `fosite.RFC6749Error` to implement `Cause() error` which would make it possible to unwrap the original error!

oauth2/oauth2_refresh_token_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+		// process worker results
+
+		// successCount is the number of workers that were able to call "NewAccessResponse" without receiving an error.
+		// if the successCount at the end of a test run is bigger than one, it means that multiple access/refresh tokens
+		// were issued using the same refresh token! - https://knowyourmeme.com/memes/scared-hamster
+		var successCount int
+		for err := range errorsCh {
+			if err != nil {
+				switch err := errors.Cause(err).(type) {
```
aeneasr: There's actually a bug in `errors.Cause` which can cause a panic if `Cause()` returns `nil`. This hasn't been fixed in upstream but you can use `github.com/ory/x/errorx.Cause()` as a workaround. Maybe this also has been fixed already, not sure.

aaslamin: Done ✅

go.mod
```diff
@@ -54,11 +54,16 @@ require (
...
 	golang.org/x/tools v0.0.0-20200313205530-4303120df7d8
-	gopkg.in/ini.v1 v1.54.0 // indirect
+	gopkg.in/gorp.v1 v1.7.2 // indirect
 	gopkg.in/square/go-jose.v2 v2.4.1
 )
+
+replace (
+	github.com/ory/fosite => github.com/aaslamin/fosite v0.21.6-0.20200324061735-29317aed4141
+	github.com/ory/x => github.com/aaslamin/x v0.0.8-0.20200324063550-0e5572e8199d
+)
```
aaslamin: ℹ️FYI @aeneasr this is just temporary to get CI to build until:

- https://github.com/ory/x/pull/129
- https://github.com/ory/fosite/pull/402

have been merged 🚀

oauth2/fosite_store_sql.go
```diff
@@ -289,7 +290,14 @@ func (s *FositeSQLStore) createSession(ctx context.Context, signature string, re
...
 	if _, err := db.NamedExecContext(ctx, query, data); err != nil {
-		return sqlcon.HandleError(err)
+		err = sqlcon.HandleError(err)
+		switch errCause := errors.Cause(err).(type) {
+		case *herodot.DefaultError:
+			if errCause == sqlcon.ErrConcurrentUpdate {
+				return errors.Wrap(fosite.ErrSerializationFailure, err.Error())
+			}
+		}
+		return err
```
aaslamin: The reasoning here is to avoid `fosite` having a dependency on `sqlcon`. So the error is mapped from `sqlcon.ErrConcurrentUpdate` to `fosite.ErrSerializationFailure` from which `fosite` will be able to handle and return an appropriate RFC 6749 error.

oauth2/oauth2_refresh_token_test.go
```diff
@@ -0,0 +1,158 @@
...
+			"were able to use the refresh token. Exactly ONE was expected to be have been successful.",
+			run,
+			successCount,
+			workers)
+
+		// reset state for the next test iteration
+		_ = postgresRegistry.OAuth2Storage().RevokeRefreshToken(ctx, request.ID)
+		_ = postgresRegistry.OAuth2Storage().CreateRefreshTokenSession(ctx, tokenSignature, request)
+	}
+}
```
aaslamin: @aeneasr if you have a spare cycle, it might be good to get you to try running this test locally so you get a feel for the value it provides. 

🎩 instructions:

- `git fetch`
- `git checkout aaslamin:fix/1735-and-1719`
- `cd <to hydra dir>`
- `cd oauth2`
- start docker
- `go test -race -run=TestCreateRefreshTokenSessionStress -count=1 -v`

Now, just to see what happens when something breaks:

Go to https://github.com/ory/hydra/pull/1766/files#diff-3d061c6808115be018958df459c564e2R465 and change the consistency level to `sql.LevelReadCommitted` and re-run the test. Can we make the failure experience better for the developer debugging the test?

aeneasr:
Looks like we have some failing test cases - could you please take a look? :)

PR #1771, opened 2020-03-25
chore: bump github.com/ory/fosite to v0.30.5

vancity-amir:
🙏 @aeneasr


aeneasr:
We can fix the linting in another pr

Issue #1755, opened 2020-03-12
Large latencies in sql-conn-query


inigohu:
**Describe the bug**

I have some `sql-conn-query` spans that have large latencies when running this query:
`SELECT h.* FROM hydra_oauth2_consent_request_handled as h JOIN hydra_oauth2_consent_request as r ON (h.challenge = r.challenge) WHERE r.subject=$1 AND r.client_id=$2 AND r.skip=FALSE AND (h.error='{}' AND h.remember=TRUE) ORDER BY h.requested_at DES`

Some of them take up to 46798.673 ms, 7196.002 ms, or 4821.986 ms for example.

**Reproducing the bug**

Make API Request to with `curl https://mydomain/oauth2/auth`

*Server configuration*

```yaml
    LOG_LEVEL: "debug",
    SERVE_PUBLIC_PORT: "8080",
    SERVE_ADMIN_PORT: "9000",
    URLS_SELF_ISSUER: accountUrl,
    URLS_LOGIN: signInUrl + "/login",
    URLS_LOGOUT: signInUrl + "/logout",
    URLS_CONSENT: signInUrl + "/consent",
    URLS_ERROR: signInUrl + "/error",
    URLS_POST_LOGOUT_REDIRECT: appUrl,
    OIDC_SUBJECT_IDENTIFIERS_ENABLED: "public,pairwise",
    TTL_LOGIN_CONSENT_REQUEST: "1h",
    TTL_ACCESS_TOKEN: "1h",
    TTL_REFRESH_TOKEN: "2160h",  // 90 days in hours
    TTL_ID_TOKEN: "1h",
    TTL_AUTH_TOKEN: "10m",
```

**Expected behavior**

Lower latency

**Environment**

* Version: v1.0.9
* Environment: Docker in kubernetes

aeneasr:
I believe this has long been fixed, can you still replicate this with Hydra 1.3.2?

inigohu:
> I believe this has long been fixed, can you still replicate this with Hydra 1.3.2?

Can you point me to the change? I would like to check it with my team

aeneasr:
It’s quite an effort to find it but I do know that we added indices for those rows after 1.0 plus your version is pretty old and you should upgrade anyways.

rickwang7712:
My commit [Add several SQL lookup indices (#1654) ](https://github.com/ory/hydra/pull/1654)is included in v1.1.0.
Maybe @inigohu can try versions later or equal to v1.1.0.

[...]

PR #1790, opened 2020-04-08
fix: consent cockroachdb perfomance issue with zigzag join query 

### LINKED ISSUE #1789: cockroachdb: consent query slow
**Describe the bug**

Related issues:

https://github.com/ory/hydra/issues/1755
https://github.com/cockroachdb/cockroach/issues/47179

Apparently there is an intermittent (sometimes not reproducible) issue when zigzag join queries hit multiple regions.

More info about zigzag joins here:

https://github.com/cockroachdb/cockroach/issues/23520#issue-302972941

As CockroachDB moves the leaseholder location and learns from usage, this is not reproducible always, but can be performing slow for a while, sometimes days.

**Additional context**

Underperformant query (sometimes we saw times as high as 50s):

```sql
SELECT h.* 
FROM hydra_oauth2_consent_request_handled as h 
JOIN hydra_oauth2_consent_request as r ON (h.challenge = r.challenge) 
WHERE r.subject='my@subject.com' AND r.client_id='my-client' AND r.skip=FALSE 
AND (h.error='{}' AND h.remember=TRUE) 
ORDER BY h.requested_at DESC;
```

Its query plan (EXPLAIN <query>):

```
	distributed	true
	vectorized	false
sort		
 │	order	-requested_at
 └── render		
      └── lookup-join		
           │	table	hydra_oauth2_consent_request_handled@primary
           │	type	inner
           │	equality	(challenge) = (challenge)
           │	equality cols are key	
           │	parallel	
           │	pred	"(@9 = '{}') AND (@7 = true)"
           └── lookup-join		
                │	table	hydra_oauth2_consent_request@primary
                │	type	inner
                │	equality	(challenge) = (challenge)
                │	equality cols are key	
                │	parallel	
                │	pred	@4 = false
                └── zigzag-join		
                     │	type	inner
                     │	pred	"(@3 = 'my@subject.com') AND (@2 = 'my-client')"
                     ├── scan		
                     │	table	hydra_oauth2_consent_request@hydra_oauth2_consent_request_client_id_idx
                     │	fixedvals	1 column
                     └── scan		
	table	hydra_oauth2_consent_request@hydra_oauth2_consent_request_subject_idx
	fixedvals	1 column
```

You can see the culprit zigzag join here.

lopezator:
Add an index over subject and client_id in order to avoid the
(sometimes) underperformant zigzag join query.

Closes #1789
Related to #1755 https://github.com/cockroachdb/cockroach/issues/47179

After applying the proposed index, the query plan now avoids the zigzag join query, reducing the query times from 50s to 2s in worst cases in my tests.

```
	distributed	true
	vectorized	false
sort		
 │	order	-requested_at
 └── render		
      └── lookup-join		
           │	table	hydra_oauth2_consent_request_handled@primary
           │	type	inner
           │	equality	(challenge) = (challenge)
           │	equality cols are key	
           │	parallel	
           │	pred	"(@9 = '{}') AND (@7 = true)"
           └── filter		
                │	filter	skip = false
                └── index-join		
                     │	table	hydra_oauth2_consent_request@primary
                     └── scan		
	table	hydra_oauth2_consent_request@hydra_oauth2_consent_request_client_id_subject_idx
	spans	"/""myclient""/""my@subject.com""-/""my-client""/""my@subject.com""/PrefixEnd"
```

### DIFF
--- a/persistence/sql/migrations/20200521071434_consent.cockroach.down.sql
+++ b/persistence/sql/migrations/20200521071434_consent.cockroach.down.sql
@@ -1,0 +1,1 @@
+DROP INDEX hydra_oauth2_consent_request@hydra_oauth2_consent_request_client_id_subject_idx;
--- a/persistence/sql/migrations/20200521071434_consent.cockroach.up.sql
+++ b/persistence/sql/migrations/20200521071434_consent.cockroach.up.sql
@@ -1,0 +1,1 @@
+CREATE INDEX hydra_oauth2_consent_request_client_id_subject_idx ON hydra_oauth2_consent_request (client_id, subject);
--- a/persistence/sql/migrations/20200521071434_consent.mysql.down.sql
+++ b/persistence/sql/migrations/20200521071434_consent.mysql.down.sql
@@ -1,0 +1,1 @@
+DROP INDEX hydra_oauth2_consent_request ON hydra_oauth2_consent_request_client_id_subject_idx;
--- a/persistence/sql/migrations/20200521071434_consent.mysql.up.sql
+++ b/persistence/sql/migrations/20200521071434_consent.mysql.up.sql
@@ -1,0 +1,1 @@
+CREATE INDEX hydra_oauth2_consent_request_client_id_subject_idx ON hydra_oauth2_consent_request (client_id, subject);
--- a/persistence/sql/migrations/20200521071434_consent.postgres.down.sql
+++ b/persistence/sql/migrations/20200521071434_consent.postgres.down.sql
@@ -1,0 +1,1 @@
+DROP INDEX hydra_oauth2_consent_request_client_id_subject_idx;
--- a/persistence/sql/migrations/20200521071434_consent.postgres.up.sql
+++ b/persistence/sql/migrations/20200521071434_consent.postgres.up.sql
@@ -1,0 +1,1 @@
+CREATE INDEX hydra_oauth2_consent_request_client_id_subject_idx ON hydra_oauth2_consent_request (client_id, subject);


aeneasr:
Thank you! The migrations still need to be improved but since @zepatrik is refactoring that at the moment I think it makes sense to include this in his PR / after his merge!

zepatrik:
Hey @lopezator 
The migration PR is finally done, can you adopt this PR to the new migration format? Basically you just have to split the migration in two `*.cockroach.up.sql` and `*.cockroach.down.sql` files.
Please also add empty migrations for the other databases, I'm pretty sure the migrator will be confused when the numbers don't match (files without the database name in it will be applied to all databases without specific files)

lopezator:
@zepatrik What would be exactly the file naming convention? Something related to date? I see that every migration file starts by `2019`

zepatrik:
You can either use [gobuffalo's soda cli](https://github.com/gobuffalo/pop) to generate the files or just copy an existing one and increase the number. It is usually of the format `YYYYMMDDHHMMSS_name(.database)?.(up|down).(sql|fizz)`

lopezator:
Ok! 

Do you prefer empty migrations for `postgres` and `mysql` or are you ok on adding the index for all the databases? 

I think that this index might not add as much value as it adds for cockroachdb, but as long as there are queries on that table using both fields in the where clause this could be beneficial for them also.

Let me know what you think.

zepatrik:
Think you can apply it to all or @aeneasr ?

aeneasr:
apply for all should be ok!

lopezator:
> apply for all should be ok!

Cool! 

PTAL @zepatrik

zepatrik:
right let me fix that I know what the problem is

zepatrik:
It's actually not a problem with the test.
@lopezator you have to apply your migration after all others, just put the current date/time there.
Or use `soda generate sql consent-request-index`.
If the migration for all databases is the same you should be able to omit the database name in the file completely and just have one up and one down.

zepatrik:
The question remains: why did the mysql test pass??

lopezator:
> The question remains: why did the mysql test pass??

Good question 😅 

PTAL @zepatrik

aeneasr:
Thank you both for the great work!

lopezator:
You are welcome @aeneasr !

PR #1816, opened 2020-04-22
fix(docker): Add nsswitch.conf into the dockerfiles

silveraid:
Go's netgo implementation currently does not respect hostname overrides
defined in /etc/hosts if the /etc/nsswitch.conf does not exists.

Made changes to the Dockerfiles to add a standard /etc/nsswitch.conf
to fix this issue.

## Related issue

No issue have been opened for this.

## Proposed changes

When hydra is being used in a form of a docker image in an environment where hostname overrides are being used through /etc/hosts the overrides are not respected. This is not an issue with hydra all, but a well known one with golang and alpine, as the alpine base image a few years ago got rid of the nsswitch.conf saying that it is not needed anymore. Unfortunately golang depends on the presence of this file and when it is missing it just ignores whatever is in the /etc/hosts file.

This change adds the nsswitch.conf back to the docker image of hydra. 

## Checklist

      the changes.
      works.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,10 @@ FROM scratch
 COPY --from=0 /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
 COPY hydra /usr/bin/hydra
 
+# set up nsswitch.conf for Go's "netgo" implementation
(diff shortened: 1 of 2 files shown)


aeneasr:
Is it this issue https://github.com/golang/go/issues/35305?

aeneasr:
Looks like it!

PR #1843, opened 2020-05-05
docs: Adding details about CSRF cookie problems

robhinds:
Issue I experienced today, running Hydra 1.4.10 in dangerous HTTP mode, the CSRF cookie defaulted to SameSite=None, but the cookie was not marked as secure (which makes sense, as Hydra is running over HTTP), so the cookie gets ignored (and was getting CSRF value not present errors).

I was able to get around it by either overriding the SameSite setting, or by switching to TLS termination.

## Proposed changes
Additional line in the docs about CSRF issues if SameSite=None and running over HTTP (the cookie won't be marked as Secure, and modern browsers such as Chrome will just ignore the cookie)

## Checklist

      the changes.
      works.


aeneasr:
Thank you!

PR #1889, opened 2020-05-30
feat: support jwt signing alg other than RS256

### LINKED ISSUE #1817: Hydra OIDC server does not support PS256 algorithm for client_assertion
I have used Hydra Server v1.4.5. I am running the Hydra server with default configuration. .hydra.yml is blank. When I invoke the token end point, i am using client_assertion_type and client_assertion. I have passed the client_assertion JWT with algorithm PS256. But the token end point giving the error stating the algorithm has to be RS256. Then I changed the JWT with RS256 algorithm and it is working fine.

After discussing in ory hydra community forum it appears that while the code supports PS256 and others, the client definition currently hardcodes the RS256 requirement. 

Please fix this issue.

thanks,
Karthik

ka3de:
Fixes #1817 .

## Proposed changes
Add new field `TokenEndpointAuthSigningAlgorithm` for Client struct which by default will be returned as `RS256` from the "getter" function `GetTokenEndpointAuthSigningAlgorithm` unless a defined value is set in client payload when performing a request to the `Create OAuth 2.0 client` endpoint. This of course implies a migration to be applied to the `hydra_cient` table to include this new field.

## Checklist

      the changes.
      works.

## Further comments
I have added no doc as I believe the new field will be auto generated  by CI as a consequence of this PR.

Regarding testing I though no new tests were needed because the logic has not changed and any new algorithm specified other than the previously default option of RS256 will not imply any change to hydra code, but will be delegated to fosite library, but if you believe otherwise let me know and I'll try to work on it.

I tested the change with JWT client assertions using PS256 and ES256 algorithms nd seems to be working fine (upgrading fosite library to v0.32.0 version).
I did not upgraded fosite library to v0.32.0 in this PR, let me know if I should take care of that here.

Happy to hear any feedback regarding any topic etc.
Thanks.


persistence/sql/migrations/20200527215731_client.up.sql
```diff
@@ -0,0 +1 @@
+ ALTER TABLE hydra_client ADD token_endpoint_auth_signing_alg VARCHAR(10) NOT NULL DEFAULT '';
```
aeneasr, line 1: Nice! This would also need a down migration, plus a test migration ([example](https://github.com/ory/hydra/blob/c92ab20dae186585f1cd3f9388791d3ae9d24ebb/persistence/sql/migratest/testdata/20190100000006_testdata.sql)) :)

This could also `DEAFULT` to `RS256`

client/client.go
```diff
@@ -263,7 +266,10 @@ func (c *Client) GetJSONWebKeys() *jose.JSONWebKeySet {
 }
 
 func (c *Client) GetTokenEndpointAuthSigningAlgorithm() string {
```
aeneasr, line 268: This new functionality needs a test: https://github.com/ory/hydra/blob/f0609ad96d8393ad96b670314faa0cff0ea39605/client/manager_test_helpers.go#L71

It also needs validation (and a test for validation):

https://github.com/ory/hydra/blob/master/client/validator.go#L116

ka3de:
Hi @aeneasr, thanks for your review, I was sure was gonna miss a few things.

I tried to address your comments.

Regarding the default `RS256` option for the new parameter, I decided to keep it at application level because I thought it's more clear for future developments, but if you believe it should be set at DB schema level let me know and I'll change that.

Thanks.

aeneasr:
Awesome, great work, this looks really solid to me! I agree with your comment but I think it would still be better to set the default for existing clients. This will make it easier for future queries for example because there is no need for implicit knowledge of the application's default token signing algorithm and also means that we can switch the default signing algorithm without breaking changes :)

ka3de:
Hi @aeneasr , thanks for giving quick feedback about this.

I agree with you that it can be misleading for DB queries to have the default option `RS256` set on application code, but I have some doubts regarding this, because I think it can be misleading to have an `hydra_client` with `token_endpoint_auth_signing_alg` to `RS256` when maybe its auth method is `client_secret_basic`.

So, do you agree on modifying the migration up to update the `hydra_client` table to set signing alg to `RS256` only for the clients which have auth method set to `private_key_jwt`?
E.g.:
```
ALTER TABLE hydra_client ADD token_endpoint_auth_signing_alg VARCHAR(10) NOT NULL DEFAULT '';
UPDATE TABLE hydra_client SET token_endpoint_auth_signing_alg = 'RS256' WHERE token_endpoint_auth_method = 'private_key_jwt'
```

And removing the default option for the "getter" method. E.g.:
```
func (c *Client) GetTokenEndpointAuthSigningAlgorithm() string {
	return c.TokenEndpointAuthSigningAlgorithm
}
```

What do you think?
Thanks.

aeneasr:
Sorry for my late reply and you're absolutely correct, that is the proper way to do it!

ka3de:
Hi @aeneasr , thanks for your reply.

I was working on this thinking that it was gonna be an easy change, but I've encountered an unexpected problem with cockroach migration. So my migration statement for cockroach was:
```
ALTER TABLE hydra_client ADD token_endpoint_auth_signing_alg VARCHAR(10) NOT NULL DEFAULT '';
UPDATE hydra_client SET token_endpoint_auth_signing_alg = 'RS256' WHERE token_endpoint_auth_method = 'private_key_jwt';
```
But this resulted in an error when performing migrations like:
```
driver.go:101
        	            				driver.go:168
        	            				asm_amd64.s:1357
        	Error:      Received unexpected error:
        	            	pq: column "token_endpoint_auth_signing_alg" does not exist
        	            	error executing 20200527215731_client.cockroach.up.sql, sql: ALTER TABLE hydra_client ADD token_endpoint_auth_signing_alg VARCHAR(10) NOT NULL DEFAULT '';
        	            	UPDATE hydra_client SET token_endpoint_auth_signing_alg = 'RS256' WHERE token_endpoint_auth_method = 'private_key_jwt';

```
I spent quite some time digging deeper on this, but could not find the root cause or explanation, I guess it has to do with they way cockroach processes the alter table because the same statement executed also by fizz is working fine for MySQL and Postgres, so I guess fizz is not the problem... I don't know it's a weird behavior honestly.

To work around this I was planning to "separate"  the cockroach migration in two migrations, one for the alter table and one for the update statement.
Let me know if you agree on this or if you know what could be the problem regarding the initial cockroach migration file which contains both, an alter table and an update table statement.

Thanks.

persistence/sql/migrations/schema.sql
```diff
@@ -0,0 +1,910 @@
+--
```
aeneasr: I believe this file was commited by accident?

persistence/sql/migrations/20200527215731_client.cockroach.up.sql
```diff
@@ -1 +1 @@
- ALTER TABLE hydra_client ADD token_endpoint_auth_signing_alg VARCHAR(10) NOT NULL DEFAULT '';
```
aeneasr: Rename to `20200527215731_client.up.sql`

persistence/sql/migrations/20200527215731_client.cockroach.down.sql
```diff
@@ -1 +1 @@
- ALTER TABLE hydra_client DROP COLUMN token_endpoint_auth_signing_alg;
+ALTER TABLE hydra_client DROP COLUMN token_endpoint_auth_signing_alg;
```
aeneasr, line 1: This is correct - basically you can rename this file to `20200527215731_client.down.sql` and override the existing file. By having `ALTER` and `UPDATE` in two files this should work for all migrations :)

persistence/sql/migrations/20200527215732_client.cockroach.up.sql
```diff
@@ -0,0 +1,3 @@
+-- complement for migration 20200527215731_client.cockroach.up
```
aeneasr, line 1: Rename to `20200527215732_client.up.sql`

persistence/sql/migrations/20200527215732_client.cockroach.down.sql
```diff
@@ -0,0 +1 @@
+UPDATE hydra_client SET token_endpoint_auth_signing_alg = '' WHERE token_endpoint_auth_method = 'private_key_jwt' AND token_endpoint_auth_signing_alg = 'RS256';
```
aeneasr, line 1: This should just be an empty file!

persistence/sql/migrations/20200527215731_client.up.sql
```diff
@@ -1 +1,2 @@
- ALTER TABLE hydra_client ADD token_endpoint_auth_signing_alg VARCHAR(10) NOT NULL DEFAULT '';
+ALTER TABLE hydra_client ADD token_endpoint_auth_signing_alg VARCHAR(10) NOT NULL DEFAULT '';
```
aeneasr, line 1: Delete this file

persistence/sql/migrations/20200527215731_client.down.sql
```diff
@@ -1 +1 @@
- ALTER TABLE hydra_client DROP COLUMN token_endpoint_auth_signing_alg;
```
aeneasr: Delete this file

aeneasr:
ps: tests fail but will be fixed with https://github.com/ory/hydra/pull/1897

ka3de:
Isn't it required an specific `cockroach` migration file for each migration? 
I mean, a `*_client.up.sql` migration file will be executed to MySQL and Postgres DBs, but to be applied to cockroach DB you need a specific `*_client.cockroach.up.sql` file. Did I got this wrong?

That's why I created specific files for coackroach and separated the `ALTER` statement from the `UPDATE` statement in these.
If I only have two `client.up.sql` files, one for the `ALTER` statement and one for the `UPDATE` statement, the migration test fails for cockroach DB saying that the new field does not exist.

aeneasr:
I believe this is no longer the case (which is why we refactored migrations) but hopefully @zepatrik can confirm this.

Also, I merged the CI fix on master so if you rebase tests should no longer fail :)

aeneasr:
Yup, confirmed :)

ka3de:
Yes, apologies, I could see migrations being applied for cockroach too by only using `client.[up|down].sql` files, I still get some migration test errors though, so I'm figuring it out.

aeneasr:
No worries!

aeneasr:
Great work and sorry for the long response times :)

ka3de:
No, thanks, feedback was great, I got tangled up with the migrations for such a small change honestly :D

BTW I had one consideration because we are stating that now it supports ECDSA, but that depends on the underlying Fosite library, which should be upgraded to v0.32 for that matter?

aeneasr:
yup! :)

Issue #1827, opened 2020-04-24
/userinfo endpoint misses www-authenticate header for 401 response


mikapfl:
Hi,

the [/userinfo API endpoint](https://www.ory.sh/hydra/docs/reference/api/#openid-connect-userinfo) returns a `401 Unauthorized` Response when the token is not correct, but does not send a `WWW-Authenticate` header.
According to [the spec for 401 Unauthorized](https://tools.ietf.org/html/rfc7235#section-3.1), a server MUST send a `WWW-Authenticate` header when generating a `401 Unauthorized` response.

Unfortunately, several OIDC resource provider libraries like for example mozilla-django-oidc require the `WWW-Authenticate` header (see for example https://github.com/mozilla/mozilla-django-oidc/blob/master/mozilla_django_oidc/contrib/drf.py#L80).
Therefore, it would be good if Hydra would send a `WWW-Authenticate` header as required by the spec.

1. Run hydra.
2. Make API Request to hydra's `/userinfo` endpoint with any invlaid token, for example `curl -i -H 'Authorization: bearer XXX' http://127.0.0.1:4444/userinfo`
3. Request fails with this output:
```
HTTP/1.1 401 Unauthorized
Content-Type: application/json
Date: Fri, 24 Apr 2020 14:04:11 GMT
Content-Length: 231

{"error":"request_unauthorized","error_description":"The request could not be authorized","error_hint":"Check that you provided valid credentials in the right format.","status_code":401,"error_debug":": not_found","request_id":""}
```
However, it should also contain a `WWW-Authenticate` header.

_Server logs_
```
INFO[0154] started handling request                      method=GET remote="127.0.0.1:41164" request=/userinfo
ERRO[0154] An error occurred while handling a request    code=401 debug= details="map[]" error=request_unauthorized reason="Check that you provided valid credentials in the right format." request-id= status=401 writer=JSON
INFO[0154] completed handling request                    measure#hydra/public: http://127.0.0.1:4444/.latency=252096 method=GET remote="127.0.0.1:41164" request=/userinfo status=401 text_status=Unauthorized took="252.096µs"
```

_Server configuration_

```yaml
# config for test servers
log:
  level: info

serve:
  public:
    port: 4444
    host: 127.0.0.1
    access_log:
      disable_for_health: false
  admin:
    port: 4445
    host: 127.0.0.1
    access_log:
      disable_for_health: false

dsn: memory
oauth2:
  expose_internal_errors: true
```

**Expected behavior**

[...]

PR #1891, opened 2020-06-03
oauth2: add www-authenticate header in case of unauthorized error for userinfo endpoint

bayansar:
## Related issue

Here is the issue that this PR is trying to fix `/userinfo endpoint misses www-authenticate header for 401 response` #1827

## Proposed changes

As far as I understood, apparently there are two cases that `/userinfo` API returns 401 error. One is if introspection API returns 401 and the other one is if token type is not `fosite.AccessToken`. Thus, I covered these two cases by adding `www-authenticate` header with reason coming from `RFC6749Error`

I've also added a new test case but needed small changes to make it happen.

This is my first PR, and I'm not familiar with the codebase. I'd be glad to hear better ways to do these changes if any.

## Checklist

      the changes.
      works.


aeneasr:
Good job and thank you for the contribution :)

heinzburgstaller:
This is the same for then /oauth2/token endpoint. If there is a 401 response the response header "WWW-Authenticate" is missing

aeneasr:
Contributions welcomed!

PR #1918, opened 2020-06-19
fix: Improve Makefile dependency management

### LINKED ISSUE #1916: Makefile checks for unneeded dependencies
The `Makefile` checks that dependencies for all possible targets are installed before running. This means if I only want to `make install-stable`, I still need `npm` and `docker-compose` installed.

dalcde:
This install dependencies only when you make a target that needs it.

This also removes the check that certain system dependencies (e.g. go)
are installed. Instead, we simply let the target fail. This ensures we
only test for the desired dependencies.

Fixes #1916 

Since I don't have the dependencies installed, I'm relying on CI to tell me if this is working properly...

      the changes.
      works.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,7 +49,7 @@ jobs:
           keys:
             - hydra-v1-{{ checksum "go.sum" }}-{{ checksum "Makefile" }}
       - run: go mod download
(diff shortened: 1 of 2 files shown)


aeneasr:
Great job, thank you!

PR #1919, opened 2020-06-20
chore: Optimize CircleCI workflow

sawadashota:
## Proposed changes

Optimize CircleCI workflow to catch error and end faster.

CircleCI jobs work 4 jobs maximum at same time because of free OSS plan.
In hydra, there are 12 jobs to be kicked at created PR but actually just each random 4 jobs proceed.

I think there are 2 problems.
First, we cannot catch error faster if similar and long jobs such as `test-e2e-*` or `test-legacy-*` are kicked.
Second, error PR blocks other PR because concurrency limitation is shared. If on of `test-e2e-*` failed, others are likely to fail for same reason.

So I changed workflow like this

before
image

after
image

## Checklist

      the changes.
      works.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -138,12 +138,16 @@ workflows:
               only: /.*/
       - test-legacy-migrations:
           name: test-legacy-migrations-mysql
+          requires:
+            - test-legacy-migrations-postgres
           flavor: mysql
           filters:
             tags:
               only: /.*/
       - test-legacy-migrations:
           name: test-legacy-migrations-cockroach
+          requires:
+            - test-legacy-migrations-postgres
           flavor: cockroach
           filters:
             tags:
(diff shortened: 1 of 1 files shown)


aeneasr:
Thank you! The problem really is CircleCI here which doesn't support canceling parallel jobs. Do you know if GitHub Actions does that? I think that the wait time is caused by the "executor" machine type which we currently use to check the legacy migrations. I was thinking of removing these tests once 1.5.2 is out because, unless we make changes, I don't think that there will be any issues with the new migration system.

aeneasr:
I just checked and it appears that GitHub actions supports up to 20 concurrent workflows per repository, which would be much much more than we have on CircleCI right now. Since it's free for open source without hard usage limits (as opposed to CircleCI) it might be something we need to switch to. Would be a lot of work though :/

sawadashota:
Yeah, GitHub Actions seems better and it will need  a lot of effort.
I couldn't find canceling parallel jobs spec or action but it seems it's possible if we use GitHub API.
https://developer.github.com/v3/actions/

.circleci/config.yml
```diff
@@ -217,12 +227,10 @@ workflows:
             - nancy/test
             - test
             - golangci/lint
-            - test-e2e-memory
             - test-e2e-postgres
             - test-e2e-mysql
             - test-e2e-cockroach
             - test-e2e-plugin
-            - test-legacy-migrations-postgres
```
aeneasr, line 225: Why did you remove these?

aeneasr:
Yeah I also saw that it's not possible to do manual job approval, which we currently use for sending out the newsletter :/

.circleci/config.yml
sawadashota: Because it still depend on indirectly. `test-legacy-migrations-mysql` and `test-legacy-migrations-cockroach` depend on `test-legacy-migrations-postgres`. So does `test-e2e-memory`.

aeneasr:
Thank you!

PR #1949, opened 2020-07-12
fix: logout error hint

phi2039:
Simple nitpick to correct a copy/paste error in the error_hint during a failed logout flow (now, with a corrected test case, too ;-) )


aeneasr:
Thank you! :)

Issue #1754, opened 2020-03-10
Client allowed_cors_origins not working


Alt0252:
**Describe the bug**

Enabling CORS support on the public endpoints seems to be working as expected by enabling the settings in the configuration, however when setting the `allowed_cors_origins` setting in the client, the additional origin URLs are not being allowed.

**Reproducing the bug**

1. Create a hydra environment version 1.3.2 with the below configuration
2. create an authorization_code client `my_client` with allowed_cors_origin set to an alternate (additional) Origin URL
3.  POST to  the /oauth2/token endpoint with the Origin: header (ie 'Origin: https://client-app.example.com'), and client_id: my_client in the body of the request. 
4.  Request should return the following response headers
access-control-allow-credentials: true
access-control-allow-origin: https://client-app.example.com

*Server logs*

*Server configuration*

**Expected behavior**

expect https://client-app.example.com to be an allowed origin

**Environment**

* Version: v1.3.2
* Environment: Docker

**Additional context**

aeneasr:
Would it be possible for you to provide a server config and client config (JSON or CLI command are ok) to make the reproduction easier?

This feature has several tests as it had some issues in the past and it will make bug hunting much easier :)

Alt0252:
huh... I swear I had the server config in my initial post....

```
serve:
  public:
    cors:
      enabled: true
      allowed_origins: 
        - https://**.example.com
  tls:
    allow_termination_from:
      - 10.0.0.0/16
strategies:
  access_token: jwt
urls:
  consent: https://login.example.com/consent
  login: https://consent.example.com/login
  self:
    issuer: https://hydra.example.com/
```

Client configuration
```{
    "client_id": "op-events-app",
    "client_name": "",
    "redirect_uris": [
        "https://myapp.example.biz/loggedIn",
    ],
    "grant_types": [
        "authorization_code",
        "refresh_token"
    ],
    "response_types": [
        "token",
        "code",
        "id_token"
    ],
    "scope": "openid offline offline_access",
    "audience": [
        "https://resource.server.example.com"
    ],
    "owner": "",
    "policy_uri": "",
    "allowed_cors_origins": [
        "https://myapp.example.biz",
    ],
    "tos_uri": "",
    "client_uri": "",
    "logo_uri": "",
    "contacts": [],
    "client_secret_expires_at": 0,
    "subject_type": "public",
    "token_endpoint_auth_method": "none",
    "userinfo_signed_response_alg": "none",
[...]

PR #1959, opened 2020-07-21
refactor: cleanup the code for CORS handling

harsimranmaan:
## Related issue

#1754 

## Proposed changes

This change was originally intended to fix the issue described in the bug. However, I am not able to repro it as shown in the newly added test.

## Checklist

      the changes.
      works.
- [N/A] I have added or changed [the documentation](docs/docs).

## Further comments

I removed many extra loops for validating the origin against the CORS allowlist. The viper config was shared across test cases causing dependency on the order of the test cases. It also hid some assumptions about the setup.

### DIFF
--- a/driver/cors_test.go
+++ b/driver/cors_test.go
@@ -34,7 +34,7 @@ import (
 	. "github.com/ory/hydra/driver"
 	"github.com/ory/hydra/internal"
 	"github.com/ory/hydra/oauth2"
-
+	"github.com/ory/hydra/x"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
@@ -68,7 +68,7 @@ func TestOAuth2AwareCORSMiddleware(t *testing.T) {
 				viper.Set("serve.public.cors.allowed_origins", []string{"http://not-test-domain.com"})
 			},
 			code:         http.StatusNotImplemented,
-			header:       http.Header{"Origin": {"http://foobar.com"}, "Authorization": {"Basic Zm9vOmJhcg=="}},
+			header:       http.Header{"Origin": {"http://foobar.com"}, "Authorization": {fmt.Sprintf("Basic %s", x.BasicAuth("foo", "bar"))}},
 			expectHeader: http.Header{"Vary": {"Origin"}},
 		},
 		{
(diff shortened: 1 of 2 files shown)


driver/cors_test.go
```diff
@@ -168,9 +178,21 @@ func TestOAuth2AwareCORSMiddleware(t *testing.T) {
...
+			d: "should accept when basic auth client exists and origin is set at the client as well as the server",
+			prep: func() {
+				viper.Set("serve.public.cors.enabled", true)
+				viper.Set("serve.public.cors.allowed_origins", []string{"http://**.example.com"})
+				r.ClientManager().CreateClient(context.Background(), &client.Client{ID: "foo8", Secret: "bar", AllowedCORSOrigins: []string{"http://myapp.example.biz"}})
+			},
+			code:         http.StatusNotImplemented,
+			header:       http.Header{"Origin": {"http://client-app.example.com"}, "Authorization": {"Basic Zm9vLTU6YmFy"}},
+			expectHeader: http.Header{"Access-Control-Allow-Credentials": []string{"true"}, "Access-Control-Allow-Origin": []string{"http://client-app.example.com"}, "Access-Control-Expose-Headers": []string{"Content-Type"}, "Vary": []string{"Origin"}},
+		},
```
aeneasr, line 191: I think the bug that was reported would be reproduced like this (checking against the origin domain, not the server one):

```

 		{
 			d: "should accept when basic auth client exists and origin is set at the client as well as the server",
 			prep: func() {
 				viper.Set("serve.public.cors.enabled", true)
 				viper.Set("serve.public.cors.allowed_origins", []string{"http://**.example.com"})
 				r.ClientManager().CreateClient(context.Background(), &client.Client{ID: "foo8", Secret: "bar", AllowedCORSOrigins: []string{"http://myapp.example.biz"}})
 			},
 			code:         http.StatusNotImplemented,
 			header:       http.Header{"Origin": {"http://myapp.example.biz"}, "Authorization": {"Basic Zm9vLTU6YmFy"}},
 			expectHeader: http.Header{"Access-Control-Allow-Credentials": []string{"true"}, "Access-Control-Allow-Origin": []string{"http://client-app.example.com"}, "Access-Control-Expose-Headers": []string{"Content-Type"}, "Vary": []string{"Origin"}},
 		},

```

harsimranmaan: Lemme give it a try. I exepect the Allow-Origin to be set to `http://myapp.example.biz`.

harsimranmaan: @aeneasr I could not repro the defect. Added the new test-case. Also refactored auth headers for readability.

aeneasr: Ok, in that case I don't think there's any more to do! And you are correct, the expected value was a copy/paste mistake on my end!

aeneasr:
Awesome 🎉 

PR #1971, opened 2020-08-05
feat: Expose introspection endpoint at OIDC Discovery

GaryGSC:
## Related issue
None. I did reference some similar PRs when creating this, though.

Refs: https://github.com/ory/hydra/pull/1355

## Proposed changes
[RFC 8414](https://tools.ietf.org/html/rfc8414) states that `introspection_endpoint` can optionally be added to `/.well-known/openid-configuration`. A library I'm using for OIDC Discovery would like to reference that property, but Hydra isn't currently providing it. This PR (I believe) should cause Hydra to provide that property.

## Checklist
      the changes.
- [X] I have added tests that prove my fix is effective or that my feature
      works.
- [X] I have added or changed [the documentation](docs/docs).

## Further comments

Hopefully, I updated all the docs where necessary. Please direct me if I haven't.


oauth2/handler.go
```diff
@@ -225,6 +225,7 @@ func (h *Handler) WellKnownHandler(w http.ResponseWriter, r *http.Request) {
 		TokenURL:                           urlx.AppendPaths(h.c.IssuerURL(), TokenPath).String(),
 		JWKsURI:                            urlx.AppendPaths(h.c.IssuerURL(), JWKPath).String(),
 		RevocationEndpoint:                 urlx.AppendPaths(h.c.IssuerURL(), RevocationPath).String(),
+		IntrospectionEndpoint:              urlx.AppendPaths(h.c.IssuerURL(), IntrospectPath).String(),
```
aeneasr, line 228: Introspection is running on the Admin endpoint and would thus be incorrectly referenced here.

GaryGSC: Okay, thanks for the review! That's fine with me if we reference the Admin endpoint.

I don't see a variable for the Admin endpoint anywhere in this file, nor can I find `h.c.AdminURL()` or any related-looking term. `OAuth2ClientRegistrationURL()` is about the closest that I see.

I imagine that if I can find the Admin endpoint, we would use that instead of `h.c.IssuerURL()`. Could you please direct me to where I could find that?

aeneasr: Yeah that doesn't exist because it wasn't needed until now. Why don't we create - similar to e.g. `h.c.OIDCDiscoverySupportedClaims(),` a new config value that, if set, sets this key?

aeneasr:
Closing due to inactivity

Issue #2007, opened 2020-08-21
Document GitLab setup with ORY Hydra as the OpenID Connect Provider


aeneasr:
**Is your feature request related to a problem? Please describe.**

We combined GitLab with ORY Hydra and should [write a guide](https://github.com/ory/hydra/tree/master/docs/docs/guides) in the ORY Hydra docs on it.

The private repository for this is at: https://github.com/ory/hydra-gitlab-example

Related #2005 #2006

Let's also do this soon as the memory is still fresh, hopefully :D

aeneasr:
This is resolved on master!

PR #2014, opened 2020-08-24
docs: gitlab hydra integration #2007

k9ert:
## Related issue

#2007 @aeneasr`

## Proposed changes

Documenting the integration with gitlab. We're not directly using OpenID-connect in this guide but the reasons are mentioned in the appendix of gitlab.mdx.

I have not added this document to any version yet.

## Checklist

- [ X] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [ X ] I have read the [security policy](../security/policy).
- [ X ] I confirm that this pull request does not address a security
      the changes.
- [  ] I have added tests that prove my fix is effective or that my feature
      works.
- [ X] I have added or changed [the documentation](docs/docs).


docs/docs/guides/gitlab.mdx
```diff
@@ -0,0 +1,435 @@
+---
+id: gitlab
+title: Gitlab Hydra integration
+---
+
+import useBaseUrl from '@docusaurus/useBaseUrl'
```
aeneasr, line 6: This is no longer required :)

docs/docs/guides/gitlab.mdx
```diff
@@ -0,0 +1,435 @@
...
+
+Gitlab has several OAuth2 related features. The relevant here is the possibility
+to sign into GitLab with (almost) any OAuth2 provider, in this case hydra. The documentation for this
```
aeneasr, line 9:
```suggestion
to sign into GitLab with (almost) any OAuth2 provider, in this case ORY Hydra. The documentation for this
```

docs/docs/guides/gitlab.mdx
```diff
@@ -0,0 +1,435 @@
...
+is quite generic and also not very detailed. So, in this guide, we'll connect
```
aeneasr, line 10:
```suggestion
is kept generic on purpose. So, in this guide, we'll connect
```

docs/docs/guides/gitlab.mdx
```diff
@@ -0,0 +1,435 @@
...
+gitlab's omniauth-connector to ORY Hydra. We'll do that in a docker-based
```
aeneasr, line 11: Please also add a link to omniauth-connector!

```suggestion
GitLab's omniauth-connector to ORY Hydra. We'll do that in a docker-based
```

docs/docs/guides/gitlab.mdx
```diff
@@ -0,0 +1,435 @@
...
+command-line-client available is quite useful. So let's install hydra as
+explained in the [installation-guide](../install).
+You'll also need [docker](https://docs.docker.com/get-docker/) and
+[docker-compose](https://docs.docker.com/compose/install/).
+
+The [hydra-5-min-tutorial](../5min-tutorial) might
+be worth checking out upfront. It'll a give a nice quick overview how OAuth2 is
+working within hydra with a minimal example. We assume basic knowledge, here.
+
+If you have not yet the source-code of hydra, which we'll need for the docker-compose
```
aeneasr, line 27:
```suggestion
If you have not yet the source code of hydra, which we'll need for the `docker-compose`
```

docs/docs/guides/gitlab.mdx
```diff
@@ -0,0 +1,435 @@
...
+yaml-files, clone this repository:
+
+```shell
+git clone git@github.com:ory/hydra-gitlab-example.git
```
aeneasr, line 31: Is this correct? If so, we need to open source that first, right?

docs/docs/guides/gitlab.mdx
```diff
@@ -0,0 +1,435 @@
...
+```
+
+## Spin up the instances
+
+Gitlab has some [documentation](https://docs.gitlab.com/omnibus/docker/) about
+how to use their docker-images. It has also an example for docker-compose. So change into the directory with the Hydra 
+source code and create a new file quickstart-gitlab.yml with this content:
```
aeneasr, line 38:
```suggestion
source code and create a new file `quickstart-gitlab.yml` with this content:
```

docs/docs/guides/gitlab.mdx
```diff
@@ -0,0 +1,435 @@
...
+      - "443:443" # https
+      - "8000:8000" # http
+      - "2222:22" # ssh
+    volumes:
+    - '$GITLAB_HOME/config:/etc/gitlab'
+    - '$GITLAB_HOME/logs:/var/log/gitlab'
+    - '$GITLAB_HOME/data:/var/opt/gitlab'
+```
+
+Now we can spinup the instances.
```
aeneasr, line 61:
```suggestion
Now we can spinup the instances:
```

docs/docs/guides/gitlab.mdx
```diff
@@ -0,0 +1,435 @@
...
+
+```shell
+export GITLAB_HOME=$(pwd)/gitlab
+docker-compose -f quickstart.yml \
+    -f quickstart-postgres.yml -f quickstart-gitlab.yml \
+    up --build
+```
+
+Gitlab will take some time until it's up. You'll see the logs in the output. So you need to open a new terminal for the next steps.
```
aeneasr, line 70:
```suggestion
GitLab will take some time until it's up. You'll see the logs in the output. So you need to open a new terminal for the next steps. 
```

docs/docs/guides/gitlab.mdx
```diff
@@ -0,0 +1,435 @@
...
+
+The gitlab-configuration can be modified in the container like such:
+
+```shell
+docker-compose -f quickstart.yml  -f quickstart-postgres.yml -f quickstart-gitlab.yml exec gitlab vim /etc/gitlab/gitlab.rb
+```
+
+Using vim might not be your first choice and if you never used it, it's quite
+unintuitive and confusing but it's worth learning the basics to do simple changes like this. Check the basics
+[here](https://www.youtube.com/watch?v=ggSyF1SVFr4).
```
aeneasr, line 92: Why not use `nano` instead here? Might help people get this done quicker

docs/docs/guides/gitlab.mdx
```diff
@@ -0,0 +1,435 @@
...
+
+The first configuration is already done by overriding the
+[external_url](https://docs.gitlab.com/omnibus/settings/configuration.html#configuring-the-external-url-for-gitlab)
+via an environment-variable within the quickstart.yaml. Now, at around line 433,
+add this:
+
+```ruby
```
aeneasr, line 99: Is it not possible to include this file in the GitHub repository and tell gitlab to mount it? That way we could link people to it and you could make modifications using your IDE instead of SSHing

docs/docs/guides/gitlab.mdx
```diff
@@ -0,0 +1,435 @@
...
+      # of this strategy (appears in urls and Gitlab login buttons)
+      # If you do this, you must also replace oauth2_generic, everywhere it appears above, with the new name.
+      name: 'ORY_Hydra', # display name for this strategy
+      #strategy_class: "OmniAuth::Strategies::OAuth2Generic" # Devise-specific config option Gitlab uses to find renamed strategy
+    }
+  }
+]
+```
+
+After you did these changes, restart the container:
```
aeneasr, line 126: NOTE TO SELF: Reviewed until here

docs/docs/guides/gitlab.mdx
k9ert: No, it's not correct. Instead the user should checkout hydra. In the 5min-tutorial that is omitted and it starts with "change to the directory where the hydra-sourcecode ..." but i like the idea to include that in the Guide. Happy to also remove it completely. What do you think?

Csirme:
I'm so happy to see that you've written the document quickly. I plan to build an internal SSO with Hydra, and I'll try to integrate it into gitlab, rancher, or others.

Thank you for your excellent product. It's meaningful

docs/docs/guides/gitlab.mdx
k9ert: Maybe this is doable but i have at least some concerns:
1. What's wrong with asking people to do such complex things like editing files on folders. It's also a kind of filter. I don't want to argue with someone how to edit files. If he doesn't know that, he should learn it first.
2. We would pollute the repo with such non-hydra-specific stuff. The same would be true for the quickstart-gitlab.yml which i avoided to create upfront because of this argument.
3. Potential issues with access-rights. The gitlab-container seem to run as root and so that file is not even readable on the "host".  It is probably solvable but  it's will also probably increase non-hydra specific complexity of the guide and/or will encourage insecure practices. Maybe 3 is entirely wrong but ...

I can't see much pro-arguments other than it's more convenient. This is not a typical dev-guide. This is an operations-guide and operators core-compentency should be dealing with such issues seamlessly. It's also not very complicated as the guide-user simply needs to copy&paste a command.
We have something very similiar with the hydra-binary. Instead of installing it manually on the host, the command in the 5min-tutorial gets a bit longer and uses the binary on the container.

docs/docs/guides/gitlab.mdx
k9ert: Yes, good point.

docs/docs/guides/gitlab.mdx
aeneasr: > Maybe this is doable but i have at least some concerns:

Good documentation that is easy to follow is what we should strive for. Great products have great documentation and the time it takes us to implement will be worth the investment if we have 25 happy adopters who just got it working with little effort and frustration. There is already enough frustration involved in OAuth2 and adjacent products.

> We would pollute the repo with such non-hydra-specific stuff. The same would be true for the quickstart-gitlab.yml which i avoided to create upfront because of this argument.

Typically, stuff like this would end up in the `/contrib` folder, see for example https://github.com/ory/kratos/tree/master/contrib/quickstart

I would be ok with having this here. We could also move the GitLab docker-compose file there, which would clean up the root view.

> I can't see much pro-arguments other than it's more convenient. This is not a typical dev-guide. This is an operations-guide and operators core-compentency should be dealing with such issues seamlessly. It's also not very complicated as the guide-user simply needs to copy&paste a command.

The problem I see is that editing stuff in Docker Container FS is usually a bad idea because the FS is ethermal. A simple volume clean up or accidental kill will mean that you have to redo everything. It's just not good practice and given the very easy alternatives (mounting a volume with the config). Most ops people I encountered will bash you with a club if you add configuration to a Docker Image (in prod/stage) 😅

docs/docs/guides/gitlab.mdx
aeneasr: No let's keep it here. There will be at least 10 people who do not understand that they need to run the commands (and especially which ones!) from another link. It's better to have explicit instructions.

You can move the gitlab docker-compose files to `contrib/gitlab`

docs/docs/guides/gitlab.mdx
```diff
@@ -0,0 +1,427 @@
...
+services:
+  gitlab:
+    image: gitlab/gitlab-ce:latest
+    restart: always
+    hostname: gitlab.example.com
+    environment:
+      GITLAB_OMNIBUS_CONFIG: |
+        external_url 'http://gitlab.example.com:8000/'
+    ports:
+      - "443:443" # https
```
aeneasr, line 50: Is `443` a privileged port that needs sudo?

docs/docs/guides/gitlab.mdx
k9ert: Fully agree to the importance of good documentation :-).

> Most ops people I encountered will bash you with a club if you add configuration to a Docker Image (in prod/stage) 

I guess i'd be one of them! However, a POC-Guide like that is not a prod-setup. There are dozens of additional things you should consider if you're running on prod, some of them in the hydra-context but most in a very generic way.

Let me see how complicated that would be in the contrib-folder like the kratos one.

docs/docs/guides/gitlab.mdx
k9ert: IMHO yes, as all ports < 1024 are priviledged. However, the port is not used in the POC-guide , so i'll remove that mapping! Also the ssh-port is not used. The editing is currently not done via ssh and we even consider to do the editing via the host-file-system. So i'll remove that one as well.

docs/docs/guides/gitlab.mdx
```diff
@@ -0,0 +1,433 @@
...
+      - "8000:8000" # http
```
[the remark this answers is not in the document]
k9ert: Why do you want to change it to 80? This is also a privileged port. Depending on how docker is installed on a system, this might or might not be an issue. On Linux it's not a problem as the dpkg-based installation is using root-privileges. However it's a good idea to avoid issues with that.

docs/docs/guides/gitlab.mdx
aeneasr: SGTM :)

docs/docs/guides/gitlab.mdx
```diff
@@ -0,0 +1,427 @@
...
+      - "2222:22" # ssh
+    volumes:
+    - '$GITLAB_HOME/config:/etc/gitlab'
+    - '$GITLAB_HOME/logs:/var/log/gitlab'
+    - '$GITLAB_HOME/data:/var/opt/gitlab'
```
aeneasr, line 56: Given that `$GITLAB_HOME` is set to `GITLAB_HOME=$(pwd)/gitlab` we can probably reduce errors (e.g. on Windows) by just using `./gitlab/config:/etc/gitlab`

k9ert:
I've changed the whole guide to a "quick-up-and-running+troubleshooting" style, rather than a "failfast-and-fix-step-by-step" style.
Another review triggered!

docs/docs/guides/gitlab.mdx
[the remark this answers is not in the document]
Csirme: Sorry, I can't understand the official image “gitlab/gitlab-ce:13.0.6-ce.0” code is “ EXPOSE 22 443 80”，Why do you use 8000 to map 8000，Not 8000 mapping 80 .

aeneasr (approved):

Issue #2031, opened 2020-08-31
Expired token is considered an error


catper:
**Is your feature request related to a problem? Please describe.**

When Hydra receives a request with a token that is no longer valid, it logs an error with what appears to be debug output:
```
{"debug":"Token is expired","description":"Token expired","error":"token_expired","hint":"The token expired.","level":"error","msg":"An error occurred","time":"2020-08-28T20:14:40Z"}
```

At high loads, logs can get completely bogged down with these messages, which isn't only annoying but also affects the performance of the application as I/O is expensive.

**Describe the solution you'd like**

An invalid token does not indicate an error with the application and therefore I don't think it should be logged as such. Since it is not an event that requires action, I'd argue that it should be on a level no higher than `info`.

aeneasr:
Yeah this should definitely move to the `AuditLogger()`. Contributions welcomed! :)

PR #2038, opened 2020-09-04
WIP: Downgrade log level for access rejections as a consequence of invalid credentials/expired token

catper:
## Related issue

[#2031](https://github.com/ory/hydra/issues/2031)

## Proposed changes

Invalid credentials or an expired token is a normal thing to happen and should be logged on "info" level, not on "error" level as is currently the case.

## Checklist

      the changes.
      works.

### DIFF
--- a/oauth2/handler.go
+++ b/oauth2/handler.go
@@ -551,10 +551,12 @@ func (h *Handler) TokenHandler(w http.ResponseWriter, r *http.Request) {
 	accessRequest, err := h.r.OAuth2Provider().NewAccessRequest(ctx, r, session)
 
 	if err != nil {
-		switch err.Error() {
-		case fosite.ErrServerError.Error():
-		case fosite.ErrTemporarilyUnavailable.Error():
-		case fosite.ErrMisconfiguration.Error():
+		switch errors.Cause(err) {
+		case fosite.ErrServerError:
+			fallthrough
+		case fosite.ErrTemporarilyUnavailable:
+			fallthrough
+		case fosite.ErrMisconfiguration:
 			x.LogError(r, err, h.r.Logger())
 		default:
 			x.LogAudit(r, err, h.r.Logger())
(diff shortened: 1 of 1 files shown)


aeneasr:
Looking great, thank you!

oauth2/handler.go
```diff
@@ -587,7 +587,7 @@ func (h *Handler) TokenHandler(w http.ResponseWriter, r *http.Request) {
 
 	accessResponse, err := h.r.OAuth2Provider().NewAccessResponse(ctx, accessRequest)
 	if err != nil {
-		x.LogError(r, err, h.r.Logger())
+		x.LogAudit(r, err, h.r.Logger())
```
aeneasr, line 590: Why did you change this here? :)

catper: perhaps i'm mistaken but is this not where the error is logged in case credentials are invalid? my aim is for that error log to be downgraded to something less severe.

aeneasr: That is correct, however there could be a number of issues which would warrant a error log such as a broken database connection and more. Maybe we should check for the error type and choose the logger only for cases where invalid credentials are sent?

catper: okay, my bad, i thought invalid credentials was the only remaining error here. will fix!

oauth2/handler.go
```diff
@@ -549,8 +549,16 @@ func (h *Handler) TokenHandler(w http.ResponseWriter, r *http.Request) {
 	var ctx = r.Context()
 
 	accessRequest, err := h.r.OAuth2Provider().NewAccessRequest(ctx, r, session)
+
 	if err != nil {
-		x.LogError(r, err, h.r.Logger())
+		switch err.Error() {
+		case fosite.ErrServerError.Error():
+		case fosite.ErrTemporarilyUnavailable.Error():
```
aeneasr, line 556: Golang does not automatically fallthrough, you need to add the fallthrough statement to do so. In general I think you could simplify this as:

```
if errors.Is(err, fosite.ErrUnauthorizedError) {
  x.LogAudit(r, err, h.r.Logger())
} else {
  x.LogError(r, err, h.r.Logger())
}
```

catper: sorry about the fallback, guess that shows how much go experience i have ...

the problem with using `ErrRequestUnauthorized` (I assume this is what you mean, there doesn't seem to be an `ErrUnauthorizedError`?) is that that's not actually what's returned, we get something more specific like `ErrInvalidClient`, `ErrUnsupportedGrantType` or similar.

aeneasr: No problem, it is a good learning experience :)

The error I meant was `fosite. ErrUnauthorizedClient` and `fosite.ErrInactiveToken`

catper: the error comparisons won't return true though. one could potentially check e.g.  `strings.Compare(err.Error(), fosite.ErrInvalidClient.Name) == 0` instead but not sure that that's the nicest solution.

regardless, i still think that anything that doesn't correspond to a status code >= 500 should be passed to the audit log rather than the errror log.

zepatrik: You would want to use `errors.Cause(err) == fosite.Err...` from https://github.com/pkg/errors#retrieving-the-cause-of-an-error

aeneasr (approved):
Awesome 🎉 

PR #2048, opened 2020-09-10
docs: minor text typo for limitations.md

jhuggett:
It said "an maximum" but I believe it should be "a maximum".


aeneasr:
Thank you!

Issue #2020, opened 2020-08-26
Update clients from cli


dselyuzhitskiy:
Hello!

It will be great to have `hydra clients update` command because sometimes we need to add/delete client callbacks. Do you have any plans to add such abitlity? Yes, i know that it is possible through rest api, but it is more comfortable to do this by hydra cli.

aeneasr:
Haha I just closed another issue for this recently - sounds good :)

matheusfm:
I think an 'upsert' strategy can also be useful. What do you think?

dselyuzhitskiy:
What about update like PATCH instead of PUT?  For me 'update' is primarily for updating fields that i specified in request.  I am writing service for updating client callbacks and for this operation i need to get client from hydra and then provide all client configuration with changed fields instead of making just one request with changed fields. What do you think about it?

aeneasr:
PATCH is now available!

erolkskn:
Hi @aeneasr, i want to contribute to this project, by taking this issue. After looking for cobra's documentation about flags, i've seen that there is no concept that provides "nullable flag" in cobra(or should i say pflag). A flag's value will always be set to it's default, even though user provides no value at all. But in order to make mentioned feature in this issue work, users should be able to skip some flags, and those skipped flags should not be included in the body of JSONPatch. Afaik Only way to get this kind of functionality is looking up whether flag is changed by user at all. This can be accomplished by something like this: `cmd.Flags().Lookup("flagName").Changed`, if it equals to false it means that user didn't provide the flag, so it shouldn't be included in JSONPatch body. 

So here is the proposed solution, a helper method that has the signature like this: 
`func NewJsonPatchFromFlags(flags *pflag.FlagSet, map[string]string mapping) *JsonPatch` can be made. 
The `mapping` map consists of pair "flagName" and "jsonPatchPath". 
For example: `{"frontchannel-logout-callback": "/frontchannel_logout_callback"}` this mapping configuration tells that if user sets the value of flag named *frontchannel-logout-callback*, a **replace** operation with path named */frontchannel_logout_callback* would be added to the resulting **JsonPatch Doc** so it would look like this: 
`[{"op": "replace", "path": "/frontchannel_logout_callback", "value": "given value from cli"}]`. I think that this functionality might be placed on "ory/x" package, it might be useful on other projects in ORY ecosystem (Kratos, Gatekeeper etc.).

[...]

PR #2049, opened 2020-09-11
feat: add client update command in CLI

matheusfm:
## Related issue

#2020 

## Proposed changes

Add client update command in CLI.

## Checklist

      the changes.
      works.

## Further comments

Only if you pass the `secret` flag, the client secret will be updated and returned.
The same strategy as the API.

### DIFF
--- a/cmd/clients_update.go
+++ b/cmd/clients_update.go
@@ -7,14 +7,17 @@ import (
 // updateCmd represents the update command
 var clientsUpdateCmd = &cobra.Command{
 	Use:   "update <id>",
-	Short: "Update an OAuth 2.0 Client",
-	Long: `This command updates an OAuth 2.0 Client by its ID.
+	Short: "Update an entire OAuth 2.0 Client",
+	Long: `This command replaces an OAuth 2.0 Client by its ID.
 
-Example:
-  hydra clients update client-1 -n "my app" -c http://localhost/cb -g authorization_code -r code -a core,foobar
+Please be aware that this command replaces the entire client.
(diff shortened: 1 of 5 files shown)


cmd/clients_update.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"github.com/spf13/cobra"
+)
+
+// updateCmd represents the update command
+var clientsUpdateCmd = &cobra.Command{
+	Use:   "update <id>",
+	Short: "Update an OAuth 2.0 Client",
+	Long: `This command updates an OAuth 2.0 Client by its ID.
```
aeneasr, line 11: Given that an update is a PUT and not a PATCH operation (it replaces the client) I think we should explicitly tell people that here. Otherwise someone might think that he/she can change the grant type only but instead everything else is deleted.

aeneasr (changes requested):
Cool, thank you! Here are some ideas for improvement!

cmd/clients_update.go
matheusfm: Thanks for your review! 
I agree with your comments and I will improve the description.

What do you think of CLI supporting partial updates? With a flag or other command, for example: 
`hydra clients update client-1 --partial -g client_credentials` 
or 
`hydra clients patch client-1 -g client_credentials`
Maybe we can discuss this in another issue.

matheusfm:
@aeneasr I changed the command description! Could you review again?

aeneasr (approved):
Awesome!

Issue #1728, opened 2020-02-18
Add endpoint to Admin API to revoke access tokens


aeneasr:
**Is your feature request related to a problem? Please describe.**

It's currently not possible to revoke access tokens from `client_credentials` grants. This has been requested and I think it is a good idea to allow deletion of tokens based on the client_id. This makes sense if you lose trust in a client, for example, or if the secret changes.

**Describe the solution you'd like**

Similar to [this API endpoint](https://www.ory.sh/docs/hydra/sdk/api#revokes-consent-sessions-of-a-subject-for-a-specific-oauth-20-client) it should be possible to revoke tokens without a consent session attached as well. A good endpoint would be:

```
DELETE <hydra-admin>/oauth2/tokens?client_id=...
```

**Additional context**

https://community.ory.sh/t/disable-client-in-hydra-for-a-while-and-re-enable-it/1510/8

lpedrosa:
This would be a pretty nice addition.

Would being able to revoke individual `client_credential` grants make sense as well? Or would that go against the use-case of this type of grant i.e. service-to-service authz?

I'm thinking about multiple instance of the same service holding different tokens.

aeneasr:
> Would being able to revoke individual client_credential grants make sense as well? Or would that go against the use-case of this type of grant i.e. service-to-service authz?

I think it's quite difficult to track that from an admin perspective because there is no session identifier. Also, client_credential doesn't have refresh tokens so they expire after an hour.

sagarshah1983:
Would this also result in deletion of access tokens created through authorization code workflow for the same client?

PR #2058, opened 2020-09-17
feat: Admin API for deleting access token of a client

ajanthan:
## Related issue
#1728

## Proposed changes

Introduced a new admin API for deleting tokens of a client. The API is in the following format

`DELETE <hydra-admin>/oauth2/tokens?client_id=...`

## Checklist

      the changes.
      works.

## Further comments

I haven't generated the documentation since my local copy was changing the format of the existing contents. If that is ok I can run the doc gen task and commit the generated docs.


zepatrik (changes requested):
Thank you, this looks good :+1: 
I think the only thing that's missing is tests for the storage.

ajanthan:
> Thank you, this looks good 👍
> I think the only thing that's missing is tests for the storage.

Added a new test for the storage

zepatrik (approved):
Thank you

oauth2/handler.go
```diff
@@ -726,6 +727,37 @@ func (h *Handler) forwardError(w http.ResponseWriter, r *http.Request, err error
 	http.Redirect(w, r, urlx.CopyWithQuery(h.c.ErrorURL(), query).String(), http.StatusFound)
 }
 
+// swagger:route DELETE /oauth2/token admin deleteOAuth2Token
```
aeneasr, line 730: I would prefer to not re-use the OAuth2 endpoint here but instead make it an explicit REST resource by using the plural:

```suggestion
// swagger:route DELETE /oauth2/tokens admin deleteOAuth2Token
```

What do you think?

ajanthan: +1

aeneasr (approved):
Awesome 🎉 

Issue #2072, opened 2020-09-29
Prometheus endpoint should not require x-forwarded-proto header


amihalj:
Similar to heathcheck, metrics endpoint should not require x-forwarded-proto.

We are running hydra service in K8 cluster as TLS terminated but we are unable to scrape metrics due to issue with header. Is there a way to exclude it like in health routes?

thank you,
Ante

aeneasr:
Makes sense!

amihalj:
I will provide PR, please take a loot at it and include in next release if possible.

thanks.

amihalj:
Hi @aeneasr , will this PR be included in 1.8.x version or it will wait for 1.9.0 version?  we would rather have it sooner if it is possible. thank you

aeneasr:
Hopefully we get 1.8 properly released today, had some upstream issues - so this will be 1.8.x

PR #2074, opened 2020-09-30
feat: metrics prometheus endpoint should not require x-forwarded-proto header

amihalj:
The purpose of this PR is to allow /metrics/prometheus endpoint to be invoked without https or x-forwarded-proto: https header.
In some cases prometheus is already in cluster behind LBS and (without a lot of hassle) it does not set this header.

changes:

- moved MetricsPrometheusPath constant to metrics/prometheus/metrics.go
- added rule to allow insecure requests for MetricsPrometheusPath endpoint (the same as for health routes)
- arranged tls_termination_test.go test to cover all cases in RejectInsecureRequests function

relates to issue #2072
@aeneasr , please take a look at this PR

## Checklist

      the changes.
      works.

### DIFF
--- a/x/tls_termination_test.go
+++ b/x/tls_termination_test.go
@@ -35,27 +35,18 @@ func TestDoesRequestSatisfyTermination(t *testing.T) {
 		assert.EqualValues(t, http.StatusBadGateway, res.Code)
 	})
 
-	t.Run("case=missing-x-forwarded-proto", func(t *testing.T) {
-		viper.Set(configuration.ViperKeyAllowTLSTerminationFrom, "126.0.0.1/24,127.0.0.1/24")
-
+	t.Run("case=forced-http", func(t *testing.T) {
+		c := internal.NewConfigurationWithDefaults()
 		res := httptest.NewRecorder()
-		RejectInsecureRequests(r, c)(res, &http.Request{Header: http.Header{}, URL: new(url.URL)}, panicHandler)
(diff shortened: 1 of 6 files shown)


x/tls_termination_test.go
```diff
@@ -35,51 +35,83 @@ func TestDoesRequestSatisfyTermination(t *testing.T) {
 		assert.EqualValues(t, http.StatusBadGateway, res.Code)
```
aeneasr, line 35: The diff here is very difficult to read for me because it looks like the test order has been changed? Could you please do one of the two:

- Reverse the order change and append your new test cases to the bottom and change the test cases you might have changed
- Comment on every existing test change why it was changed

aeneasr (changes requested):
Awesome, thank you for your contribution! This looks pretty good but I have some ideas how to improve it further :)

x/tls_termination_test.go
amihalj: I have ordered tests like original and have placed 2 new tests at the bottom. for changed tests I have made a comment why I have changed them.

amihalj: I have previously reordered tests to follow the actual flow of steps execution, but it made PR reviewing hard. sorry for that.

aeneasr: All good, thank you for revisiting!

aeneasr (approved):
Awesome 🎉 

PR #2131, opened 2020-10-14
fix: revert additional usage of transaction

zepatrik:
## Related issue
Performance.

## Proposed changes
@SzilardS this reverts all the additional transactions so that there are only transactions where they are on the current master as well. Can you have another look on the performance?


aeneasr:
Closing because the transaction merged currently is the correct way to go.

PR #2175, opened 2020-11-16
ci(docs): rename consts key in package.json to config

zepatrik:
## Related issue
npm@v7 implements https://github.com/npm/rfcs/blob/latest/implemented/0021-reduce-lifecycle-script-environment.md
This means that only stuff under the `config` key can be referenced in scripts. This change is still compatible with npm@v6.

## Proposed changes

Rename `consts` to `config` in the package.json


aeneasr:
Shouldn't this be done in docusaurus-template?

zepatrik:
yes

PR #2183, opened 2020-11-18
feat: Only pass state if it was set

jaylinski:
## Related issue

I didn't open an issue because I prefer to directly show the problem by submitting a PoC.

## Proposed changes

Using `state` in the logout flow is optional, so `state` can be
empty. In order to avoid an ugly `/post-logout-redirect-uri?state=`
URI, the state should only be appended if it is not empty.

Relevant specification: https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout

## Checklist

      the changes.
      works.

## Further comments

If you are willing to accept this change, I'll add tests (and fix the typo in the commit).

### DIFF
--- a/consent/strategy_default_test.go
+++ b/consent/strategy_default_test.go
@@ -181,7 +181,7 @@ func logoutHandler(strategy Strategy, writer herodot.Writer, w http.ResponseWrit
 	http.Redirect(w, r,
 		urlx.CopyWithQuery(
 			urlx.ParseOrPanic(res.RedirectTo),
-			url.Values{"front_channel_logout_urls": {fmt.Sprintf("%s", res.FrontChannelLogoutURLs)}},
+			url.Values{},
 		).String(),
 		http.StatusFound,
 	)
@@ -243,6 +243,7 @@ func runLogout(t *testing.T, method string) {
 		sessionID        string
 		lph              func(t *testing.T) func(w http.ResponseWriter, r *http.Request)
 		expectBody       string
+		expectRequestURI string
 		backChannels     []func(t *testing.T) func(w http.ResponseWriter, r *http.Request)
 		expectStatusCode int
 		jar              http.CookieJar
@@ -416,6 +417,7 @@ func runLogout(t *testing.T, method string) {
 			expectStatusCode: http.StatusOK,
 			jar:              newValidAuthCookieJar(t, reg, logoutServer.URL, "logout-session-temp5", "logout-subject-temp5"),
 			expectBody:       "redirected to default server1234custom",
+			expectRequestURI: "/custom?state=1234",
 		},
 		{
 			d:   "should pass rp-inititated flow",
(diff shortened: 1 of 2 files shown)


aeneasr:
Thank you for your contribution! Could you please provide a reference to the specification where this is defined?

jaylinski:
> Thank you for your contribution! Could you please provide a reference to the specification where this is defined?

It is specified here: https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout

aeneasr:
Looking good, thank you! Could you maybe also add a test to prevent regressions here?

consent/strategy_default_test.go
```diff
@@ -181,7 +181,7 @@ func logoutHandler(strategy Strategy, writer herodot.Writer, w http.ResponseWrit
 	http.Redirect(w, r,
 		urlx.CopyWithQuery(
 			urlx.ParseOrPanic(res.RedirectTo),
-			url.Values{"front_channel_logout_urls": {fmt.Sprintf("%s", res.FrontChannelLogoutURLs)}},
```
jaylinski, line 184: I removed this because it made the assertions easier (and I don't know what it was for).

consent/strategy_default_test.go
```diff
@@ -463,6 +465,26 @@ func runLogout(t *testing.T, method string) {
...
+					"exp": time.Now().Add(-time.Hour).Unix(),
+					"iat": time.Now().Add(-time.Hour * 2).Unix(),
+				})},
+			},
+			expectStatusCode: http.StatusOK,
+			jar:              newValidAuthCookieJar(t, reg, logoutServer.URL, "logout-session-temp7", "logout-subject-temp7"),
+			subject:          "logout-subject-7",
+			expectBody:       "redirected to default servercustom",
+			expectRequestURI: "/custom",
+		},
```
jaylinski, line 487: An alternative would be to remove the `state` value from `expectBody` and instead add the `expectRequestURI` to all test-cases.

consent/strategy_default_test.go
```diff
@@ -522,6 +544,10 @@ func runLogout(t *testing.T, method string) {
 
 			assert.EqualValues(t, tc.expectStatusCode, resp.StatusCode, "%s\n%s", resp.Request.URL.String(), out)
 			assert.EqualValues(t, tc.expectBody, strings.Trim(string(out), "\n"), "%s\n%s", resp.Request.URL.String(), out)
+
+			if tc.expectRequestURI != "" {
+				assert.EqualValues(t, tc.expectRequestURI, resp.Request.URL.RequestURI(), "%s\n%s", resp.Request.URL.String(), out)
+			}
```
jaylinski, line 550: The `if` could be removed if a `expectRequestURI` would be added to all test-cases.

jaylinski:
> Looking good, thank you! Could you maybe also add a test to prevent regressions here?

@aeneasr I added a test case. Please review. :octocat:

consent/strategy_default_test.go
aeneasr: Hm strange, looks like this isn't used anywhere. Maybe it was there for an older test case, it should be ok to remove.

aeneasr (approved):
Awesome 🎉 

jaylinski:
@aeneasr Thanks for this nice piece of software! :raised_hands: Greetings from Linz.

Issue #2171, opened 2020-11-11
SDK documentation for javascript client returns a 404 page not found


CSRaghunandan:
API documentation for nodejs client for hydra returns a 404 

The link mentioned in the documentation: https://github.com/ory/hydra/blob/master/sdk/js/swagger/README.md

zepatrik:
Right, the whole page needs an overhaul https://github.com/ory/hydra/blob/master/docs/docs/sdk/js.md
The SDKs are now at https://github.com/ory/sdk/tree/master/clients/hydra/typescript

vinckr:
What exactly needs to be done here? 

Change all links in the documentation from {{Project}}/sdk  to ory/sdk/{{Project}} ?

Should the existing documentation also be moved to ory/sdk/{{Project}}; 
or just point the sdk packages to ory/sdk and have the rest of documentation from {{Project}}/docs?

zepatrik:
1. change all the links to point to the right repo
2. update the javascript page to make clear that it is a typescript and javascript sdk, just get inspired by the text from the sdk readme :wink:

aeneasr:
Resolved on master :)

PR #2195, opened 2020-11-24
docs: sdk doc typescript

vinckr:
Fixed some links and first draft of the typescript info page.
It is basically the readme copied, but the commands would have to be changed, right?
Just wanted to know if I am going in the right direction.

#2171

## Checklist

      the changes.
      works.

## Further comments

Maybe the javascript sdk change should be in a different PR to address https://github.com/ory/sdk/issues/22 ?

### DIFF
--- a/docs/docs/sdk/js.md
+++ b/docs/docs/sdk/js.md
@@ -2,12 +2,58 @@
 id: js
 title: JavaScript
 ---
+  
+This generator creates TypeScript/JavaScript client that utilizes [axios](https://github.com/axios/axios). The generated Node module can be used in the following environments:
 
-To install the JavaScript SDK, run:
+Environment
+* Node.js
+* Webpack
+* Browserify
+
+Language level
+* ES5 - you must have a Promises/A+ library installed
+* ES6
+
+Module system
+* CommonJS
+* ES6 module system
+
+It can be used in both TypeScript and JavaScript. In TypeScript, the definition should be automatically resolved via `package.json`. ([Reference](http://www.typescriptlang.org/docs/handbook/typings-for-npm-packages.html))
+
+### Building
+
+To build and compile the typescript sources to javascript use:
+```
+npm install
+npm run build
+```
+
+### Publishing
+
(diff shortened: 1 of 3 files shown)


aeneasr (approved):
~ignore-me~

docs/docs/sdk/js.md
```diff
@@ -3,12 +3,67 @@ id: js
...
+Module system
+
+- CommonJS
+- ES6 module system
+
+It can be used in both TypeScript and JavaScript. In TypeScript, the definition
+should be automatically resolved via `package.json`.
+([Reference](http://www.typescriptlang.org/docs/handbook/typings-for-npm-packages.html))
+
+### Building
```
aeneasr, line 30: It is not clear where this should be used. Did you simply copy the README? How we build and publish the SDK is not of relevance here imo - we should just show how people can use the SDK.

docs/docs/sdk/js.md
```diff
@@ -3,12 +3,67 @@ id: js
...
+### Consuming
+
+navigate to the folder of your consuming project and run one of the following
+commands.
+
+_published:_
 
 ```
-npm install --save @oryd/hydra-client
+npm install @oryd/hydra-client@v1.9.0-alpha.2 --save
```
aeneasr, line 51: If this is added, you need to update this tag on every new release. This can be automated using something like this: https://github.com/ory/hydra/blob/master/docs/config.js#L11-L16

aeneasr (changes requested):

docs/docs/sdk/js.md
vinckr: Makes sense! 
Is there documentation for that or is the old documentation still correct?
Related: https://github.com/ory/sdk/issues/22

If there is no documentation yet, what would be the best way to correct this?

aeneasr: I don't think it is still correct. We use the SDK in https://github.com/ory/kratos-selfservice-ui-node

PR #2264, opened 2020-12-21
Fix typo

bfrisbie-brex:


docs/versioned_docs/version-v1.9/concepts/oauth2.mdx
```diff
@@ -184,7 +184,7 @@ Will not issue an OAuth 2.0 Refresh Token:
 
 - https://ory-hydra.example/oauth2/auth?response_type=token&...
 - https://ory-hydra.example/oauth2/auth?response_type=token+id_token&...
-- https://ory-hydra.example/oauth2/token?grant_type=client_redentials&...
+- https://ory-hydra.example/oauth2/token?grant_type=client_credentials&...
```
aeneasr, line 187: @vinckr could you add this to the root docs in another pr as well please?

aeneasr (changes requested):
Thank you for these changes! The documentation is a bit tricky at the moment as the change has to be made in two places, the file you edited and: https://github.com/ory/hydra/blob/05897d3e9f6ee1a2231eaa27b78f52ee7faf4d27/docs/docs/concepts/oauth2.mdx

bfrisbie-brex:
@aeneasr fixed in the 2nd spot too; thanks for pointing that out.

aeneasr:
Awesome 🎉 

PR #2349, opened 2021-02-09
fix: Revert mode default and maximum values

arunas-ka:
I made a mistake in previous pull request, these socket mode values are in decimal, not octal format. Sorry.


aeneasr:
All good!

Issue #2338, opened 2021-02-04
Improve error message for SQL connection failure


nyvelius:
The `msg` when failing to connect to a SQL database could be improved. The `error` accurately informs us that the connection was refused, but the message is currently misleading.

Example:
```
time=2021-02-04T14:07:38Z 
level=fatal 
msg=Could not ensure that signing keys for "hydra.openid.id-token" exists. If you are running against a persistent SQL database this is most likely because your "secrets.system" ("SECRETS_SYSTEM" environment variable) is not set or changed. When running with an SQL database backend you need to make sure that the secret is set and stays the same, unless when doing key rotation. This may also happen when you forget to run "hydra migrate sql".. 
audience=application 
error=map[message:failed to connect to `host=IP user=USER database=DB`: dial error (dial tcp IP:5432: connect: connection refused)] service_name=ORY Hydra service_version=v1.9.2
```

aeneasr:
Contributions welcomed!

lukestoward:
^ I've had a go at categorising the error to avoid the misleading message. This is very generalised though and is only a small QoL improvement.

PR #2367, opened 2021-02-20
fix: Add network specific error message to avoid confusion

lukestoward:
## Related issue

#2338

## Proposed changes

In the event of a general network error, avoid logging a misleading/confusing message.

## Checklist

      the changes.
      works.

### DIFF
--- a/driver/registry_base.go
+++ b/driver/registry_base.go
@@ -2,6 +2,8 @@ package driver
 
 import (
 	"context"
+	"errors"
+	"net"
 	"net/http"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


driver/registry_base.go
```diff
@@ -316,6 +318,11 @@ func (m *RegistryBase) ScopeStrategy() fosite.ScopeStrategy {
 
 func (m *RegistryBase) newKeyStrategy(key string) (s jwk.JWTStrategy) {
 	if err := jwk.EnsureAsymmetricKeypairExists(context.Background(), m.r, new(jwk.RS256Generator), key); err != nil {
+		var netError net.Error
+		if errors.As(err, &netError) {
```
aeneasr, line 322: Shouldn't this be `**netError`?

aeneasr (commented):

driver/registry_base.go
lukestoward: In this case I don't believe so, as `net.Error` is an interface. It worked when I tested it locally.

aeneasr: Right, that makes sense! Maybe we could add a small test to ensure we don't regress on this in future iterations?

lukestoward: Sure, I'll see what I can do.

aeneasr: Awesome, thank you!

lukestoward: @aeneasr i'm having trouble writing tests for this due to the fact that we call `.Fatalf` as part of the error logging, which inevitably leads to `os.Exit(1)`. Do have any advice/examples on how I might write tests for this? Thanks.

aeneasr: It's ok, you can call return explicitly here!

aeneasr (approved):

PR #2381, opened 2021-03-04
feat: flush inactive/expired login and consent requests

Benehiko:
## Related issue

This is related to the `hydra_oauth2_authentication_request` table growth https://github.com/ory/hydra/issues/1574 caused by expired/inactive login and consent flows never being purged from the database.

## Proposed changes

Delete records from `hydra_oauth2_authentication_request` and `hydra_oauth2_consent_request` table which are of no use anymore causing the database growth to occur. This would then still follow the `notAfter` safe-guard approach to ensure we don't delete records the user does not want deleted.

~~Add this feature to the `oauth2/flush` endpoint https://github.com/Benehiko/hydra/blob/7f64e1a503646a3947af5aef9428dbc1b2792a3a/oauth2/handler.go#L520~~

~~This could also possibly be added to its own endpoint under the login/consent flows `oauth2/auth/flush` since this has to do specifically with login and consent flows rather than oauth2 tokens. 
https://github.com/Benehiko/hydra/blob/7f64e1a503646a3947af5aef9428dbc1b2792a3a/consent/handler.go#L797~~

Add this to a `janitor` command in the CLI which will be a replacement for the current `oauth2/flush` endpoint. 

## Checklist

      the changes.
      works.

## Further comments

Related test run to verify the growth of tables on Hydra:

**Before change**

This graph shows the growth over time after 200 000 clients have been created and each have tried the login / consent flows.
The failure-rate (number of clients that fail to log in) were set to be 97% which did not make a significant change in the growth over time against a test of successful client logins. The growth in clients created can be ignored as the tests were creating a singular client per login (`hydra_client` table - yellow). The `hydra_oauth2_authentication_request` table (purple) has been identified as being a problem table that would grow over time as clients perform the login and consent flows.

flush-310_ttl-97_failures-100_cycles-2k_clients-pg

**After change**

This graph shows the growth over time after 20 000 clients have been created and each have tried the login / consent flows.
The failure-rate (number of clients that fail to log in) were set to be 97% which did not make a significant change in the growth over time against a test of successful client logins. The growth in clients created can be ignored as the tests were creating a singular client per login (`hydra_client` table - red). The `hydra_oauth2_authentication_request` table (light-blue) has been identified as being a problem table that would grow over time as clients perform the login and consent flows.

flush-60_ttl-97_failures-10_cycles-2k_clients-pg


zepatrik:
Can you add the charts you had? Just the discussion here, not in git
Do you have some with the flushing applied as in this PR?

x/fosite_storer.go
```diff
@@ -42,5 +42,9 @@ type FositeStorer interface {
 
 	FlushInactiveAccessTokens(ctx context.Context, notAfter time.Time) error
 
+	// flush the consent/login requests from the database.
+	// this will address the database long-term growth issues.
```
zepatrik, line 46:
```suggestion
	// Flush the consent/login requests from the database.
	// This will address the database long-term growth issues discussed in https://github.com/ory/hydra/issues/1574.
```

Benehiko:
> Can you add the charts you had? Just the discussion here, not in git
> Do you have some with the flushing applied as in this PR?

Can do, I do have one

persistence/sql/persister_consent.go
```diff
@@ -438,3 +438,17 @@ func (p *Persister) VerifyAndInvalidateLogoutRequest(ctx context.Context, verifi
...
+
+func (p *Persister) FlushInactiveLoginConsentRequests(ctx context.Context, notAfter time.Time) error {
+	var lr consent.LoginRequest
+	err := p.Connection(ctx).RawQuery(
+		fmt.Sprintf("DELETE FROM %s WHERE requested_at < ? AND requested_at < ?", (&lr).TableName()),
+		time.Now().Add(-p.config.ConsentRequestMaxAge()),
+		notAfter,
+	).Exec()
+
+	if err == sql.ErrNoRows {
```
zepatrik, line 450:
```suggestion
	if errors.Is(sql.ErrNoRows) {
```

persistence/sql/persister_consent.go
```diff
@@ -438,3 +438,17 @@ func (p *Persister) VerifyAndInvalidateLogoutRequest(ctx context.Context, verifi
...
+		return errors.Wrap(fosite.ErrNotFound, "")
```
zepatrik, line 451:
```suggestion
		return errors.WithStack(fosite.ErrNotFound)
```

zepatrik: Actually, as this is not called by fosite, you don't have to return a fosite error here. Just returning `sqlcon.HandleError(err)` should be fine IMO.

zepatrik (changes requested):

persistence/sql/persister_consent.go
```diff
@@ -438,3 +438,14 @@ func (p *Persister) VerifyAndInvalidateLogoutRequest(ctx context.Context, verifi
 		return nil
 	})
 }
+
+func (p *Persister) FlushInactiveLoginConsentRequests(ctx context.Context, notAfter time.Time) error {
+	var lr consent.LoginRequest
+	err := p.Connection(ctx).RawQuery(
+		fmt.Sprintf("DELETE FROM %s WHERE requested_at < ? AND requested_at < ?", (&lr).TableName()),
```
aeneasr, line 445: I don't think this will do what you want - it removes stuff from the login request table which is older than the request max age. The problem however is that we want to keep this stuff around for successful flows - otherwise all tokens associated with this request will be invalidated! So we probably need to use some JOIN magic to achieve that!

aeneasr (commented):
Thank you, I think this is going in the right direction! I actually have another proposal - we currently have the oauth2 flush endpoint, but I think that's actually a bad pattern. We should probably have a subcommand `hydra janitor` (or `hydra clean`) which does what the flush endpoint does today. It would be similar to `hydra migrate sql` and connect directly to the database. We could then have flags (e.g. `--keep-if-younger 30d`) which would configure this stuff.

Its a bit of refactoring but since we tackle this anyways in this PR I think it makes sense. If you need some guidance I am happy to help with pointers!

persistence/sql/persister_consent.go
Benehiko: Oh i see - so this table also gets checked for the token validity?
So basically when removing from here - also check that the token has expired or has no session?

aeneasr: You're actually right, I think it is ok to remove entries from this table. I thought it might have cascading effects but as far as I can tell it might not. The only cascading delete I could find was on [`hydra_oauth2_authentication_request_handled`](https://github.com/ory/hydra/blob/9a81fbcf804b199da0f02c9f07e1adf42d418927/internal/fizzmigrate/consent/migrations/sql/postgres/7.sql#L43) which *should* be ok.

This will however not work for the consent table as removing stuff from the consent table that is valid will break a lot of flows.

However, I think that this approach is still not the best one. Instead, you should delete only rows that have failed flows. This means flows which:

- Do not have a corresponding `*_handled` row which means that the flow never completed
- Have a `*_handled` flow which does not have the status of accepted

persistence/sql/persister_consent.go
```diff
@@ -438,3 +438,92 @@ func (p *Persister) VerifyAndInvalidateLogoutRequest(ctx context.Context, verifi
...
+	// Using NOT EXISTS instead of LEFT JOIN or NOT IN due to
+	// LEFT JOIN not supported by Postgres and NOT IN will have performance hits with large tables.
+	// https://stackoverflow.com/questions/19363481/select-rows-which-are-not-present-in-other-table/19364694#19364694
+	// Cannot use table aliasing in MYSQL, will work in Postgresql though...
+	err := p.Connection(ctx).RawQuery(fmt.Sprintf(`
+	DELETE
+	FROM %s
+	WHERE NOT EXISTS
+		(
+		SELECT NULL
```
aeneasr, line 467: Is `NULL == NOT EXISTS` or `NULL == EXISTS`?

persistence/sql/persister_consent.go
```diff
@@ -438,3 +438,92 @@ func (p *Persister) VerifyAndInvalidateLogoutRequest(ctx context.Context, verifi
...
+	// - hydra_oauth2_consent_request_handled
+	// AND
+	// - hydra_oauth2_authentication_request.requested_at < ttl.login_consent_request
+	// - hydra_oauth2_authentication_request.requested_at < notAfter
+
+	// Using NOT EXISTS instead of LEFT JOIN or NOT IN due to
+	// LEFT JOIN not supported by Postgres and NOT IN will have performance hits with large tables.
+	// https://stackoverflow.com/questions/19363481/select-rows-which-are-not-present-in-other-table/19364694#19364694
+	// Cannot use table aliasing in MYSQL, will work in Postgresql though...
+	err := p.Connection(ctx).RawQuery(fmt.Sprintf(`
```
aeneasr, line 462: Both queries should be done in a transaction so that we roll back if an error occurrs and don't end up with a broken state!

aeneasr (changes requested):

cmd/janitor.go
```diff
@@ -0,0 +1,29 @@
...
+	### Warning ###
+
+	This is a destructive command and will purge data directly from the database.
+	Please use this command with caution if you need to keep historic data for any reason.`,
+	Run: cmdHandler.Janitor.Purge,
+}
+
+func init() {
+	RootCmd.AddCommand(janitorCmd)
+	janitorCmd.Flags().StringP("keep-if-younger", "k", "", "Keep database records that are younger than a specified duration e.g. 1s, 1m, 1h.")
```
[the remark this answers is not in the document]
Benehiko: Do you mean this in light of the `consent-request-lifespan`, `refresh-lifespan` and `access-lifespan` flags?

persistence/sql/persister_consent.go
Benehiko: So wrap the calls in `p.transaction(ctx, func(ctx context.Context, c *pop.Connection) error {}` will execute it in the transaction?

persistence/sql/persister_consent.go
Benehiko: The `Select NULL` part is irrelevant and just there as a place holder for MYSQL - https://dev.mysql.com/doc/refman/8.0/en/exists-and-not-exists-subqueries.html. 

In PostgreSQL does not need anything in the Select to work. See this SO thread https://stackoverflow.com/questions/19363481/select-rows-which-are-not-present-in-other-table/19364694#19364694.

persistence/sql/persister_consent.go
zepatrik: Yes, and it will also continue a transaction that is already running and stored in the context. So you can safely call other functions within the persister as well.

cmd/janitor.go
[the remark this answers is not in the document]
aeneasr: yes, all of them!

aeneasr:
@Benehiko good for another 👀 ?

Benehiko:
> @Benehiko good for another  ?

I still need to write tests for the persister:
https://github.com/ory/hydra/pull/2381#discussion_r593048173

And when handler removes login/consent requests
https://github.com/ory/hydra/pull/2381#discussion_r593046711

Other than that most of the small fixes have been done and 2/3 dockertests have been implemented alongside the 3/3 sqlite tests for the cli.

oauth2/handler.go
```diff
@@ -517,6 +517,11 @@ func (h *Handler) FlushHandler(w http.ResponseWriter, r *http.Request, _ httprou
 		return
 	}
 
+	if err := h.r.OAuth2Storage().FlushInactiveLoginConsentRequests(r.Context(), fr.NotAfter); err != nil {
```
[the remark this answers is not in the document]
Benehiko: Shouldn't I remove this endpoint completely due to it being a CLI command now?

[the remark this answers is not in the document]
aeneasr: Not yet, we'll remove it in the release after

cmd/janitor.go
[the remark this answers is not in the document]
Benehiko: The `consent-request-lifespan` will clash with config "c" flag so will have to make it shorthand `o`.

Shouldn't I rather make this a slice of string rather and get rid of all the flags? So e.g. `janitorCmd.Flags().StringSliceP("keep-if-younger", "k", []string{}, "Keep database records that are younger than a specified duration of options [notAfter, access-lifespan, refresh-lifespan, consent-request-lifespan] e.g. 1s, 1m, 1h, 2h.")`

This will result in k[0] =1s, k[1]=1m, k[2]=1h and k[3]=2h. Might be a bit cleaner in terms of flags?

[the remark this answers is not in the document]
aeneasr: No, this is a very confusing API to use where it's easy to make typos!

The `-o` flag is also reserved for output and should not have other meanings, similar as `-f` means force. Maybe it's best to not use shorthand flags here at all and instead require the long arguments!

persistence/sql/persister_consent.go
```diff
@@ -438,3 +438,92 @@ func (p *Persister) VerifyAndInvalidateLogoutRequest(ctx context.Context, verifi
 		return nil
 	})
 }
+
+func (p *Persister) FlushInactiveLoginConsentRequests(ctx context.Context, notAfter time.Time) error {
```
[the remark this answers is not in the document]
Benehiko: So all the dockertests happen in one place via `persister_test.go`? I have implemented the dockertests + sqlite inside the `handler_janitor_test.go` file (https://github.com/Benehiko/hydra/blob/25a12c384f06785a3765cb361585a256aef28d46/cmd/cli/handler_janitor_test.go#L233)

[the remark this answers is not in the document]
Benehiko: So it would be something similar to this test here - https://github.com/ory/hydra/blob/a8ca333bb65fe591d1795ce5b33760c97ff54d65/oauth2/fosite_store_helpers.go#L177-L176

So should I then move all the tests to helper methods and call them in `manager_test_helper.go`
https://github.com/ory/hydra/blob/e17a074774eafc93f4cb5620f2f186e87a9ec96e/consent/manager_test_helpers.go#L261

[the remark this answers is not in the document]
aeneasr: Yup

cmd/cli/handler_janitor.go
```diff
@@ -0,0 +1,129 @@
...
+	conn := p.Connection(cmd.Context())
+
+	if conn == nil {
+		return fmt.Errorf("%s\n%s\n", cmd.UsageString(),
+			"Janitor can only be executed against a SQL-compatible driver but DSN is not a SQL source.")
+	}
+
+	if err := conn.Open(); err != nil {
+		return errors.Wrap(errorsx.WithStack(err), "Could not open the database connection")
+	}
```
aeneasr, line 85: Do p.Init() instead here, then you have access to an alive and well db connection pool! p.Init() will also retry the DB if the connection is not yet up

cmd/cli/handler_janitor.go
```diff
@@ -0,0 +1,129 @@
...
+
+	if onlyTokens && !onlyRequests {
+		routines = append(routines, cleanup(p.FlushInactiveAccessTokens, "access tokens"))
+		routines = append(routines, cleanup(p.FlushInactiveRefreshTokens, "refresh tokens"))
+	}
+
+	if onlyRequests && !onlyTokens {
+		routines = append(routines, cleanup(p.FlushInactiveLoginConsentRequests, "login-consent requests"))
+	}
```
aeneasr, line 105: I think we should change how the janitor works here and have flags `--tokens` and `--requests` as arguments. We then do:

```
if (cmdx.MustGetBool(cmd, "tokens") {
		routines = append(routines, cleanup(p.FlushInactiveAccessTokens, "access tokens"))
 		routines = append(routines, cleanup(p.FlushInactiveRefreshTokens, "refresh tokens"))
}
```

If both flags are unset, we can throw an error telling the user to specify at least one of `tokens` or `requests`

driver/registry_sql.go
```diff
@@ -81,7 +81,13 @@ func (m *RegistrySQL) Init(ctx context.Context) error {
...
 
 		// if dsn is memory we have to run the migrations on every start
-		if m.C.DSN() == dbal.SQLiteInMemory {
+		// TODO: might need something better for this since sqlite in-memory can be written in different styles depending on the
+		// use case - such as
+		// - just in memory
+		// - shared connection
+		// - shared but unique in the same process
+		// see: https://sqlite.org/inmemorydb.html
+		if m.C.DSN() == dbal.SQLiteInMemory || m.C.DSN() == dbal.SQLiteSharedInMemory || strings.Contains(m.C.DSN(), "mode=memory") {
```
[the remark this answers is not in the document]
aeneasr: You can update this now! You can also use custom go modules (e.g. dependencies not yet released) by doing `go get github.com/foo/bar@git-hash`. You can even use forks or local directories with the `replace` directive

persistence/sql/persister_test.go
```diff
@@ -1,34 +1,33 @@
...
-		"memory": internal.NewRegistryMemory(t, conf),
+		"memory": internal.NewRegistryMemory(t, internal.NewConfigurationWithDefaults()),
 	}
 
 	if !testing.Short() {
 		registries["postgres"], registries["mysql"], registries["cockroach"], _ = internal.ConnectDatabases(t)
 	}
 
 	for k, m := range registries {
+		/*m.Config().MustSet(config.KeyAccessTokenLifespan, time.Hour)
```
aeneasr, line 27: Clean up

aeneasr (commented):
Alano, seriously, good job! This looks great and uses all the new programming paradigms we have in Ory. I will need to give this another in-depth look but the code style is on the absolute right way :)

cmd/cli/handler_janitor.go
Benehiko: it seems `Persister` doesn't have an `Init` but `Registry` does.

So is the flow then something like:
```go

d.Init(cmd.Context()
p := d.Persister()

```

aeneasr:
Tests are failing ;)

cmd/cli/handler_janitor.go
```diff
@@ -0,0 +1,114 @@
...
+func (j *JanitorHandler) Purge(cmd *cobra.Command, args []string) error {
+	var d driver.Registry
+
+	co := []configx.OptionModifier{
+		configx.WithFlags(cmd.Flags()),
+		configx.SkipValidation(),
+	}
+
+	keys := map[string]string{
+		"access-lifespan":          config.KeyAccessTokenLifespan,
```
aeneasr, line 35: Shouldn't these be using the constants?

cmd/cli/handler_janitor.go
```diff
@@ -0,0 +1,114 @@
...
+		if x := flagx.MustGetString(cmd, k); x != "" {
+			if xp, err := time.ParseDuration(x); err == nil {
+				co = append(co, configx.WithValue(v, xp))
+			}
+		}
+	}
+
+	notAfter := time.Now()
+
+	if keepYounger := flagx.MustGetString(cmd, "keep-if-younger"); keepYounger != "" {
```
aeneasr, line 50: Use constant

cmd/cli/handler_janitor.go
```diff
@@ -0,0 +1,114 @@
...
+		if keepYoungerDuration, err := time.ParseDuration(keepYounger); err == nil {
+			notAfter = notAfter.Add(-keepYoungerDuration)
+		}
+	}
+
+	if !flagx.MustGetBool(cmd, "read-from-env") && len(flagx.MustGetStringSlice(cmd, "config")) == 0 {
```
aeneasr, line 56: Use constant for `read-from-env`

cmd/cli/handler_janitor_test.go
```diff
@@ -0,0 +1,247 @@
...
+				len(flagx.MustGetStringSlice(cmd, "config")) == 0 {
+
+				fmt.Printf("%s\n", cmd.UsageString())
+				return fmt.Errorf("%s\n%s\n%s\n",
+					"A DSN is required as a positional argument when not passing any of the following flags:",
+					"- Using the environment variable with flag -e, --read-from-env",
+					"- Using the config file with flag -c, --config")
+			}
+
+			if (!flagx.MustGetBool(cmd, onlyTokens) && !flagx.MustGetBool(cmd, onlyRequests)) || (flagx.MustGetBool(cmd, onlyTokens) && flagx.MustGetBool(cmd, onlyRequests)) {
```
aeneasr, line 45: Does this mean we can't use `--tokens --requests` together? If so, why?

Benehiko: I think I misunderstood your review here https://github.com/ory/hydra/pull/2381#discussion_r597475800

I can re-add if both are used together.

aeneasr: Yeah, it should be possible to clean up everything with just one command!

Benehiko:
@aeneasr I have refactored the janitor cli. 
I had to move how the cobra command was being created from the `cmd` package to the `cli` package. So all the flags and their descriptions are created in the `cli` package now. This is because I would've gotten a cyclic dependency error by importing the `cmd` constants inside the `cli` package as well as using the `janitorCmd` inside `handler_janitor_test.go`. This removes the duplication code for setting up the cli tests.

persistence/sql/persister_consent.go
```diff
@@ -438,3 +438,85 @@ func (p *Persister) VerifyAndInvalidateLogoutRequest(ctx context.Context, verifi
...
+				WHERE %[1]s.challenge = %[3]s.login_challenge AND %[4]s.error = '{}'
+				)
+			AND requested_at < ?
+			AND requested_at < ?
+			`,
+			(&lr).TableName(),
+			(&lrh).TableName(),
+			(&cr).TableName(),
+			(&crh).TableName()),
+			time.Now().Add(-p.config.ConsentRequestMaxAge()),
```
aeneasr, line 489: Shouldn't this be login + consent request max age?

aeneasr (commented):
I've got some comments, you can address them beforehand or we do it together in the call!

cmd/cli/handler_janitor_test.go
```diff
@@ -0,0 +1,203 @@
...
+)
+
+func newJanitorCmd() *cobra.Command {
+	return newJanitorHandler().Command()
+}
+
+func TestJanitorHandler_PurgeTokenNotAfter(t *testing.T) {
+	ctx := context.Background()
+	testCycles := testhelpers.NewConsentJanitorTestHelper("").GetNotAfterTestCycles()
+
```
aeneasr, line 24:
```suggestion
require.True(t, len(testCycles) > 0)
```

PR #2383, opened 2021-03-05
feat: Refresh Token Reuse Detection

### LINKED ISSUE #2022: Reuse Detection in Refresh Token Rotation

I am currently using Hydra as Authorization Server in my enterprise applications suite and I am facing with security and user experience concerns in Single Page Applications. Our applications use Auth Code Flow and we currently avoid using refresh tokens in such an insecure environment.
I understand that Hydra implements the Refresh Token Rotation best practice, but in case of a stolen refresh token there is no automatic way to recognise that a refresh token is being reused to invalidate the whole refresh token family associated with that consent session.

I have seen that there are examples of implementations of a reuse detection feature, like [what happens in Auth0](https://auth0.com/blog/securing-single-page-applications-with-refresh-token-rotation/). Are there chances that a similar solution is (or will be) implemented in Hydra too?

Thanks in advance for any feedback on this.

svrakitin:
## Related issue

Closes #2022

## Proposed changes

This PR leverages support for Refresh Token reuse Detection added in https://github.com/ory/fosite/pull/567. It makes hydra's `Persister` stop deleting refresh tokens, but deactivating them similar to whats done for authorization codes.

Modules need to be updated if PR for fosite gets merged.

## Checklist

      the changes.
      works.

## Further comments

See https://github.com/ory/fosite/pull/567 for fosite PR.

### DIFF
--- a/persistence/sql/persister_oauth2.go
+++ b/persistence/sql/persister_oauth2.go
@@ -231,15 +231,15 @@ func (p *Persister) findSessionBySignature(ctx context.Context, rawSignature str
 			return errorsx.WithStack(fosite.ErrNotFound)
 		} else if err != nil {
 			return sqlcon.HandleError(err)
-		} else if !r.Active && table == sqlTableCode {
+		} else if !r.Active {
 			fr, err = r.toRequest(ctx, session, p)
 			if err != nil {
 				return err
-			} else {
+			} else if table == sqlTableCode {
 				return errorsx.WithStack(fosite.ErrInvalidatedAuthorizeCode)
+			} else if table == sqlTableRefresh {
+				return errorsx.WithStack(fosite.ErrInactiveToken)
 			}
-		} else if !r.Active {
-			return errorsx.WithStack(fosite.ErrInactiveToken)
 		}
 
 		fr, err = r.toRequest(ctx, session, p)
@@ -247,7 +247,7 @@ func (p *Persister) findSessionBySignature(ctx context.Context, rawSignature str
 	})
 }
 
-func (p *Persister) deleteSession(ctx context.Context, signature string, table tableName) error {
+func (p *Persister) deleteSessionBySignature(ctx context.Context, signature string, table tableName) error {
 	signature = p.hashSignature(signature, table)
 
 	/* #nosec G201 table is static */
@@ -257,7 +257,7 @@ func (p *Persister) deleteSession(ctx context.Context, signature string, table t
 			Exec())
 }
 
-func (p *Persister) revokeSession(ctx context.Context, id string, table tableName) error {
+func (p *Persister) deleteSessionByRequestID(ctx context.Context, id string, table tableName) error {
 	/* #nosec G201 table is static */
 	if err := p.Connection(ctx).RawQuery(
 		fmt.Sprintf("DELETE FROM %s WHERE request_id=?", OAuth2RequestSQL{Table: table}.TableName()),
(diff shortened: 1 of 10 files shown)


svrakitin:
@aeneasr No idea what happened to e2e tests, but should be good to go, let me know if I forgot to cover something.

persistence/sql/persister_oauth2.go
```diff
@@ -231,23 +231,23 @@ func (p *Persister) findSessionBySignature(ctx context.Context, rawSignature str
...
 			if err != nil {
 				return err
-			} else {
+			} else if table == sqlTableCode {
 				return errorsx.WithStack(fosite.ErrInvalidatedAuthorizeCode)
+			} else if table == sqlTableRefresh {
+				return errorsx.WithStack(fosite.ErrInactiveToken)
 			}
-		} else if !r.Active {
-			return errorsx.WithStack(fosite.ErrInactiveToken)
```
aeneasr, line 242: This introduces a serious issue as it is not covering the case where `r.Active` is false and table is none of the options. Please add a test for this!

aeneasr (changes requested):
Sorry, I had reviewed 90% of the PR and for some reason stopped in the middle of it, which is why I didn't complete it.

> @aeneasr No idea what happened to e2e tests, but should be good to go, let me know if I forgot to cover something.

It might make sense to add a full [e2e test for refresh tokens](https://github.com/ory/hydra/blob/master/cypress/integration/oauth2/refresh_token.js). Writing and testing these is covered here: https://github.com/ory/hydra#e2e-tests

> We also need some garbage collection in place for refresh tokens. I think I've seen some other PR covering this.

Yes, that was merged!

persistence/sql/persister_oauth2.go
svrakitin: What would be preferred behaviour if something is inactive? Here it falls through as before.

aeneasr: Before, if `r.Active` was false, you would get `fosite.ErrInactiveToken`. Now, you won't get that and end up at `r.toRequest`. You need to add a `return errorsx.WithStack(fosite.ErrInactiveToken)` after

```
 			} else if table == sqlTableRefresh {
 				return errorsx.WithStack(fosite.ErrInactiveToken)
 			}
```

svrakitin: Sorry I overlooked that. I don't have to modify this method at all, it was working as required.

svrakitin:
Added e2e test to go through Auth Code Flow with PKCE in the browser and reuse refresh token. This is closer to the real scenario when it can happen.

aeneasr:
> CircleCI got an incident, so e2e tests failed again. Did you consider moving tests to Github Actions? :)

Yeah, unfortunately it's quite an effort to do that so it's gonna take a while as we have lots of tooling around circleci.

Sorry about the flaky e2e test - I will rerun the CI.

Is this good for another 👀 ?

svrakitin:
@aeneasr Yeah, please review.

svrakitin:
@aeneasr Please let me know if any other change is expected.

aeneasr:
Thank you for bubbling this up in my inbox, I must have missed your push! Taking a look now

aeneasr (approved):
Awesome, thank you! 🎉🎉🎉🎉 Your contribution makes Ory better :)

pharman:
Sweet! 👍

Issue #2382, opened 2021-03-05
hydra token user: add ability to not terminate on successful auth flow


verzac:
Either way, this is a nice-to-have, so it's not 100% necessary - just wanted to get the idea out there and looked at just in case a contributor comes along and wants to contribute. Happy to help when I have the time though!

**Is your feature request related to a problem? Please describe.**

When I'm testing the auth flow locally with the custom IdP I'm developing, I'd like for the token client to not close when auth flow is complete, so that I can keep on testing live changes to the auth flow iteratively.

**Describe the solution you'd like**

Make this line here optional: https://github.com/ory/hydra/blob/master/cmd/token_user.go#L238

We can probably do this through a _very optional_ flag called --no-shutdown.

Now, I know it generates a `nonce` and a `state` when it starts up, and I'm not 100% aware if nonces/states can be reused when auth-ing with Hydra (it probably can't and shouldn't), so we _might_ need `hydra token` to reset when `GET /callback` is called.

**Describe alternatives you've considered**

It's just a nice-to-have for lazy bums like me, so the alternative is running `hydra token user` again and again every time I want to test the auth flow.

aeneasr:
Sounds good to me! Would you be up for a contribution? :) `state` and `nonce` should probably be regenerated when we redo the flow :)

PR #2386, opened 2021-03-06
feat: add --no-shutdown flag to "hydra token user" to prevent auto-termination (#2382)

verzac:
## Related issue

#2382 @aeneasr

## Proposed changes

Added `--no-shutdown` flag to `hydra token user`, which signifies that `shutdown()` should not be called. Also added the capability to regenerate state/nonce upon success/error - basically whenever `/callback` is called and whenever shutdown should've occurred if `--no-shutdown` was not provided.

## Checklist

      the changes.
      works.

## Further comments

Can't seem to figure out which docs I should change, as some of them are versioned and some of them aren't. Happy to update the docs if I'm pointed to the right direction!

This change regenerates state/nonce on error too, not just on success. Not sure if we're keen on this one, but I tried to keep the behaviour consistent with "we're terminating" -> "if --no-shutdown is provided, we're restarting the flow".

Also, added a nifty "Back to Welcome Page" button so that people can just click it to go back to the start of the flow.

RE: Tests, I don't think there's any that I should update? Feel free to CMIIW tho, I'm pretty new to the codebase :)


aeneasr (approved):

PR #2399, opened 2021-03-10
fixed basic examples for the golang SDK

cdreier:
the examples were a bit confusing, because the hydra-client was named `admin` - which conflicts with the correct import for the `admin.NewCreateOAuth2ClientParams` call.

i had to battle a bit around, and tried to rename a few variables and added correct imports, so the examples should just work when copy & pasting


aeneasr:

- `docs/docs/path/to/file`

aeneasr:
(sorry, the close me clicking enter too soon)

cdreier:
thank you for clarification :)
i think now all changes are in both locations

aeneasr (approved):

Issue #2385, opened 2021-03-05
feat: Support partial client updates


mbonnell-wish:
**Is your feature request related to a problem? Please describe.**

Currently, the only client updates supported by the REST API are updates with PUT semantics, meaning the entire client gets replaced. This is undesirable when one only wants to update a certain property, like a callback URI. 

This desire is referenced [here](https://github.com/ory/hydra/issues/2020#issuecomment-764668918) and [here](https://github.com/ory/hydra/pull/2049#discussion_r487434394).

**Describe the solution you'd like**

It would be good to also support partial client updates through a PATCH request. For the CLI, we could add some flag like `--partial` as suggested by @matheusfm. 

**Describe alternatives you've considered**

A clear and concise description of any alternative solutions or features you've
considered.

**Additional context**

Would be happy to tackle this over the weekend.

mattbonnell:
@aeneasr let me know if you'd be open to accepting a PR for this feature 😄 
Would be a little bit of work so just want to get the green-light from you before I start on it.

aeneasr:
Yes, that would be great! There's an open issue for discussing ideas how to do this here: https://github.com/ory/meta/issues/2

PR #2411, opened 2021-03-19
feat: implement partial client updates

mattbonnell:
## Related issue

#2385 

## Proposed changes
- Adds a PATCH handler to the client admin API.
- Patch implemented as [JSON Patch](http://jsonpatch.com/) as per https://github.com/ory/meta/issues/2 using https://github.com/evanphx/json-patch

## Checklist

      the changes.
      works.


mattbonnell:
wow, git made such a strange diff :) probably easiest to just look at the `client/handler.go` file directly

mattbonnell:
getting the following error when I try to regenerate the sdk:
```bash
❯ make sdk
GOBIN=/Users/mattbonnell/dev/go/github.com/ory/hydra/.bin/ go install  github.com/ory/cli
# github.com/ory/jsonschema/v3/httploader
../../../../../go/pkg/mod/github.com/ory/jsonschema/v3@v3.0.1/httploader/httploader.go:27:15: undefined: httpx.NewResilientClientLatencyToleranceMedium
# github.com/ory/cli/cmd/dev/newsletter
../../../../../go/pkg/mod/github.com/ory/cli@v0.0.35/cmd/dev/newsletter/pkg.go:86:12: undefined: httpx.NewResilientClientLatencyToleranceMedium
make: *** [.bin/cli] Error 2
```

client/handler.go
```diff
@@ -21,10 +21,14 @@
 package client
 
 import (
+	"context"
 	"encoding/json"
+	"io"
 	"net/http"
 	"time"
 
+	jsonpatch "github.com/evanphx/json-patch"
```
aeneasr, line 30: Is this the best library for the job? What alternatives do we have?

client/handler.go
```diff
@@ -146,25 +151,92 @@ func (h *Handler) Update(w http.ResponseWriter, r *http.Request, ps httprouter.P
...
+	}
+	modified, err := patch.Apply(original)
+	if err != nil {
+		h.r.Writer().WriteError(w, r, err)
+		return
+	}
+	if err := json.Unmarshal(modified, c); err != nil {
 		h.r.Writer().WriteError(w, r, err)
 		return
 	}
```
aeneasr, line 215: I think we could wrap this, including DecodePatch, in a helper function in x and add a few test there? Alternatively we could also add this to `ory/x/decoderx`

aeneasr (commented):
This looks pretty much like what I imagined! We obviously need a few tests and so on :) But I think this is almost good as is

client/handler.go
mattbonnell: The [JSON Patch](http://jsonpatch.com/) site gives two options for Go implementations, this and another library, https://github.com/mattbaird/jsonpatch. Between the two, this one has more starts (551 vs 89) and more contributors (41 vs 8). This one was also updated more recently (January vs last August). Additionally, this library also supports [Merge Patch](https://tools.ietf.org/html/rfc7396), which was indicated as a potential alternative in https://github.com/ory/meta/issues/2. 

For these reasons, I felt this was the better choice between the two.

client/handler.go
mattbonnell: I think that would be a good idea! Will investigate this.

client/handler.go
aeneasr: Perfect, thank you!

client/handler.go
aeneasr: Awesome, thank you :)

mattbonnell: done! defined a helper in `x/json.go`

aeneasr:
> getting the following error when I try to regenerate the sdk:

You can bum ory/cli to a newer version which should fix that!

mattbonnell:
> > getting the following error when I try to regenerate the sdk:
> 
> You can bum ory/cli to a newer version which should fix that!

Hey, can you help me understand the `make sdk` rule? The first command is:
`swagger generate spec -m -o ./spec/api.json -x internal/httpclient -x gopkg.in/square/go-jose.v2`

Isn't `internal/httpclient` the swagger-generated client code? I think I'm missing something here, this seems circular. How can we go about adding definitions to the swagger spec in this case? Additionally, it seems the `sdk` rule doesn't consider changes made in the `client` package.

I had added the `PatchRequest` and `PatchDocument` definitions to `api.json`, but this would overwrite them and then I get some other errors. If I remove this first line, then the rule runs as expected and I end up getting the new generated code in `internal/httpclient` (check the files which this PR adds to `internal/httpclient`).

The PR is failing in CI currently at this sdk generation step.

aeneasr:
I'll look into this now

aeneasr:
`make sdk` works fine for me but fails with:

```
The swagger spec at "./spec/api.json" is invalid against swagger specification 2.0.
See errors below:
- "updateOAuth2Client" is defined 2 times
```

because

```
// swagger:route PATCH /clients/{id} admin updateOAuth2Client
```

and

```
// swagger:route PUT /clients/{id} admin updateOAuth2Client
```

share the operation id. So one should be update and the other maybe patchOAuth2Client?

spec/api.json
```diff
@@ -292,21 +232,51 @@
...
         }
       },
+      "patch": {
+        "description": "Patch an existing OAuth 2.0 Client. If you pass `client_secret` the secret will be updated and returned via the API. This is the only time you will be able to retrieve the client secret, so write it down and keep it safe.\n\nOAuth 2.0 clients are used to perform OAuth 2.0 and OpenID Connect flows. Usually, OAuth 2.0 clients are generated for applications which want to consume your OAuth 2.0 or OpenID Connect capabilities. To manage ORY Hydra, you will need an OAuth 2.0 Client as well. Make sure that this endpoint is well protected and only callable by first-party components.",
+        "consumes": ["application/json"],
+        "produces": ["application/json"],
+        "schemes": ["http", "https"],
+        "tags": ["admin"],
+        "summary": "Patch an OAuth 2.0 Client",
+        "operationId": "patchOAuth2Client",
```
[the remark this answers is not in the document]
mattbonnell: ohhh, i see. i think i was doing things wrong. hang on.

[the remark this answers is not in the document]
mattbonnell: ok. think i've fixed it now.

mattbonnell:
Hm, generate step failing in pipeline with 

```bash

#!/bin/bash -eo pipefail
swagger generate spec -m -o spec/api.json $(printf -- " -x %s" internal/httpclient -x gopkg.in/square/go-jose.v2)
expected argument for flag `-x, --exclude', but got option `-x'

```

aeneasr:
LMK when good for another 👀

mattbonnell:
> LMK when good for another 👀

hey, yes ready, looks like all checks are passing

zepatrik:
I opened #2434 to fix the nancy warning

client/handler.go
```diff
@@ -146,25 +149,76 @@ func (h *Handler) Update(w http.ResponseWriter, r *http.Request, ps httprouter.P
...
-	c.UpdatedAt = time.Now().UTC().Round(time.Second)
-	if err := h.r.ClientManager().UpdateClient(r.Context(), &c); err != nil {
+	id := ps.ByName("id")
+	c, err := h.r.ClientManager().GetConcreteClient(r.Context(), id)
+	if err != nil {
+		h.r.Writer().WriteError(w, r, err)
+		return
+	}
+
+	if err := x.ApplyJSONPatch(patchJSON, c); err != nil {
```
aeneasr, line 194: It shouldn't be allowed to update the ID of the client as that is immutable iirc!

mattbonnell: Good catch, will fix

mattbonnell: fixed.

aeneasr (commented):
Awesome! This looks pretty much good to go, but I think we need another test for the handler which also includes the ID behavior test! Another option would be to add a „deny list“ to the JSON PATCH helper which returns an error if a specific path is updated (e.g. ID). This would make it more general purpose and easier to test :)

mattbonnell:
> Awesome! This looks pretty much good to go, but I think we need another test for the handler which also includes the ID behavior test! Another option would be to add a „deny list“ to the JSON PATCH helper which returns an error if a specific path is updated (e.g. ID). This would make it more general purpose and easier to test :)

done!

client/handler.go
```diff
@@ -21,7 +21,9 @@
 package client
```
aeneasr, line 21: Could you still add a test for this handler please? :)

mattbonnell: @aeneasr done :)

aeneasr (approved):

aeneasr:
Thank you! Appreciate your contributions as always :)

PR #2427, opened 2021-03-25
fix: version schema

zepatrik:
## Related issue

Since #2392 the config schema is not in the same place anymore. Downstream fix: https://github.com/ory/ci/pull/50


aeneasr:
Orb `ory/goreleaser` has been incremented to `ory/goreleaser@0.1.31`.

PR #2448, opened 2021-04-07
docs: guide for merging system.secrets

vinckr:
## Related issue
https://github.com/ory/hydra/discussions/2445

As before, all credit goes to @bnm12. 

## Checklist

      the changes.
      works.


docs/docs/guides/merge-multiple-db-secrets.mdx
```diff
@@ -0,0 +1,64 @@
...
+```
+
+Then we open each of the dump files and manually edit `hydra_clients.pk` be
+entirely set to the value `DEFAULT`
+([SQL keyword](https://www.w3schools.com/sql/sql_ref_default.asp)) to make it
+follow the sequence in the target DB correctly.
+
+We also delete the last line of the dump, since it sets the value of the primary
+key sequence to the number of the donor database. So we need to delete this
+line.
```
bnm12, line 53: If there's any way of putting emphasis on this that would be good, it's very important.
This whole manual edits section is very important, but the first part will most likely just fail importing and not cause any harm if not done correctly, and can be recovered with a correct import afterwards (with a few harmless duplicate key errors). Where as if you forget to remove this line it will reset the sequence of `hydra_clients.pk` to whatever was in the source DB potentially breaking client creation going forwards until the sequence gets back in sync

vinckr: Thanks for the input! 
I put in some warnings, to make clear that this can potentially break client creation if you mess it up.

aeneasr (approved):

PR #2454, opened 2021-04-09
fix: Fixed WWW-Authenticate header in userinfo handler

mitar:
## Proposed changes

Fixing the header value based on the spec.

## Checklist

      the changes.
      works.

### DIFF
--- a/oauth2/handler.go
+++ b/oauth2/handler.go
@@ -263,6 +263,10 @@ func (h *Handler) WellKnownHandler(w http.ResponseWriter, r *http.Request) {
 //
 // For more information please [refer to the spec](http://openid.net/specs/openid-connect-core-1_0.html#UserInfo).
 //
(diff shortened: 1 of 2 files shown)


aeneasr:
Could you please add a test case and also link the place in the RFC relevant to this?

aeneasr:
Are you still up for the changes? :) If you need any help, let us know!

mitar:
I documented the change. But as explained in #570, I do not know how or where to write a test for this.

aeneasr:
The easiest way to find tests is to search for a unique string like "WWW-Authenticate". This will drop you at e.g. [here](https://github.com/ory/hydra/blob/master/oauth2/handler_test.go#L187-L195) which you could copy to add a test case for what you have changed :)

mitar:
I updated tests.

aeneasr (approved):
Thank you! :))

PR #2455, opened 2021-04-10
fix: Audience should include client ID

mitar:
## Proposed changes

[The spec](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo) says:

> The aud value SHOULD be or include the RP's Client ID value. 

So this makes sure client ID is always there. It is true though that that section talks only about the situation when response is signed.

## Checklist

      the changes.
      works.

### DIFF
--- a/oauth2/handler.go
+++ b/oauth2/handler.go
@@ -308,9 +308,22 @@ func (h *Handler) UserinfoHandler(w http.ResponseWriter, r *http.Request) {
 	delete(interim, "sid")
 	delete(interim, "jti")
 
-	if aud, ok := interim["aud"].([]string); !ok || len(aud) == 0 {
-		interim["aud"] = []string{c.GetID()}
+	aud, ok := interim["aud"].([]string)
+	if !ok || len(aud) == 0 {
+		aud = []string{c.GetID()}
+	} else {
+		found := false
+		for _, a := range aud {
+			if a == c.GetID() {
+				found = true
+				break
+			}
+		}
+		if !found {
+			aud = append(aud, c.GetID())
+		}
 	}
+	interim["aud"] = aud
 
 	if c.UserinfoSignedResponseAlg == "RS256" {
 		interim["jti"] = uuid.New()
--- a/oauth2/handler_test.go
+++ b/oauth2/handler_test.go
@@ -223,36 +223,9 @@ func TestUserinfo(t *testing.T) {
 
 						return fosite.AccessToken, &fosite.AccessRequest{
 							Request: fosite.Request{
-								Client:  &client.Client{},
-								Session: session,
-							},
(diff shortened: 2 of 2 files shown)


oauth2/handler.go
```diff
@@ -308,9 +308,19 @@ func (h *Handler) UserinfoHandler(w http.ResponseWriter, r *http.Request) {
 	delete(interim, "sid")
 	delete(interim, "jti")
 
-	if aud := interim["aud"].([]string); !ok || len(aud) == 0 {
+	if aud, ok := interim["aud"].([]string); !ok || len(aud) == 0 {
```
mitar, line 311: This was a bug, BTW.

aeneasr:
Could you please add a test for this? Thank you!

oauth2/handler.go
```diff
@@ -308,9 +308,19 @@ func (h *Handler) UserinfoHandler(w http.ResponseWriter, r *http.Request) {
...
 		interim["aud"] = []string{c.GetID()}
 	}
+	found := false
+	for _, a := range interim["aud"].([]string) {
+		if a == c.GetID() {
+			found = true
+			break
+		}
+	}
+	if !found {
```
aeneasr, line 321: It would probably make sense to type assert once and set the value on interim later. You might use ory/x/stringslice for convenience!

aeneasr (commented):
LGTM! Could you please add a test for this? :)

aeneasr:
Are you still up for the changes? :) If you need any help, let us know!

aeneasr:
Converting to draft due to lack of response

oauth2/handler.go
mitar: Updated. I didn't find an opportunity to use ory/x/stringslice though.

mitar:
I could not find any existing tests for this handler, so I do not know how to write a test for this. If somebody can help with this. Otherwise I think the PR is ready.

aeneasr:
Basically you could copy&paste [this test case](https://github.com/ory/hydra/blob/master/oauth2/handler_test.go#L209-L235) two or three times and cover the different audience format scenarios here.

As a maintainer I have been bitten so often by contributions without proper tests - from bugs to security issues to regressions. That's why we're also going to change our contribution guidelines which demand tests and might even introduce some coverage tests that fail to pass if test coverage is reduced with contributions.

So unfortunately I can't merge this until proper tests have been added.

mitar:
I completely understand about tests. Thanks for pointing them out.

mitar:
I added tests. But I could not run them locally (see https://github.com/ory/hydra/issues/2532).

aeneasr (approved):
Thank you @mitar :)

PR #2459, opened 2021-04-12
feat: force kid with public prefix

naveenpaul1:
## Related issue

https://github.com/ory/hydra/discussions/2435 
@aeneasr  @Benehiko 

## Proposed changes

remove prefix public from key which is exposed in  .well-known/jwks.json

## Checklist

      the changes.
- [] I have added tests that prove my fix is effective or that my feature
      works.
- [] I have added or changed [the documentation](docs/docs).

@aeneasr


aeneasr:
I think this could be breaking some stuff - I have to look into this but currently am not able due to time constraints...

jwk/helper.go
```diff
@@ -120,6 +128,27 @@ func FindKeyByPrefix(set *jose.JSONWebKeySet, prefix string) (key *jose.JSONWebK
...
+	}
+
+	return First(keys.Keys), nil
+}
+
+func ExcludeKeysByPrefix(set *jose.JSONWebKeySet, prefix string) *jose.JSONWebKeySet {
+	keys := new(jose.JSONWebKeySet)
+
+	for _, k := range set.Keys {
+		if !strings.HasPrefix(k.KeyID, prefix+":") {
```
aeneasr, line 144: We can actually use type assertion here on jose.JSONWebKey to check if the thing is a private key or not (like here https://github.com/ory/hydra/pull/2459/files#diff-d2d06a70d7e57e038f22beffe343a39d7efa66e3be691e00342697c5b7a909c9R125 ). I think that would be even better than doing this prefix stuff here.

aeneasr (changes requested):
Sorry for taking ages on reviewing this. There's a couple of things we should address:

1. We could use type assertions to filter public keys instead
2. Need to add tests

jwk/helper.go
naveenpaul1: @aeneasr : will do the changes

naveenpaul1:
@aeneasr : Code changes are done, could you please review the same

aeneasr (approved):

aeneasr:
Yup, waiting for CI to pass!

jwk/helper.go
```diff
@@ -120,6 +127,25 @@ func FindKeyByPrefix(set *jose.JSONWebKeySet, prefix string) (key *jose.JSONWebK
...
+func ExcludePrivateKeys(set *jose.JSONWebKeySet) *jose.JSONWebKeySet {
+	keys := new(jose.JSONWebKeySet)
+	for _, k := range set.Keys {
+		if _, ok := k.Key.(*rsa.PrivateKey); !ok {
+			keys.Keys = append(keys.Keys, k)
+		}
+	}
+	return keys
+}
+
```
aeneasr, line 148: Could you please add a test for these two functions which:

1. Generate an RSA private/public keypair
2. Generate an ECDSA private/public keypair
3. Run `FindPublicKey` and `ExcludePrivateKeys`
4. Expect no private keys to be exposed, but public keys to be exposed

jwk/handler_test.go
```diff
@@ -41,36 +42,58 @@ import (
...
+		require.NoError(t, reg.KeyManager().AddKeySet(context.TODO(), x.OpenIDConnectKeyName, IDKS))
+		res, err := http.Get(testServer.URL + JWKPath)
+		require.NoError(t, err, "problem in http request")
+		defer res.Body.Close()
+
+		var known jose.JSONWebKeySet
+		err = json.NewDecoder(res.Body).Decode(&known)
+		require.NoError(t, err, "problem in decoding response")
+
+		require.Len(t, known.Keys, 2)
```
aeneasr, line 90: Shouldn't this be `1`? The first key is a private key

jwk/helper.go
```diff
@@ -120,6 +127,25 @@ func FindKeyByPrefix(set *jose.JSONWebKeySet, prefix string) (key *jose.JSONWebK
...
+		return nil, errors.New("key not found")
+	}
+
+	return First(keys.Keys), nil
+}
+
+func ExcludePrivateKeys(set *jose.JSONWebKeySet) *jose.JSONWebKeySet {
+	keys := new(jose.JSONWebKeySet)
+	for _, k := range set.Keys {
+		if _, ok := k.Key.(*rsa.PrivateKey); !ok {
```
aeneasr, line 142: This only works for RSA keys, but what if we use ECDSA? This should not exclude private keys, it should only include "allow listed" public keys.

aeneasr (changes requested):

jwk/helper.go
naveenpaul1: Let me check that @aeneasr

aeneasr:

Thank you!

jwk/helper_test.go
```diff
@@ -71,3 +71,22 @@ func TestIder(t *testing.T) {
 	assert.True(t, len(Ider("public", "")) > len("public:"))
 	assert.Equal(t, "public:foo", Ider("public", "foo"))
 }
+
+func TestHandlerFindPublicKey(t *testing.T) {
```
TimoDoernenburg, line 75: created a new test case to find public RSA and ECDSA keys.

jwk/helper.go
```diff
@@ -139,7 +139,8 @@ func FindPublicKey(set *jose.JSONWebKeySet) (key *jose.JSONWebKey, err error) {
 func ExcludePrivateKeys(set *jose.JSONWebKeySet) *jose.JSONWebKeySet {
 	keys := new(jose.JSONWebKeySet)
 	for _, k := range set.Keys {
-		if _, ok := k.Key.(*rsa.PrivateKey); !ok {
```
TimoDoernenburg: Added ecdsa keys to the loop

jwk/helper.go
```diff
@@ -120,6 +127,26 @@ func FindKeyByPrefix(set *jose.JSONWebKeySet, prefix string) (key *jose.JSONWebK
...
+
+func ExcludePrivateKeys(set *jose.JSONWebKeySet) *jose.JSONWebKeySet {
+	keys := new(jose.JSONWebKeySet)
+	for _, k := range set.Keys {
+		_, ecdsaOk := k.Key.(*ecdsa.PrivateKey)
+		if _, ok := k.Key.(*rsa.PrivateKey); !ok && !ecdsaOk {
+			keys.Keys = append(keys.Keys, k)
+		}
+	}
+	return keys
```
aeneasr, line 147: This list should not filter private keys out, it should only keep public keys in :)

aeneasr (changes requested):

jwk/helper.go
TimoDoernenburg: It's a bit unclear for me what I should change in this function. 
Should the function delete the private keys from the existing list instead of copying the public keys into another list?

TimoDoernenburg: Something like this ? 

`func ExcludePrivateKeys(set *jose.JSONWebKeySet) *jose.JSONWebKeySet {
	keys := new(jose.JSONWebKeySet)
	i := 0 // output index
	for _, k := range set.Keys {
		_, ecdsaOk := k.Key.(*ecdsa.PublicKey)
		_, dsaOK := k.Key.(*dsa.PublicKey)
		_, ed25519OK := k.Key.(*ed25519.PublicKey)
		_, rsaOK := k.Key.(*rsa.PublicKey)

		if !(ecdsaOk || dsaOK || ed25519OK || rsaOK) {
			set.Keys[i] = k
			i++
		}
	}
	return keys
}`

aeneasr: Exactly! Sorry for the late reply.

jwk/helper.go
```diff
@@ -137,14 +139,20 @@ func FindPublicKey(set *jose.JSONWebKeySet) (key *jose.JSONWebKey, err error) {
 }
 
 func ExcludePrivateKeys(set *jose.JSONWebKeySet) *jose.JSONWebKeySet {
```
TimoDoernenburg, line 141: Updated the function based on the last comment.

aeneasr:
Thank you, however it looks like test are not passing. Could you please address the failures?

jwk/helper_test.go
```diff
@@ -71,3 +71,23 @@ func TestIder(t *testing.T) {
...
+	var testRSGenerator = RS256Generator{}
+	var testECDSAGenerator = ECDSA256Generator{}
+
+	t.Run("Test_Helper/Run_FindPublicKey_With_RSA", func(t *testing.T) {
+		RSIDKS, _ := testRSGenerator.Generate("test-id-1", "sig")
+		keys, err := FindPublicKey(RSIDKS)
+		require.NoError(t, err)
+		assert.Equal(t, keys.KeyID, Ider("public", "test-id-1"))
+
```
aeneasr, line 84: One last thing, please ensure here that the key is indeed a public key:


```suggestion
		assert.IsType(t, keys.Key, new(rsa.PublicKey))
```

jwk/helper_test.go
```diff
@@ -71,3 +71,23 @@ func TestIder(t *testing.T) {
...
+	})
+	t.Run("Test_Helper/Run_FindPublicKey_With_ECDSA", func(t *testing.T) {
+		ECDSAIDKS, _ := testECDSAGenerator.Generate("test-id-2", "sig")
+		keys, err := FindPublicKey(ECDSAIDKS)
+		require.NoError(t, err)
+		assert.Equal(t, keys.KeyID, Ider("public", "test-id-2"))
+
```
aeneasr, line 91:
```suggestion
		assert.IsType(t, keys.Key, new(ecdsa.PublicKey))
```

aeneasr (changes requested):

TimoDoernenburg:
The tests are passing now.

Issue #2436, opened 2021-03-31
ImportKeys command does not support updates?


fjvierap:
**Describe the bug**

While importing keys keys are not update but removed if they were already presented.

**Reproducing the bug**

Run `keys import setName priv.pem pub.pem --skip-tls-verify` once
Check `setName` is presented in table `hydra_jwk`
Run again `keys import setName priv.pem pub.pem --skip-tls-verify` 
Check `setName` is not presented in table `hydra_jwk`

_Server logs_

{"_http_request":{"headers":{"accept-encoding":"gzip","user-agent":"Go-http-client/1.1"},"host":"hydra-admin:4445","method":"PUT","path":"/keys/setName","query":null,"remote":"10.0.36.165:59012","scheme":"https"},"_http_response":{"status":409,"text_status":"Conflict","took":3036394},"_pid":1,"level":6,"level_name":"info","short_message":"completed handling request","timestamp":1617188391.0631,"version":"1.1"}

**Additional context**

I think the issue is in `cmd/cli/handler_jwk.go`

```
var set jose.JSONWebKeySet
err = json.NewDecoder(response.Body).Decode(&set)
```

There as the keys were previously removed them `response.Body` contains the  set key which the will be added again... So when `set` contains every key duplicated. I think this could be fixed by replacing:
```
var set jose.JSONWebKeySet
err = json.NewDecoder(response.Body).Decode(&set)
```
for:
```
set := jose.JSONWebKeySet{}
```

PR #2460, opened 2021-04-12
feat: ImportKeys command does not support updates

naveenpaul1:
## Related issue

https://github.com/ory/hydra/issues/2436

@aeneasr  @Benehiko 

## Proposed changes

While importing keys keys are not update but removed if they were already presented.

Run keys import setName priv.pem pub.pem --skip-tls-verify once
Check setName is presented in table hydra_jwk
Run again keys import setName priv.pem pub.pem --skip-tls-verify
Check setName is not presented in table hydra_jwk

## Checklist

      the changes.
      works.

## Further comments

I think the issue is in cmd/cli/handler_jwk.go

var set jose.JSONWebKeySet
err = json.NewDecoder(response.Body).Decode(&set)
There as the keys were previously removed them response.Body contains the set key which the will be added again... So when set contains every key duplicated. I think this could be fixed by replacing:

var set jose.JSONWebKeySet
err = json.NewDecoder(response.Body).Decode(&set)
for:

set := jose.JSONWebKeySet{}


aeneasr:
Thank you! Could you please add a failing test case which shows the broken behaviour so it is easier to understand what is going on and also prevent future regressions? :)

cmd/cli/handler_jwk_test.go
```diff
@@ -60,4 +68,35 @@ func Test_toSDKFriendlyJSONWebKey(t *testing.T) {
...
+	cmd.Flags().Bool("fake-tls-termination", false, "Sets the \"use\" value of the JSON Web Key if not \"use\" value was defined by the key itself")
+	cmd.Flags().String("access-token", "", "Set an access token to be used in the Authorization header, defaults to environment variable OAUTH2_ACCESS_TOKEN")
+	cmd.Flags().String("endpoint", "", "Set the URL where ORY Hydra is hosted, defaults to environment variable HYDRA_ADMIN_URL. A unix socket can be set in the form unix:///path/to/socket")
+	cmd.Flags().Bool("skip-tls-verify", true, "Foolishly accept TLS certificates signed by unknown certificate authorities")
+	os.Setenv("HYDRA_URL", testServer.URL)
+	t.Run("tt.name", func(t *testing.T) {
+		NewHandler().Keys.ImportKeys(&cmd, []string{"setName", "../test/private_key.json", "../test/public_key.json"})
+		//running again to make sure the row in storage is not deleted
+		NewHandler().Keys.ImportKeys(&cmd, []string{"setName", "../test/private_key.json", "../test/public_key.json"})
+
```
aeneasr, line 99: We still need to test if the keys are in the store as expected right? :)

naveenpaul1: @aeneasr : unit test is updated, now we are checking the keys in db after key import

aeneasr (changes requested):
I reverted the changes in handler_jwk to see if the test caused a failure and received the following error:

```
{
	"error": "Unable to insert or update resource because a resource with that value exists already",
	"error_description": ""
}
```

indicating that the key was already imported. The test does thus not describe the original bug which was reported as:

1. Run `hydra keys import setName priv.pem pub.pem --skip-tls-verify`
2. Run `hydra keys import setName priv.pem pub.pem --skip-tls-verify` again
3. JWK from the first / second call is not in the store

Instead I observed (in the test):

1. Run `hydra keys import setName priv.pem pub.pem --skip-tls-verify`
2. Run `hydra keys import setName priv.pem pub.pem --skip-tls-verify` again but fails because key already exists
3. JWK is still in the store

This also makes sense when looking at the test. The import is not a replace operation but instead appends the key(s) to the existing set - thus "echoing" the existing keys back to the API.

Could you maybe elaborate more specifically what is happening here?

naveenpaul1:
> I reverted the changes in handler_jwk to see if the test caused a failure and received the following error:
> 
> ```
> {
> 	"error": "Unable to insert or update resource because a resource with that value exists already",
> 	"error_description": ""
> }
> ```
> 
> indicating that the key was already imported. The test does thus not describe the original bug which was reported as:
> 
> 1. Run `hydra keys import setName priv.pem pub.pem --skip-tls-verify`
> 2. Run `hydra keys import setName priv.pem pub.pem --skip-tls-verify` again
> 3. JWK from the first / second call is not in the store
> 
> Instead I observed (in the test):
> 
> 1. Run `hydra keys import setName priv.pem pub.pem --skip-tls-verify`
> 2. Run `hydra keys import setName priv.pem pub.pem --skip-tls-verify` again but fails because key already exists
> 3. JWK is still in the store
> 
> This also makes sense when looking at the test. The import is not a replace operation but instead appends the key(s) to the existing set - thus "echoing" the existing keys back to the API.
> 
> Could you maybe elaborate more specifically what is happening here?

naveenpaul1:
> I reverted the changes in handler_jwk to see if the test caused a failure and received the following error:
> 
> ```
> {
> 	"error": "Unable to insert or update resource because a resource with that value exists already",
> 	"error_description": ""
> }
> ```
> 
> indicating that the key was already imported. The test does thus not describe the original bug which was reported as:
> 
> 1. Run `hydra keys import setName priv.pem pub.pem --skip-tls-verify`
> 2. Run `hydra keys import setName priv.pem pub.pem --skip-tls-verify` again
> 3. JWK from the first / second call is not in the store
> 
> Instead I observed (in the test):
> 
> 1. Run `hydra keys import setName priv.pem pub.pem --skip-tls-verify`
> 2. Run `hydra keys import setName priv.pem pub.pem --skip-tls-verify` again but fails because key already exists
> 3. JWK is still in the store
> 
> This also makes sense when looking at the test. The import is not a replace operation but instead appends the key(s) to the existing set - thus "echoing" the existing keys back to the API.
> 
> Could you maybe elaborate more specifically what is happening here?


> I reverted the changes in handler_jwk to see if the test caused a failure and received the following error:
> 
> ```
> {
> 	"error": "Unable to insert or update resource because a resource with that value exists already",
> 	"error_description": ""
> }
> ```
> 
> indicating that the key was already imported. The test does thus not describe the original bug which was reported as:
> 
> 1. Run `hydra keys import setName priv.pem pub.pem --skip-tls-verify`
> 2. Run `hydra keys import setName priv.pem pub.pem --skip-tls-verify` again
> 3. JWK from the first / second call is not in the store
> 
> Instead I observed (in the test):
> 
> 1. Run `hydra keys import setName priv.pem pub.pem --skip-tls-verify`
> 2. Run `hydra keys import setName priv.pem pub.pem --skip-tls-verify` again but fails because key already exists
> 3. JWK is still in the store
> 
> This also makes sense when looking at the test. The import is not a replace operation but instead appends the key(s) to the existing set - thus "echoing" the existing keys back to the API.
> 
> Could you maybe elaborate more specifically what is happening here?

@aeneasr : issue is in method `jwk.UpdateKeySet()`, where `DeleteKeySet()` is executing with value we are passing in `set` but `AddKeySet()` is failing because we are trying to add two key set with same `set` and `key` values and we are getting error you are mentioned in previous chat, at the same time already existing row is getting removed from db because of delete.
I tried to reproduce this in unit test but couldn't run the 
`m.GetKeySet(context.TODO(), "import-1")`
when the second import fails test case execution is getting stoped, since `cli,ImportKeys()` method exit the execution when update key set (PUT /keys/setName) retun not 200 in below line
`cmdx.CheckResponse(err, http.StatusOK, response)`,  Please correct me if am wrong

aeneasr:
Could you maybe push the test case you have problems with? That would help me understand what is failing! Thanks :)

naveenpaul1:
@aeneasr : I just updated the unit test case which covers multiple imports with the same key values, Since `ImportKeys()` method exit with `os.Exit(1)`, we need to catch it and execute the `GetKeySet()` method within the catch block to make sure imported key is not in db.

fjvierap:
> Thank you! Could you please add a failing test case which shows the broken behaviour so it is easier to understand what is going on and also prevent future regressions? :)

Hi @aeneasr I was checking this and it is not easy to add a error test for this...  Note the code which was there:
```
func (h *JWKHandler) ImportKeys(cmd *cobra.Command, args []string) {
	cmdx.MinArgs(cmd, args, 2)

	id := args[0]
	use := flagx.MustGetString(cmd, "use")
	client := &http.Client{}

	/* #nosec G402 - we want to support dev environments, hence tls trickery */
	client.Transport = &http.Transport{
		TLSClientConfig: &tls.Config{
			InsecureSkipVerify: flagx.MustGetBool(cmd, "skip-tls-verify"),
		},
	}

	u := Remote(cmd) + "/keys/" + id
	request, err := http.NewRequest("GET", u, nil)
	cmdx.Must(err, "Unable to initialize HTTP request: %s", err)

	if flagx.MustGetBool(cmd, "fake-tls-termination") {
		request.Header.Set("X-Forwarded-Proto", "https")
	}

	if token := flagx.MustGetString(cmd, "access-token"); token != "" {
		request.Header.Set("Authorization", "Bearer "+token)
	}

	response, err := client.Do(request)
	cmdx.Must(err, "Unable to fetch data from %s: %s", u, err)
	defer response.Body.Close()

	if response.StatusCode != http.StatusOK && response.StatusCode != http.StatusNotFound {
		cmdx.Fatalf("Expected status code 200 or 404 but got %d while fetching data from %s", response.StatusCode, u)
	}

	set := jose.JSONWebKeySet{}
	for _, path := range args[1:] {
		file, err := ioutil.ReadFile(path)
		cmdx.Must(err, "Unable to read file %s", path)

		if key, privateErr := josex.LoadPrivateKey(file); privateErr != nil {
			key, publicErr := josex.LoadPublicKey(file)
			cmdx.Must(publicErr, `Unable to read key from file %s. Decoding file to private key failed with reason "%s" and decoding it to public key failed with reason: %s`, path, privateErr, publicErr)

			set.Keys = append(set.Keys, toSDKFriendlyJSONWebKey(key, "public:"+uuid.New(), use))
		} else {
			set.Keys = append(set.Keys, toSDKFriendlyJSONWebKey(key, "private:"+uuid.New(), use))
		}

		fmt.Printf("Successfully loaded key from file: %s\n", path)
	}

	body, err := json.Marshal(&set)
	cmdx.Must(err, "Unable to encode JSON Web Keys to JSON: %s", err)

	request, err = http.NewRequest("PUT", u, bytes.NewReader(body))
	cmdx.Must(err, "Unable to initialize HTTP request: %s", err)

	if flagx.MustGetBool(cmd, "fake-tls-termination") {
		request.Header.Set("X-Forwarded-Proto", "https")
	}

	if token := flagx.MustGetString(cmd, "access-token"); token != "" {
		request.Header.Set("Authorization", "Bearer "+token)
	}
	request.Header.Set("Content-Type", "application/json")

	response, err = client.Do(request)
	cmdx.CheckResponse(err, http.StatusOK, response)
	defer response.Body.Close()

	fmt.Println("JSON Web Key Set successfully imported!")
}
```
There for all error scenarios `cmdx.Must` is exiting the program with `cmdx.Must`How do you unit test those cases?

I checked already `root_test.go` which actually has this code:
```
err := rootCmd.Execute()
				if c.expectErr {
					assert.Error(t, err)
				} else {
					assert.NoError(t, err)
				}
```
Which I tested and `c.expectErr`is not used because failed scenarios exit the codes... and I think that is the reason there is no errors scenarios defined there.

aeneasr:
Hm yeah the problem is that I think that this PR introduces other issues with the import command because it overwrites a key completely? And what we probably want is to append the imported keys to the list. I also checked the HTTP handlers. The method used is

https://github.com/ory/hydra/blob/64a332a98fc1c3a73e4b39b58b21b4cd61f5b240/jwk/handler.go#L246-L266

Here the handler is removing the key set and then adding all the imported keys. Therefore, the current behaviour is correct in my opinion.

I think the problem described in the opening PR is that, if you import a key that already exists, that fails, because the key is now being sent as a duplicate. The proper resolution for that should be to check in the import handler if the imported keys match one of the existing keys, and update those.

Therefore, we definitely need a test to cover this. The easiest method would probably be to have `ImportKeys(cmd *cobra.Command, args []string) error` instead and not rely on `cmd.Must()`. This would be a bit of refactoring though.

I'll mark this PR as a draft for now because it will introduce new bugs (removing existing keys) and the bug this PR solves is a special case that should be handled differently!

cmd/cli/handler_jwk.go
```diff
@@ -123,10 +123,7 @@ func (h *JWKHandler) ImportKeys(cmd *cobra.Command, args []string) {
 		cmdx.Fatalf("Expected status code 200 or 404 but got %d while fetching data from %s", response.StatusCode, u)
 	}
 
-	var set jose.JSONWebKeySet
-	err = json.NewDecoder(response.Body).Decode(&set)
-	cmdx.Must(err, "Unable to decode payload to JSON: %s", err)
-
+	set := jose.JSONWebKeySet{}
 	for _, path := range args[1:] {
```
aeneasr, line 127: Instead of ignoring the existing keys we should make sure that we update any keys that have a matching key id :)

aeneasr (changes requested):

cmd/cli/handler_jwk.go
TimoDoernenburg: Hey, I updated the branch and added an update function in the importer. Existing keys (same KeyID and SID) will be replaced and new keys (new KeyID and same SID) will be added to the list.
I also updated the test to test the changes.

aeneasr:
@TimoDoernenburg - thank you! I merged with master, let's see if the CI passes :) The CLA also has an issue:

Thank you, this looks great! It looks like the CLA bot is not properly detecting your signature. To fix this, try the following:

```
$ git commit  --amend --author="Author Name <email@address.com>"
```

Ensure that `Author Name` is replaced with your GitHub username (e.g. `aeneasr`) and that the email address is replaced with the email address you have set up in GitHub (e.g. `3372410+aeneasr@users.noreply.github.com`):

```
$ git commit  --amend --author="aeneasr <3372410+aeneasr@users.noreply.github.com>"
```

Once that is done, you can force-push your changes (make sure to push to the correct remote and branch!):

```
$ git push --force <remote> HEAD:<branch>
```

cmd/cli/handler_jwk.go
```diff
@@ -89,6 +89,21 @@ func toSDKFriendlyJSONWebKey(key interface{}, kid, use string) jose.JSONWebKey {
 	}
 }
 
+func updateKey(set jose.JSONWebKeySet,newKey jose.JSONWebKey) []jose.JSONWebKey {
```
aeneasr, line 92: It would be great if you could add a unit test for this function :)

aeneasr (changes requested):
Thank you @TimoDoernenburg - this looks much better now! :) Got some more ideas how to improve it!

cmd/cli/handler_import_jwk_test.go
```diff
@@ -52,3 +53,40 @@ func Test_toImportJSONWebKey(t *testing.T) {
 	assert.NotEmpty(t, v)
 
 }
+
+func Test_toUpdateKeyList(t *testing.T) {
```
TimoDoernenburg, line 57: I added another test for the importer updatekey function.

aeneasr:
Thank you! So is this ready for another review? Sorry that I missed this, since this is not your PR I assume you can't mark this as "Ready for Review". PRs that are marked as draft are not shown on my dashboard :/

aeneasr:
Thank you, could you please address the failing test? 🙌 

https://app.circleci.com/pipelines/github/ory/hydra/2776/workflows/f718b951-bbac-4f1d-9839-37847c217bc9/jobs/32588

aeneasr (approved):

cmd/cli/handler_import_jwk_test.go
```diff
@@ -0,0 +1,92 @@
...
+	v, _ = m.GetKeySet(context.TODO(), "import-1")
+	assert.Equal(t, len(v.Keys),3)
+
+	NewHandler().Keys.ImportKeys(&cmd, []string{"import-2", "../test/private_key.json", "../test/public_key.json"})
+	v, _ = m.GetKeySet(context.TODO(), "import-2")
+	assert.NotEmpty(t, v)
+
+}
+
+func Test_toUpdateKeyList(t *testing.T) {
```
aeneasr, line 57:
```suggestion
func TestUpdateKeyList(t *testing.T) {
```

cmd/cli/handler_import_jwk_test.go
```diff
@@ -0,0 +1,92 @@
...
+
```
aeneasr, line 58:
```suggestion
```

cmd/cli/handler_import_jwk_test.go
```diff
@@ -0,0 +1,92 @@
...
+	key := set.Key("static-sid")[0]
+	verifyKey := toSDKFriendlyJSONWebKey(anotherPEM, "static-sid", "sig")
+	assert.Equal(t, key,verifyKey)
+
+	//adding a third key with different sid
+	set.Keys = updateKey(set,toSDKFriendlyJSONWebKey(anotherPEM, "different-sid", "sig") )
+	assert.Equal(t, len(set.Keys),3)
+
+
```
aeneasr, line 90: Add a check that `another-sid` is still `dummyPEM` :)

cmd/cli/handler_jwk.go
```diff
@@ -89,6 +89,21 @@ func toSDKFriendlyJSONWebKey(key interface{}, kid, use string) jose.JSONWebKey {
 	}
 }
 
+func updateKey(set jose.JSONWebKeySet, newKey jose.JSONWebKey) []jose.JSONWebKey {
```
aeneasr, line 92:
```suggestion
// updateKey replaces a key in the JSON Key Set or appends it if the key ID is not yet known.
func updateKey(set jose.JSONWebKeySet, newKey jose.JSONWebKey) []jose.JSONWebKey {
```

aeneasr (changes requested):
Only some minor code format things (run `make format` to auto-format the code). Also one, last test - then I will merge this! :)

aeneasr:
Thank you, this looks great! The CI is failing because some files are formatted incorrectly. To format them, run:

```
$ make format
$ git commit -a -m "styles: format code"
$ git push
```

Thank you!

TimoDoernenburg:
The tests are passing now.

aeneasr (approved):
Thank you all!

PR #2473, opened 2021-04-16
fix: Respond with Redirect URL for already handled requests

### LINKED ISSUE #1569: No way to handle 409 GetLoginRequestConflict.
**Describe the bug**

Sometimes we experience `409 GetLoginRequestConflict` errors when calling Admin API **GetLoginRequest**. Logically, there is no error, as *used* login request cannot be used anymore. But the problem is that there is no way to handle it from the Identity Provider point of view:

1) Identity Provider gets request e.g. `/login?login_challenge=abc123` from the user.
2) Identity Provider gets response with 409 error with no helpful details from Hydra Admin API.
3) Identity Provider does not know anything about OAuth Client, he can get this info only doing `GetLoginRequest` to Admin API.
4) Identity Provider cannot redirect the user back to the beginning of the login process (OAuth Client), so the user can either re-authenticate or continue using the Resource Provider is authenticated.
5) Identity Provider can only fail the user request, showing typical "Something went wrong, go back where you started" or something.

Note that while we should not get this error in "normal" flow, our users still getting them occasionally, maybe due to "double-click" phenomena or other client-side issues, which we cannot always control.

**Reproducing the bug**

This can be trivially reproduced by doing typical login/consent flow, and then trying to call **GetLoginRequest** again with the same login challenge.

*Server logs*

```
"error":"[GET /oauth2/auth/requests/login][409] getLoginRequestConflict &{Code:0 Debug:Login request has been used already Hint: Name:0xc00066e9f0}"
```

**Expected behavior**

One solution can be returning 200 response from Hydra even if the login challenge is used. The response can contain `WasUsed` field to indicate that this login challenge is not available for accepting, but at least we can extract client info and redirect the user back to restart the login process.

*Note: I think that returning 200 response instead of 409 will be a breaking change.*

Or, somehow include client info in the error response.

Or, maybe, the `redirect` field should be included into all 4xx responses so Identity Provider can at least redirect the user back to restart the login process on any similar errors.

**Additional context**

Code reference: https://github.com/ory/hydra/blob/master/consent/handler.go#L266

Am I missing something? If so, please point me how to handle this problem.

Or, if this should be fixed, please let me know if any help wanted, so I can provide a PR.

svrakitin:
## Related issue

Closes #1569

## Proposed changes

- [x] Use `410 Gone` instead of `409 Conflict` for already handled requests
- [x] Respond with `redirect_to` when getting Login/Consent/Logout requests which have already been handled.
- [x] Make it more consistent by replacing `WasUsed` with `WasHandled` everywhere.

## Checklist

      the changes.
      works.


spec/api.json
```diff
@@ -2153,6 +2141,10 @@
         "Types"
       ],
       "properties": {
+        "ProtocolScheme": {
+          "description": "Protocol to use for clients connecting to the plugin.",
+          "type": "string"
+        },
```
Benehiko, line 2147: Is this necessary?

Benehiko (commented):
Was it necessary to change everything from `WasUsed` to `WasHandled`? Other than that it looks good to me :)

svrakitin:
> Was it necessary to change everything from `WasUsed` to `WasHandled`? Other than that it looks good to me :)

@Benehiko I just made it more consistent as both WasUsed and WasHandled were mentioned in the code. I believe that request cannot really be used, it can be handled tho. :)

spec/api.json
svrakitin: This is autogenerated, probably not.

Benehiko (dismissed):
Really appreciate the contribution :)
I'll approve this, but will require further feedback from @aeneasr

consent/types.go
```diff
@@ -596,14 +606,16 @@ type ConsentRequest struct {
 	// Context contains arbitrary information set by the login endpoint or is empty if not set.
 	Context sqlxx.JSONRawMessage `json:"context,omitempty" db:"context"`
 
+	// Implies that the request has already been handled.
```
aeneasr, line 609:
```suggestion
	// If set to true means that the request was already handled. This
	// can happen on form double-submit or other errors. If this is set
	// we recommend redirecting the user to `request_url` to re-initiate
	// the flow.
```

aeneasr (commented):
Awesome, thank you! I've added a more descriptive comment. Can you please apply that comment to all the definitions of `WasHandled`?

Alternatively to this approach, we could do:

```
h.r.Writer().WriteError(w, r, x.ErrConflict.WithReason("Consent request has been used already").WithDetail("redirect_to", r.ReturnURL))
```

and write a custom swagger error object for this (so not `genericError` but e.g. `requestWasHandledError`.

What do you think? Is the 429 to counter intuitive? In my view, this is an error and should be treated as an error. If the dev does not check for this, the flow will end up in an irrecoverable error for the user.

Alternatively we could also let hydra redirect the user to the RequestURL if we encounter a used login / consent request?

svrakitin:
@aeneasr I think replying with `409 Conflict` on `GET` is not very intuitive as I would only expect conflict when mutating some state. If we still want to have an error here an alternative could be using different status code (i.e. `410 Gone`) and still respond with `redirect_to`/`request_url` in a different swagger object as you mentioned.

svrakitin:
@aeneasr Updated PR to keep `409`, but respond with `redirect_to`.

aeneasr:
`410 Gone` is a good idea!

aeneasr:
True, it is a breaking change - but given that we fix the behavior which was basically completely broken before it is ok. We'll document it :)

svrakitin:
Will make it 410 then! Thanks

aeneasr:
Awesome @svrakitin !

sneko:
Hi @svrakitin ,

Just upgraded my Hydra instance to benefit from your PR (thanks).

I noticed with my Go client I'm able to get the `*admin.GetLoginRequestGone` when running `GetLoginRequest`, which seems expected.

But when doing `AcceptLoginRequest` (I was doing only this one before when the user submit credentials), I get this type of error:
`*runtime.APIError`
(and once stringified it gives `response status code does not match any response statuses defined for this endpoint in the swagger spec (status 409): {}`)

It brings to me some questions:
- It seems the 409 is still used for accepting a login?
- Should the Go client have the possible "Gone" error on `AcceptLoginRequest` as for `GetLoginRequest`? Or should I always do a `Get` before `Accept` just to make sure it's valid (considering in won't expire between both requests?

Thank you,

svrakitin:
Hey @sneko,

I think it makes sense to revisit status codes for accept endpoints as well. For now as a workaround you can get login request first. `409` is used for accept, but it's not even mentioned in the swagger spec, thats why you get this generic error.

You can either submit a PR to fix that or I will find some time later this month.

sneko:
@svrakitin just noticed that:

Some users started the flow into a specific browser, and then, due to confirmation through email... they went to another browser.

And even if they had a code challenge not used, `GetLoginRequest` was returning no `Gone` error, but the following `AcceptLoginRequest` was returning `response status code does not match any response statuses defined for this endpoint in the swagger spec (status 409): {}`.

I wonder what could make the difference so Hydra is complaining?

Let's see my code:
```
	getResponse, err := s.hydra.Admin.GetLoginRequest(&hydra_admin.GetLoginRequestParams{
		Context:        ctx,
		LoginChallenge: challenge,
	})
	if err != nil {
                 return err
	}
```
and 
```
	acceptResponse, err := s.hydra.Admin.AcceptLoginRequest(&hydra_admin.AcceptLoginRequestParams{
		Context:        ctx,
		LoginChallenge: challenge,
		Body: &hydra_models.AcceptLoginRequest{
			Subject:     &auth.UserID,
			Remember:    true,
			RememberFor: int64(0), // Must be in seconds, and "0" for browser session lifetime
		},
	})
	if err != nil {
                 return err
	}
````

For me it's not coming from the missing `Gone`, I could still look for `status 409` in the stringified error to adjust my frontend... but I don't get why the getter is a success if accepting is not :(

**Comment by svrakitin** [comment]
@sneko Could it be some race condition when you have multiple requests for the same challenge? It can be users refreshing/reusing the email confirmation page if you accept the challenge automatically there.

**Comment by sneko** [comment]
@svrakitin to be honest it happened to several users in a short time and the only common thing I can see is a potential change of browser after clicking the email.

On the Hydra instance I can see:
```
time=2021-07-28T15:17:41Z level=info msg=started handling request http_request=map[headers:map[accept:application/json accept-encoding:gzip user-agent:Go-http-client/1.1 x-forwarded-proto:http x-request-id:c2138e61-824a-907f-940d-c9e577c414e4] host:xxxxxxxxxxxx:4445 method:PUT path:/oauth2/auth/requests/login/accept query:Value is sensitive and has been redacted. To see the value set config key "log.leak_sensitive_values = true" or environment variable "LOG_LEAK_SENSITIVE_VALUES=true". remote:127.0.0.1:52180 scheme:http]
time=2021-07-28T15:17:41Z level=error msg=An error occurred while handling a request audience=application error=map[debug: message:mysql create: Error 1062: Duplicate entry 'e09da8f5a3bf40ae9d9bb8413c647208' for key 'PRIMARY': Unable to insert or update resource because a resource with that value exists already reason: status:Conflict status_code:409] http_request=map[headers:map[accept:application/json accept-encoding:gzip user-agent:Go-http-client/1.1 x-forwarded-proto:http x-request-id:c2138e61-824a-907f-940d-c9e577c414e4] host:xxxxxxxxxxxx:4445 method:PUT path:/oauth2/auth/requests/login/accept query:Value is sensitive and has been redacted. To see the value set config key "log.leak_sensitive_values = true" or environment variable "LOG_LEAK_SENSITIVE_VALUES=true". remote:127.0.0.1:52180 scheme:http] http_response=map[status_code:409] service_name=ORY Hydra service_version=v1.10.2
time=2021-07-28T15:17:41Z level=info msg=completed handling request http_request=map[headers:map[accept:application/json accept-encoding:gzip user-agent:Go-http-client/1.1 x-forwarded-proto:http x-request-id:c2138e61-824a-907f-940d-c9e577c414e4] host:xxxxxxxxxxxx:4445 method:PUT path:/oauth2/auth/requests/login/accept query:Value is sensitive and has been redacted. To see the value set config key "log.leak_sensitive_values = true" or environment variable "LOG_LEAK_SENSITIVE_VALUES=true". remote:127.0.0.1:52180 scheme:http] http_response=map[status:409 text_status:Conflict took:114.244417ms]
```

But I can get what's happening. I can just affirm that the duplicate entry identified as `e09da8f5a3bf40ae9d9bb8413c647208` is the same for all users where it failed.

I need to ding into the DB to see what makes it shared for all users...

In the meantime if you have an idea :) thanks!

**Comment by sneko** [comment]
@svrakitin after investigating I can see on the Hydra instance some logs like:
```
[2021-07-27T19:52:44.014Z] "PUT /oauth2/auth/requests/login/accept?login_challenge=e09da8f5a3bf40ae9d9bb8413c647208 HTTP/1.1" 200 - "-" "-" 67 411 21 20 "-" "Go-http-client/1.1" "6b8d4966-85fe-972f-8fa4-f446e6326499" "hydra-admin:4445" "127.0.0.1:4445" inbound|4445|http|xxxx.cluster.local 127.0.0.1:36750 10.24.0.16:4445 10.24.0.24:49778 outbound_.4445_._.xxxx.cluster.local default

[2021-07-27T19:52:51.723Z] "PUT /oauth2/auth/requests/login/accept?login_challenge=e09da8f5a3bf40ae9d9bb8413c647208 HTTP/1.1" 409 - "-" "-" 67 121 9 9 "-" "Go-http-client/1.1" "9c8158bb-d21c-93f7-91f3-f2c2eaa7fdfc" "hydra-admin:4445" "127.0.0.1:4445" inbound|4445|http|xxxx.cluster.local 127.0.0.1:36750 10.24.0.16:4445 10.24.0.24:49778 outbound_.4445_._.xxxx.cluster.local default
```

Which prove indeed there is a race condition.

I guess I have something to fix on my end to prevent those double-requests at same time (it happens to almost no one). But I think there is also to do on Hydra side because if it cannot `AcceptLoginRequest`, there is definitely no reason for it to continue returning `GetLoginRequest` without the same 410 (old 409) error.

WDYT?

PR #2476, opened 2021-04-16
feat: Split TLS config into admin and public interfaces

### LINKED ISSUE #1962: Split HTTPS handling for public/admin

When deploying ORY Hydra to prod/stage/..., the admin endpoint is typically inwards facing (internal network) and the public endpoint is usually exposed behind a LoadBalancer that does TLS termination.

The admin endpoint may or may not be secured using HTTPS (e.g. mTLS when using Istio/Linkerd/...) but since the traffic is only internal (e.g. checking if a token is valid) it should be up to the deployment team to decide if that traffic is encrypted or not (using e.g. mTLS).

Several pieces - for example ORY Oathkeeper - require some tricky configuration to get `X-Forwarded-Proto` with the right value working while not adding any actual security to the process.

Similarly, disabling strict SSL verification to work around this is also not very useful as it would make it easy to MITM traffic within the internal network.

In general, the idea to enforce HTTPs really comes from "no external internet traffic should be able to hit ORY Hydra without being TLS traffic". This however shouldn't make it painful for devs/ops to deploy the stack and use it in their own infrastructure.

I currently lack ideas of how to solve it, but the idea of this issue is to start a discussion and come up with the best possible solution.

**Additional context**

I wanted to link all the issues that have been opened because of this but will do so at a later stage.

svrakitin:
## Related issue

Closes #1962 

## Proposed changes

- [x] Allow different TLS configuration for public and admin interfaces. 
- [x] Remove duplication in config by parameterizing accessors with `ServeInterface` interface param.
- [x] Update and add some tests
- [x] Update config spec   

## Checklist

      the changes.
      works.

## Comments

Probably it is possible to remove duplication even further, but it would increase the size of this PR and it is already rather huge.


aeneasr:
Wow thank you! I will have to give this a good look which I will be able to do this week :)

consent/strategy_default.go
```diff
@@ -109,7 +109,8 @@ func (s *DefaultStrategy) matchesValueFromSession(ctx context.Context, c fosite.
```
aeneasr, line 109: These changes here are cleaning up variables and so on right? I think they should be made as a separate PR because it makes it harder for me to review what belongs to what. Thank you!

aeneasr (changes requested):

aeneasr:
Sorry for taking some time on this - it would be great if you could reduce the PR to changes which affect HTTPS only as it will make it easier for me to understand what you did :)

svrakitin:
Sure, no probs.

svrakitin:
Updated PR to include only changes related to split.

driver/config/tls.go
```diff
@@ -0,0 +1,70 @@
...
+type TLSConfig interface {
+	Strict() bool
+	AllowTerminationFrom() []string
+	Certificate() ([]tls.Certificate, error)
+}
+
+func (p *Provider) TLS(iface ServeInterface) TLSConfig {
+	strict := false
+	if !p.forcedHTTP() {
+		strict = p.p.Bool(iface.Key(KeySuffixTLSStrict))
```
aeneasr, line 33: I'm not entirely sure what the new `tls.strict` is for? Is there a reason why we need it on top of `--dangerous-force-http`? Also, doesn't that mean that we need to set `strict` to `false` (default is `true`) for `--dangerous-force-http` to have an effect?

aeneasr (commented):
Thank you, this looks really good - nice job! Just one question, I don't fully understand what `tls.strict` is for? Is it because you wanted to add the ability in this PR to use TLS for public but not for admin, and needed a place to define this? If so, maybe we need to make `--dangerous-force-http` only error if there is no TLS set for public, but not so for admin?

svrakitin:
Yeah, this is just to be able to turn off strict TLS for admin and force TLS only for public. This flag makes the setup more obvious and straightforward for an operator than termination allow-list. 

--dangerous-force-http now makes TLS non-strict for both public and admin to keep backward compatible behavior. Effectively this PR deprecates this flag in favor of "*tls.strict" parameters, which can be sourced from the config or env vars.

I updated defaults in config spec to have strict TLS only for public. Let me know if admin should force TLS by default as well.

driver/config/tls.go
svrakitin: This code should be probably improved, it just makes it respect dangerous-force-http flag over tls.strict config param.

aeneasr: Yeah makes sense, I misread what this is doing here! This is ok :)

driver/config/tls.go
```diff
@@ -0,0 +1,70 @@
...
+
+	KeyTLSAllowTerminationFrom = "serve." + KeySuffixTLSAllowTerminationFrom
+	KeyTLSCertString           = "serve." + KeySuffixTLSCertString
+	KeyTLSKeyString            = "serve." + KeySuffixTLSKeyString
+	KeyTLSCertPath             = "serve." + KeySuffixTLSCertPath
+	KeyTLSKeyPath              = "serve." + KeySuffixTLSKeyPath
+)
+
+type TLSConfig interface {
+	Strict() bool
```
aeneasr, line 25: I would rename this to `Enabled` because this is what this does :)

driver/config/tls.go
```diff
@@ -0,0 +1,70 @@
+package config
+
+import (
+	"crypto/tls"
+
+	"github.com/ory/x/tlsx"
+)
+
+const (
+	KeySuffixTLSStrict               = "tls.strict"
```
aeneasr, line 10:
```suggestion
	KeySuffixTLSStrict               = "tls.enabled"
```

driver/config/tls.go
```diff
@@ -0,0 +1,70 @@
...
+	Strict() bool
+	AllowTerminationFrom() []string
+	Certificate() ([]tls.Certificate, error)
+}
+
+func (p *Provider) TLS(iface ServeInterface) TLSConfig {
+	strict := false
+	if !p.forcedHTTP() {
+		strict = p.p.Bool(iface.Key(KeySuffixTLSStrict))
+	}
```
aeneasr, line 34: I would change this so that:

1. Public can have TLS disabled if forcedHTTP is true, otherwise it is enabled - hence remove `public.tls.strict`
2. Admin can have TLS disabled if forcedHTTP is true, or if `admin.tls.enabled` is `false`.

spec/config.json
```diff
@@ -201,6 +216,41 @@
...
+    "tls_config": {
+      "type": "object",
+      "additionalProperties": false,
+      "description": "Configures HTTPS (HTTP over TLS). If configured, the server automatically supports HTTP/2.",
+      "properties": {
+        "strict": {
+          "type": "boolean",
+          "description": "Disallow any plaintext (HTTP) traffic and force HTTPS.",
+          "default": true
+        },
```
aeneasr, line 229: Rename to `enabled` and only allow this to be used in the admin endpoint

```suggestion
        "enabled": {
          "type": "boolean",
          "description": "Setting enabled to false drops the TLS requirement for the admin endpoint, even if TLS is enabled on the public endpoint.",
          "default": true
        },
```

aeneasr (commented):
Just two more minor things then this is good to go!

aeneasr:
Note to self: check all the quickstarts and articles to see if this works with this change

svrakitin:
@aeneasr Done, kept `tls.enabled` just for admin and resolved some conflicts.

aeneasr (approved):
Great job man, thank you!

aeneasr:
@svrakitin completely forgot that we should add some docs 🤦‍♂️ Could you maybe add a bit of guidance on how to use your new features here: https://www.ory.sh/hydra/docs/production

Would highly appreciate it! :)

svrakitin:
@aeneasr Sure, I will dedicate some time tomorrow for this.

aeneasr:
Thank you!!!

PR #2483, opened 2021-04-22
feat: Possibly pass client in logout request

honzapatCZ:
## Related issue
#2468 

## Proposed changes
I haven't studied the api, but hopefully just extending the response model would work

## Checklist

      the changes.
      works.

### DIFF
--- a/internal/httpclient/models/logout_request.go
+++ b/internal/httpclient/models/logout_request.go
@@ -6,8 +6,7 @@ package models
 // Editing this file might prove futile when you re-run the swagger generate command
 
 import (
-	"context"
-
+	"github.com/go-openapi/errors"
 	"github.com/go-openapi/strfmt"
 	"github.com/go-openapi/swag"
 )
@@ -21,14 +20,14 @@ type LogoutRequest struct {
 	// identify the session.
 	Challenge string `json:"challenge,omitempty"`
 
+	// client
+	Client *OAuth2Client `json:"client,omitempty"`
+
 	// RequestURL is the original Logout URL requested.
 	RequestURL string `json:"request_url,omitempty"`
 
 	// RPInitiated is set to true if the request was initiated by a Relying Party (RP), also known as an OAuth 2.0 Client.
 	RpInitiated bool `json:"rp_initiated,omitempty"`
-	
-	// client, only set if rp_initiated
-	Client *OAuth2Client `json:"client,omitempty"`
 
 	// SessionID is the login session ID that was requested to log out.
 	Sid string `json:"sid,omitempty"`
@@ -39,11 +38,33 @@ type LogoutRequest struct {
 
 // Validate validates this logout request
 func (m *LogoutRequest) Validate(formats strfmt.Registry) error {
+	var res []error
+
+	if err := m.validateClient(formats); err != nil {
+		res = append(res, err)
+	}
+
+	if len(res) > 0 {
+		return errors.CompositeValidationError(res...)
+	}
 	return nil
 }
 
-// ContextValidate validates this logout request based on context it is used
-func (m *LogoutRequest) ContextValidate(ctx context.Context, formats strfmt.Registry) error {
+func (m *LogoutRequest) validateClient(formats strfmt.Registry) error {
+
+	if swag.IsZero(m.Client) { // not required
+		return nil
+	}
+
+	if m.Client != nil {
+		if err := m.Client.Validate(formats); err != nil {
+			if ve, ok := err.(*errors.Validation); ok {
+				return ve.ValidateName("client")
+			}
+			return err
+		}
+	}
+
 	return nil
 }
 
--- a/internal/httpclient/models/logout_request.go
+++ b/internal/httpclient/models/logout_request.go
@@ -26,6 +26,9 @@ type LogoutRequest struct {
 
 	// RPInitiated is set to true if the request was initiated by a Relying Party (RP), also known as an OAuth 2.0 Client.
 	RpInitiated bool `json:"rp_initiated,omitempty"`
+	
+	// client, only set if rp_initiated
+	Client *OAuth2Client `json:"client,omitempty"`
 
 	// SessionID is the login session ID that was requested to log out.
 	Sid string `json:"sid,omitempty"`
--- a/consent/handler.go
+++ b/consent/handler.go
@@ -767,6 +767,10 @@ func (h *Handler) GetLogoutRequest(w http.ResponseWriter, r *http.Request, ps ht
 		h.r.Writer().WriteError(w, r, err)
 		return
 	}
+
+	// We do not want to share the secret so remove it.
+	request.Client.Secret = ""
+
 	if request.WasHandled {
 		h.r.Writer().WriteCode(w, r, http.StatusGone, &RequestWasHandledResponse{
 			RedirectTo: request.RequestURL,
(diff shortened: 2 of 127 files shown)


internal/httpclient/models/logout_request.go
```diff
@@ -22,6 +22,9 @@ type LogoutRequest struct {
 
 	// RPInitiated is set to true if the request was initiated by a Relying Party (RP), also known as an OAuth 2.0 Client.
 	RpInitiated bool `json:"rp_initiated,omitempty"`
+	
+	// client, only set if rp_initiated
+	Client *OAuth2Client `json:"client,omitempty"`
```
aeneasr, line 27: These files are auto-generated, you would need to add that to the actual type ( https://github.com/ory/hydra/blob/master/consent/types.go#L398 ) and of course it also needs to be set and stored!

aeneasr (changes requested):

internal/httpclient/models/logout_request.go
honzapatCZ: > These files are auto-generated, you would need to add that to the actual type ( https://github.com/ory/hydra/blob/master/consent/types.go#L398 ) and of course it also needs to be set and stored!

OK, the thing is, the client is actually passed, so ory hydra works with it, it just isnt represented in the json form, so I suppose with my latest commit it works now?

honzapatCZ: But it already is set here https://github.com/ory/hydra/blob/24f91ab7038e4c612db5af86ff0f3ee5eb03c878/consent/strategy_default.go#L914

aeneasr: Ah I see, my bad - these files are auto-generated! What you need to do is to change this line:

https://github.com/ory/hydra/blob/cf2fe0c999f3ef4971880439e5d04b2986fc890d/consent/types.go#L451

This will then update that file automatically! The we also need a test to check if the client is actually set in JSON. For example, by doing an assertion here: https://github.com/ory/hydra/blob/cf2fe0c999f3ef4971880439e5d04b2986fc890d/consent/strategy_logout_test.go#L140

honzapatCZ: > Ah I see, my bad - these files are auto-generated! What you need to do is to change this line:
> 
> https://github.com/ory/hydra/blob/cf2fe0c999f3ef4971880439e5d04b2986fc890d/consent/types.go#L451
> 
> This will then update that file automatically! The we also need a test to check if the client is actually set in JSON. For example, by doing an assertion here:
> 
> https://github.com/ory/hydra/blob/cf2fe0c999f3ef4971880439e5d04b2986fc890d/consent/strategy_logout_test.go#L140

I already altered the types.go, but where should I put the test? There's also no test that epxects rp_inititated to be true, which is only case where client would be set.

aeneasr: This test is doing rp_initiated: https://github.com/ory/hydra/blob/cf2fe0c999f3ef4971880439e5d04b2986fc890d/consent/strategy_logout_test.go#L343

You can add the assertion there!

honzapatCZ: > This test is doing rp_initiated:
> 
> https://github.com/ory/hydra/blob/cf2fe0c999f3ef4971880439e5d04b2986fc890d/consent/strategy_logout_test.go#L343
> 
> You can add the assertion there!

Where exactly? I dont see an assert for rp_initiated, so Iam not sure.

aeneasr (approved):

aeneasr (changes requested):
Thank you! This looks better! Could you please add a test to confirm that this is working? The test also must check if `client_secret` is omitted :) Thank you!

honzapatCZ:
> Thank you! This looks better! Could you please add a test to confirm that this is working? The test also must check if `client_secret` is omitted :) Thank you!

You know I did this change a week ago? Also where should I put the test, there is no test that actually checks for rp_initiated being true, so idk.

aeneasr:
Not sure I understand? I linked the place where the tests needs to be added already, it is here: 

https://github.com/ory/hydra/blob/cf2fe0c999f3ef4971880439e5d04b2986fc890d/consent/strategy_logout_test.go#L343

That test is RP Initiated, the call here

https://github.com/ory/hydra/blob/cf2fe0c999f3ef4971880439e5d04b2986fc890d/consent/strategy_logout_test.go#L366

so this call

https://github.com/ory/hydra/blob/cf2fe0c999f3ef4971880439e5d04b2986fc890d/consent/strategy_logout_test.go#L85

which ends up here

https://github.com/ory/hydra/blob/cf2fe0c999f3ef4971880439e5d04b2986fc890d/consent/strategy_logout_test.go#L140

and there you have access to the client. I linked these lines of codes already in my comments:

- https://github.com/ory/hydra/pull/2483#discussion_r619855930
- https://github.com/ory/hydra/blob/cf2fe0c999f3ef4971880439e5d04b2986fc890d/consent/strategy_logout_test.go#L343

Sorry, but without tests I can't merge things. This is security software and needs thorough testing and good code quality. I know this might be frustrating to keep working on this but it's just the way this repo needs to work to prevent security issues. For example, I suspect that your change would currently expose the `client_secret`, so we need to add a test.

Maybe I misunderstand but I believe the instructions to be correct and accurate.

honzapatCZ:
> Not sure I understand? I linked the place where the tests needs to be added already, it is here:
> 
> https://github.com/ory/hydra/blob/cf2fe0c999f3ef4971880439e5d04b2986fc890d/consent/strategy_logout_test.go#L343
> 
> That test is RP Initiated, the call here
> 
> https://github.com/ory/hydra/blob/cf2fe0c999f3ef4971880439e5d04b2986fc890d/consent/strategy_logout_test.go#L366
> 
> so this call
> 
> https://github.com/ory/hydra/blob/cf2fe0c999f3ef4971880439e5d04b2986fc890d/consent/strategy_logout_test.go#L85
> 
> which ends up here
> 
> https://github.com/ory/hydra/blob/cf2fe0c999f3ef4971880439e5d04b2986fc890d/consent/strategy_logout_test.go#L140
> 
> and there you have access to the client. I linked these lines of codes already in my comments:
> 
> * [#2483 (comment)](https://github.com/ory/hydra/pull/2483#discussion_r619855930)
> * [#2483 (comment)](https://github.com/ory/hydra/pull/2483#discussion_r619991054)
> 
> Sorry, but without tests I can't merge things. This is security software and needs thorough testing and good code quality. I know this might be frustrating to keep working on this but it's just the way this repo needs to work to prevent security issues. For example, I suspect that your change would currently expose the `client_secret`, so we need to add a test.
> 
> Maybe I misunderstand but I believe the instructions to be correct and accurate.

But that is for acceptLogoutRequest iam passing it in getLogoutRequest

aeneasr:
@honzapatCZ I updated the code to unblock you and also help understand what I meant. Check out my commit here: https://github.com/ory/hydra/pull/2483/commits/58ac49ac6ef1b60c92cffddcd826c9405fe59dd0

Hope this helps and sorry again for the frustrating contribution experience. But we got your feature merged! :) Thank you for the help

honzapatCZ:
So it's done and this can be closed?

aeneasr:
No, it's not merged yet. CI is also failing. If you have appetite to look into it, I would appreciate it!

aeneasr:
Found the issue I believe

aeneasr:
🎉 🥳

Issue #2485, opened 2021-04-22
Session duration > 30 days not supported


romanlytvyn:
**Describe the bug**

Session duration >30 days (86400 * 30) is not behaving properly.
`oauth2_authentication_session` cookie is rejected after 30 days regardless of specified expiration time.
The session cookie expiration can be set at 10 years from now, but hydra rejects it after 30 days.

**Reproducing the bug**

1. Initiate auth flow via `/oauth2/auth` endpoint.
2. Accept login request with `remember_for: 7776000` (3600 * 24 * 90) // 90 days
Cookie examination reveals that expiration time is properly set at 90 days in the future.
2. Wait 30 days (86400 * 30 seconds) or change system time to 30 days in the future.
3. Check that `oauth2_authentication_session` cookie is still present and not expired in broswer - looks good.
4. Try to perform login with `prompt=none`.
5. An error is returned `Prompt 'none' was requested, but no existing login session was found.`

**Expected behavior**

Login flow performed without any errors since session cookie is still not expired.

**Environment**

- Version: v1.10.1

**Additional context**

A bit of a code digging reveals that cookie is not returned by `CookieStore` after 30 days:
https://github.com/ory/hydra/blob/2287ac592fc5a06b1bd3e9f54340a67395452b46/consent/strategy_default.go#L112

A `CookieStore` is initalized by `NewCookieStore`:
https://github.com/ory/hydra/blob/2287ac592fc5a06b1bd3e9f54340a67395452b46/driver/registry_base.go#L240-L246

Which sets default `MaxAge` to exactly 30 days:
https://github.com/gorilla/sessions/blob/61fa50d034f99479a7de0d1c02c5e9dea5ad30cb/store.go#L50-L61
```
func NewCookieStore(keyPairs ...[]byte) *CookieStore {
	cs := &CookieStore{
		Codecs: securecookie.CodecsFromPairs(keyPairs...),
		Options: &Options{
			Path:   "/",
			MaxAge: 86400 * 30,  //  <--- here it is, default value set to 30 days
		},
	}

	cs.MaxAge(cs.Options.MaxAge)
	return cs
}
```

Which later on is used inside `SecureCookie` `Decode` method on `Get` and returns `errTimestampExpired` error if cookie is older than `MaxAge`:
https://github.com/gorilla/securecookie/blob/f37875ef1fb538320ab97fc6c9927d94c280ed5b/securecookie.go#L339
```
// t1 - cookie issue time, extracted from the cookie value
[...]

PR #2488, opened 2021-04-22
fix: CookieStore MaxAge value (#2485)

romanlytvyn:
## Related issue

Solves #2485

## Proposed changes

Set `CookieStore` `MaxAge` option to `0` during initialization.

CookieStore MaxAge is set to 86400 * 30 by default.
This prevents secure cookies retrieval with expiration > 30 days.
MaxAge: 0 disables MaxAge check by SecureCookie, thus allowing sessions lasting > 30 days.

## Checklist

      the changes.
      works.

## Further comments

`MaxAge` can be made configurable, not sure if it's useful though.
Setting `MaxAge` to any non-zero value puts a hard cap over session `remember_for` functionality.


driver/registry_base.go
```diff
@@ -239,7 +239,10 @@ func (m *RegistryBase) KeyCipher() *jwk.AEAD {
 
 func (m *RegistryBase) CookieStore() sessions.Store {
 	if m.cs == nil {
-		m.cs = sessions.NewCookieStore(m.C.GetCookieSecrets()...)
+		cs := sessions.NewCookieStore(m.C.GetCookieSecrets()...)
+		cs.MaxAge(0)
```
aeneasr, line 243: Setting max-age to zero means that the session will be deleted from the browser when you close the browser. I don't think this is the intention here?

aeneasr (commented):

driver/registry_base.go
romanlytvyn: Nope, definitely not. But the session is actually not deleted on browser restart.

In fact, setting it to zero here has no effect on client side `oauth2_authentication_session` cookie expiration date, as it's being overwritten anyway, in case user chooses to remember session:
https://github.com/ory/hydra/blob/2287ac592fc5a06b1bd3e9f54340a67395452b46/consent/strategy_default.go#L465
And the only place `MaxAge` is used at is inside secure cookie validation then:
https://github.com/gorilla/securecookie/blob/f37875ef1fb538320ab97fc6c9927d94c280ed5b/securecookie.go#L339

But `oauth2_authentication_csrf` and `oauth2_consent_csrf` will have zero epiration though, as they rely on the default value set by `CookieStore` and are not overwritten:
https://github.com/ory/hydra/blob/2287ac592fc5a06b1bd3e9f54340a67395452b46/consent/helper.go#L66
So if someone decides to resume oauth flow on browser restart, it will break it.

Should we leave `cs.MaxAge(0)` here and consider overriding default `MaxAge` for csrf cookies?
Another alternative I thought about is adding a config `serve.cookies.max_age` with default value set to `2592000` (86400 * 30) and use it here `cs.MaxAge(max_age)` to be consistent with current behavior. And then let users override it from configuration and set to any value greater than 30 days if needed. It will extend expiration of csrf cookies as well though, not sure if it can be a problem.

Or would you recommend a better approach to the issue overall?

aeneasr: Oh I see, nice find. Not sure how we can test that though but as long as e2e tests pass it should just work. I think your patch is sufficient then but I would probably link in a codedoc to this comment for context!

romanlytvyn: Great! Would you be able to help with this part or point me to where in the docs you'd like to have it mentioned?

romanlytvyn: Added a comment in the code with short summary and a reference to more detailed explanation here.

aeneasr: perfect!

Issue #593, opened 2017-10-01
sdk/go: write interfaces for APIs & responses


arekkas:
Instead of relying on swagger-codegen to generate APIs that are non-breaking, let's introduce interfaces that make sure that there are no unexpected bc breaks.

PR #2526, opened 2021-05-10
chore: integrate with fosite `v0.40` (go-jose migration)

narg95:
## Proposed changes

This PR shows the required changes to integrate Hydra with Fosite's PR  [#593](https://github.com/ory/fosite/pull/593)


x/jwt.go
```diff
@@ -0,0 +1,15 @@
+package x
+
+import (
+	"encoding/base64"
+	"strings"
+)
+
+// Decode JWT specific base64url encoding with padding stripped
+func DecodeSegment(seg string) ([]byte, error) {
```
narg95, line 9: Fosite does not offer a backwards compatible replacement of `DecodeSegment` function from `jwt-go` since it makes no use of it.  Therefore it was copied here directly from https://github.com/dgrijalva/jwt-go

narg95:
If you are fine with those changes and once fosite's PR is merged, I can update this PR so that hydra can immediately integrate with latest `fosite`

aeneasr (dismissed):

aeneasr:
Sorry, I always have to manually trigger the github actions CI pipeline to see if OIDC passes.

consent/strategy_default.go
```diff
@@ -190,13 +190,7 @@ func (s *DefaultStrategy) getIDTokenHintClaims(ctx context.Context, idTokenHint
 	} else if err != nil {
```
mitar, line 190: What about line 188?

```
if ve := new(jwtgo.ValidationError); errors.As(err, &ve)
```

I do not think this works because new `ValidationError` does not implement [`causer` interface](https://github.com/pkg/errors#retrieving-the-cause-of-an-error) anymore? So I do not think `errors.As` really works here?

narg95: `errors.As` does not depends on `causer` interface, see [source code here](https://golang.org/src/errors/wrap.go?s=2494:2537#L67). It works because `Decode` returns `ValidationError`. `errors.As` depends on `As` interface and `Unwrap` interface. It first checks if the root error can be assigned to the target interface, which is our case, then it checks if the error implements `As` interface to outsource the check, then if the error can be `Unwrapp`ed it repeats the process with the Unwrapped error.

https://play.golang.org/p/yXkD0I0fG4F

mitar: Oh, you are right, I missed that `errors.As` just calls into stdlib: https://github.com/pkg/errors/blob/master/go113.go#L31

x/jwt.go
aeneasr: :+1:

narg95:
@aeneasr once all checks pass this PR can be merge.
I will work in bringing consistency to timestamp claims in `id_token`s next week.
Enjoy your holidays  :)

aeneasr:
Thank you! I've merged the fosite PR and released it as v0.40.2

narg95:
great!
I have updated fosite's dependency to v0.40.2 and rebased on master.
If checks pass this can be merged 🤞

consent/strategy_default_test.go
```diff
@@ -27,7 +27,7 @@ import (
 	"net/http/httptest"
 	"testing"
 
-	jwtgo "github.com/dgrijalva/jwt-go"
+	jwtgo "github.com/ory/fosite/token/jwt"
 	"github.com/stretchr/testify/require"
 
 	"github.com/ory/fosite/token/jwt"
```
mitar, line 33: I think it would be nicer to have just single import here, `"github.com/ory/fosite/token/jwt"`, no need for both `"github.com/ory/fosite/token/jwt"` and `jwtgo "github.com/ory/fosite/token/jwt"`.

mitar: Also, grouping of imports is now wrong.

Issue #2513, opened 2021-05-03
Optimize janitor's slow queries on requests tables


flavioleggio:
**Problem description**

The new `hydra janitor` command introduced very important database cleanup routines in order to tackle production environments intense grow rate on tables for tokens and requests.
The queries simply delete obsolete records in `hydra_oauth2_access` and `hydra_oauth2_refresh` tables and rejected timed out requests records in `hydra_oauth2_authentication_request` and `hydra_oauth2_consent_request`. While the tokens tables `DELETE` queries look quite efficient if they run periodically on tables which are actually cleaned up and whose sizes are limited, requests tables grow indefinitely, thus making `DELETE` on that tables slower and slower every day.

**Main solution**

I think that the problem is related to the conditions in the queries, which are mainly based on the timestamp column `requested_at`, as one can see [here](https://github.com/ory/hydra/blob/master/persistence/sql/persister_consent.go#L482). If we could avoid making the `DELETE` queries directly depend on that field, we would at least lock the tables for a short amount of time. My first idea is of separating the writing part of the query from the reading part, which obviously is the main reason for the query to be slower and slower every day. We could just `SELECT` _deletable_ records with whatever condition we like and then `DELETE ... WHERE id IN (...)` using identifiers (primary keys) from the result set of the first query.

**Alternatives**

The proposed solution does not actually optimize the queries, but it prevents long time locks on that tables, which completely block hydra service for hours, as the tables grow. A real optimization would maybe require an explicit indexed column which marks obsolete records for delete, thus implementing a logical delete.

Benehiko:
I think this makes sense! Have you been able to test this manually?

flavioleggio:
About the main solution, i have not tried it yet, but I will on my development database. I will let you know as soon as i have news on this.
About the alternatives, I am pretty sure that implementing logical delete will have a massive performance gain in production environment, but I think it will have a deeper impact on coding size, as this means that we have to introduce a new column on each entity and change the database structure accordingly.
Do you think there is some other smart way to query `deletable` records using some already existing indexed column(s)? I could not come up with any solution other than adding an explicit column.

flavioleggio:
Here are the results for the `EXPLAIN` of my version of the `SELECT` (I just substituted `DELETE` with `SELECT challenge`) in my development environment, in which `hydra_oauth2_authentication_request` table has 936 records.
|select_type|table|possible_keys|key|rows|filtered|Extra|
|-----------|-----|-------------|---|----|--------|-----|
[...]

PR #2540, opened 2021-05-20
fix: improve delete queries for janitor command

flavioleggio:
## Related issue

#2513 
@aeneasr @Benehiko 

## Proposed changes

Improve delete queries by separating the data extraction from actual delete.
Extraction is made with a configurable limit, using --limit flag, then deletes are made in batch mode with a configurable batch size, using --batch-size flag.
I chose default values for limit and batch size of respectively 100000 records and 100 records.

## Checklist

      the changes.
      works.

## Further comments

I chose to remove the transaction in cleanup for login and consent requests, which I think could be part of the problem with table locking.


cmd/janitor.go
```diff
@@ -52,6 +52,8 @@ Janitor can be used in several ways.
 		RunE: cli.NewHandler().Janitor.RunE,
 		Args: cli.NewHandler().Janitor.Args,
 	}
+	cmd.Flags().Int(cli.Limit, 10000, "Limits the number of records retrieved from database for deletion.")
```
Benehiko, line 55: So this limit here is 10K instead of 100k as suggested in the comment?

persistence/sql/persister_consent.go
```diff
@@ -439,84 +439,105 @@ func (p *Persister) VerifyAndInvalidateLogoutRequest(ctx context.Context, verifi
...
-			AND requested_at < ?
-			AND requested_at < ?
-			`,
-			(&lr).TableName(),
-			(&lrh).TableName(),
-			(&cr).TableName(),
-			(&crh).TableName()),
-			time.Now().Add(-p.config.ConsentRequestMaxAge()),
-			notAfter).Exec()
+	p.r.Logger().Printf("Deleting rejected login and consent requests which are older than %s\n", notAfter)
```
Benehiko, line 456: you can remove this logger here

cmd/janitor.go
```diff
@@ -52,6 +52,8 @@ Janitor can be used in several ways.
...
+	cmd.Flags().Int(cli.BatchSize, 100, "Limits the number of records retrieved from database for deletion.")
```
Benehiko, line 56: we might have to check if these numbers are valid, not 0 or something outrageous and give a proper response to the user?

Benehiko (commented):
I think this is looking good! :) 
Great job!

cmd/janitor.go
flavioleggio: with a proper production-tier database I think this is a reasonable batch size (as it is in my company's case), but I think this may strongly depend on the database host, so maybe we should make them mandatory like --keep-if-younger?

cmd/janitor.go
Benehiko: I think you can adjust the defaults to what has worked for you

cmd/janitor.go
Benehiko: I think just set sane-defaults and keep the flags optional. We can then add some documentation to what is recommended for large databases.

You can then also do some checks to make sure whatever the user is passing won't cause issues, like passing a 0 or something too big here https://github.com/ory/hydra/blob/eddfa2d648ae94c5056d4568a34a922c8487f6be/cmd/cli/handler_janitor.go#L39

We will most likely require some tests for that, just to test if the error given is expected or not etc.
an example is to test if `--limit = 0` or if `batch-size=0` then expect some error.

cmd/janitor.go
flavioleggio: Actually, my limit was 1 million records, but it takes an hour for the select only and delete is a long process (not affecting database resources nor service though). I tought that a lower default would be more reasonable for common use cases.

Benehiko: Then we stick with the lower limits :)

flavioleggio: Well then, we just have to decide between 10k and 100k.
In my database setup a single 100 batch delete for consent (the slowest table, I guess because of many cascade deletes) takes 1 second or so. A default limit of 10k would take a total of 100 seconds, while a limit of 100k would take 1000 seconds (i.e. almost 17 minutes). We could stick with one of these (I would take the 100k limit, if I have to choose) or get a value in the middle.

flavioleggio: @aeneasr what is you opinion on this?

oauth2/handler.go
```diff
@@ -529,12 +529,12 @@ func (h *Handler) FlushHandler(w http.ResponseWriter, r *http.Request, _ httprou
 		fr.NotAfter = time.Now()
 	}
 
-	if err := h.r.OAuth2Storage().FlushInactiveAccessTokens(r.Context(), fr.NotAfter); err != nil {
+	if err := h.r.OAuth2Storage().FlushInactiveAccessTokens(r.Context(), fr.NotAfter, 1000, 100); err != nil {
```
flavioleggio, line 532: Here we should maybe add optional (?) params in FlushInactiveOAuth2TokensRequest to let the caller choose suitable values for the use case. However I think that these values would be safe default values for an http api, to keep request processing time small, which is always a good idea imo.

oauth2/fosite_store_helpers.go
```diff
@@ -443,23 +443,23 @@ func testHelperFlushTokens(x InternalRegistry, lifespan time.Duration) func(t *t
 			require.NoError(t, err)
 		}
 
-		require.NoError(t, m.FlushInactiveAccessTokens(ctx, time.Now().Add(-time.Hour*24)))
+		require.NoError(t, m.FlushInactiveAccessTokens(ctx, time.Now().Add(-time.Hour*24), 100, 10))
```
flavioleggio, line 446: we should also ensure that batch processing is working with proper tests which take into account residual records after processing delete with request, for example

oauth2/handler.go
pharman: It might be good to document that the limit should be greater than the growth rate?

flavioleggio: This is a general rule and we should also check this in handlers and give proper error. We might also add tests for this!

flavioleggio: Oh, you mean that limit should be greater than database grow rate, other than greater than batch size, now I see. Ok, we could add that also in the docs.

persistence/sql/persister_consent.go
```diff
@@ -439,84 +439,103 @@ func (p *Persister) VerifyAndInvalidateLogoutRequest(ctx context.Context, verifi
...
+	// - hydra_oauth2_authentication_request_handled does not exist (unhandled)
+	// - hydra_oauth2_consent_request_handled has valid error (rejected)
+	// AND timed-out
+	// - hydra_oauth2_consent_request.requested_at < minimum between ttl.login_consent_request and notAfter
+	q := p.Connection(ctx).RawQuery(fmt.Sprintf(`
+		SELECT %[1]s.challenge
+		FROM %[1]s
+		LEFT JOIN %[2]s ON %[1]s.challenge = %[2]s.challenge
+		WHERE (
+			(%[2]s.challenge IS NULL)
```
aeneasr, line 469: Can `challenge` be null here?

persistence/sql/persister_consent.go
```diff
@@ -439,84 +439,103 @@ func (p *Persister) VerifyAndInvalidateLogoutRequest(ctx context.Context, verifi
...
+			OR (%[2]s.error IS NOT NULL AND %[2]s.error <> '{}' AND %[2]s.error <> '')
+		)
+		AND %[1]s.requested_at < ?
+		`,
+		(&cr).TableName(),
+		(&crh).TableName()),
+		notAfter)
+
+	if err := q.Limit(limit).All(&challenges); err == sql.ErrNoRows {
```
aeneasr, line 478: If we limit we should probably order also?

persistence/sql/persister_consent.go
```diff
@@ -439,84 +439,103 @@ func (p *Persister) VerifyAndInvalidateLogoutRequest(ctx context.Context, verifi
...
-			time.Now().Add(-p.config.ConsentRequestMaxAge()),
-			notAfter).Exec()
+	// Select challenges from all authentication requests that can be safely deleted with limit
+	// where hydra_oauth2_authentication_request were unhandled or rejected, so either of these is true
+	// - hydra_oauth2_authentication_request_handled does not exist (unhandled)
+	// - hydra_oauth2_authentication_request_handled has valid error (rejected)
+	// AND timed-out
+	// - hydra_oauth2_authentication_request.requested_at < minimum between ttl.login_consent_request and notAfter
+	q = p.Connection(ctx).RawQuery(fmt.Sprintf(`
+		SELECT %[1]s.challenge
```
aeneasr, line 506: Looks like the queries are identical - could we use one const and printf to generate them instead?

persistence/sql/persister_oauth2.go
```diff
@@ -361,30 +361,75 @@ func (p *Persister) RevokeAccessToken(ctx context.Context, id string) error {
 	return p.deleteSessionByRequestID(ctx, id, sqlTableAccess)
 }
 
-func (p *Persister) FlushInactiveAccessTokens(ctx context.Context, notAfter time.Time) error {
+func (p *Persister) FlushInactiveAccessTokens(ctx context.Context, notAfter time.Time, limit int, batchSize int) error {
```
aeneasr, line 364: We want to remove this endpoint in the long run, so it is fine to not update this IMO or just have the defaults to make the API compatible. :)

aeneasr (commented):
Great job @flavioleggio ! I have some ideas here :)

oauth2/handler.go
pharman: 👍

persistence/sql/persister_consent.go
flavioleggio: Idk, order operation is really heavy. Do you see any advantage in doing so? Is it worthy?

persistence/sql/persister_oauth2.go
flavioleggio: The janitor command calls this too, did you mean [this](https://github.com/ory/hydra/blob/b5b0791f495eda98235fbd8cff79f432dfb21653/oauth2/handler.go#L521-L543)?

persistence/sql/persister_consent.go
flavioleggio: Yes, LEFT JOIN selects records that match in both tables **and** records only present in left table: in this case, column from right tables will we NULL, so that check should mean "give me unhandled requests".

persistence/sql/persister_consent.go
flavioleggio: Yeah, I think I can give that a try!

persistence/sql/persister_consent.go
aeneasr: If you order by PK it wouldn't be heavy! However, if this task is executed only in one worker, no order should be required. If it is executed in parallel, we need to order to avoid duplicate `DELETE` which might cause errors

persistence/sql/persister_consent.go
aeneasr: I see, thank you for the explanation. I'll keep this comment open so that I can re-check the SQL query when I'm back in the office to ensure nothing gets deleted that shouldn't be deleted.

persistence/sql/persister_oauth2.go
aeneasr: Oh sorry, I mistook this for the Flush HTTP endpoint. Scrap my comment!

persistence/sql/persister_consent.go
flavioleggio: Did you have a chance to check the query?

oauth2/fosite_store_helpers.go
flavioleggio: does not seem an easy task, maybe someone will help in the future

Benehiko (commented):
I think this is looking quite good! Nice job @flavioleggio :)

I just have one request on the tests, maybe there should be a test to verify setting the limit and the batch size to confirm it actually does work, meaning if we have 5 rows in the database and you limit to 2, then we should have 3 rows still in the database - assuming the rows were all invalid/expired. Since testing the batch size isn't really possible on increments, e.g. on batch size 1 we get X, on batch size 2 we get XY. I think we should just have a test setting it to 0 and another setting it to an arbitrary number to verify 0 does not delete anything and some number above 0 does delete the rows according to the limit size.

aeneasr:
Let me know when I should take a final look :)

aeneasr:

Thank you!

flavioleggio:
Looks like there are problems with the dependency check with some transitive dependencies. The command states the following:
```
pkg:golang/github.com/gobuffalo/packr@1.22.0
1 known vulnerabilities affecting installed version
```
I tried with a simple `go mod tidy` and also upgrading transitive dependencies, but that upgraded them all and caused even more errors. Can someone give me a pointer on this field?

Benehiko (commented):
nice job! @flavioleggio 
The changes make sense to me and from my side I don't have any major suggestions or change requests :)

As for the dependency issues, I'm not too sure. Maybe @aeneasr knows something about it?

oauth2/fosite_store_helpers.go
```diff
@@ -469,6 +470,40 @@ func testHelperFlushTokens(x InternalRegistry, lifespan time.Duration) func(t *t
...
+		_, err := m.GetAccessTokenSession(ctx, requests[0].ID, ds)
+		require.Error(t, err)
+		_, err = m.GetAccessTokenSession(ctx, requests[1].ID, ds)
+		require.Error(t, err)
+		_, err = m.GetAccessTokenSession(ctx, requests[2].ID, ds)
+		require.Error(t, err)
+		_, err = m.GetAccessTokenSession(ctx, requests[3].ID, ds)
+		require.NoError(t, err)
+		_, err = m.GetAccessTokenSession(ctx, requests[4].ID, ds)
+		require.NoError(t, err)
```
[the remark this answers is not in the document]
flavioleggio: Sure thing! I would also use the `limit` value in the check condition.

aeneasr:
Nice, thank you! I will review it next week together with @Benehiko :)

Benehiko (approved):
LGTM! :)

aeneasr (commented):
Just have to verify one thing, otherwise this looks good! :)

aeneasr:
Congratulations, thank you for the hard work @flavioleggio !!!

aeneasr:
Looks like we got a flaky test :/

https://app.circleci.com/pipelines/github/ory/hydra/2826/workflows/fa46fe2b-54e0-453c-a0f8-96be01aff245/jobs/32749?invite=true#step-128-49

aeneasr:
@flavioleggio would you be open to investigate the test flake? It's unfortunately plaguing almost 50% of runs with a failure

flavioleggio:
I'll take a look this weekend, hopefully. If I can't find an answer on my own I will ask for some help on slack chat, maybe, but I will be in touch.

aeneasr:
Thank you!!

PR #2547, opened 2021-05-27
closes #2311 - wellknown is missing pkce related info

### LINKED ISSUE #2311: wellknown is missing code_challenge_methods_supported
**Describe the bug**

when pkce is enabled, i would expect the wellknown to advertise it's supported methods as defined in [RFC 8414](https://tools.ietf.org/html/rfc8414#section-2)

```
   code_challenge_methods_supported
      OPTIONAL.  JSON array containing a list of Proof Key for Code
      Exchange (PKCE) [RFC7636] code challenge methods supported by this
      authorization server.  Code challenge method values are used in
      the "code_challenge_method" parameter defined in Section 4.3 of
      [RFC7636].  The valid code challenge method values are those
      registered in the IANA "PKCE Code Challenge Methods" registry
      [IANA.OAuth.Parameters].  If omitted, the authorization server
      does not support PKCE.
```

The final sentence states: **If omitted, the authorization server does not support PKCE.**

**Expected behavior**

I expect to see the following snippet in the wellknown response:
```
...
"code_challenge_methods_supported": [
  "plain",
  "S256"
],
...
```

gbolo:
## Related issue

#2311 - wellknown is missing pkce related info

## Proposed changes

adds additional parameters to wellknown response as defined in RFC 8414

## Checklist

      the changes.
      works.

### DIFF
--- a/oauth2/doc.go
+++ b/oauth2/doc.go
@@ -149,6 +149,10 @@ type WellKnown struct {
 	// the Request Object is passed by value (using the request parameter) and when it is passed by reference
 	// (using the request_uri parameter).
 	RequestObjectSigningAlgValuesSupported []string `json:"request_object_signing_alg_values_supported"`
+
+	// JSON array containing a list of Proof Key for Code Exchange (PKCE) [RFC7636] code challenge methods supported
+	// by this authorization server.
+	CodeChallengeMethodsSupported []string `json:"code_challenge_methods_supported"`
 }
 
 // swagger:parameters deleteOAuth2Token
--- a/oauth2/handler.go
+++ b/oauth2/handler.go
@@ -251,6 +251,7 @@ func (h *Handler) WellKnownHandler(w http.ResponseWriter, r *http.Request) {
 		FrontChannelLogoutSessionSupported:     true,
 		EndSessionEndpoint:                     urlx.AppendPaths(h.c.IssuerURL(), LogoutPath).String(),
 		RequestObjectSigningAlgValuesSupported: []string{"RS256", "none"},
+		CodeChallengeMethodsSupported:          []string{"plain", "S256"},
 	})
 }
 
--- a/oauth2/handler_test.go
+++ b/oauth2/handler_test.go
@@ -436,6 +436,7 @@ func TestHandlerWellKnown(t *testing.T) {
 		FrontChannelLogoutSessionSupported:     true,
 		EndSessionEndpoint:                     urlx.AppendPaths(conf.IssuerURL(), oauth2.LogoutPath).String(),
 		RequestObjectSigningAlgValuesSupported: []string{"RS256", "none"},
+		CodeChallengeMethodsSupported:          []string{"plain", "S256"},
 	}
 	var wellKnownResp oauth2.WellKnown
 	err = json.NewDecoder(res.Body).Decode(&wellKnownResp)


aeneasr:
Thank you! The idea here being that we follow https://datatracker.ietf.org/doc/html/rfc8414 ?

gbolo:
> Thank you! The idea here being that we follow https://datatracker.ietf.org/doc/html/rfc8414 ?

Hi @aeneasr. Just pointing out section-2 of the above RFC that describes the required parameter to inform the oidc client that this authorization server supports PKCE. I'm assuming that PKCE (as described in RFC7636) is already implemented. If this is true then we must advertise the "PKCE Code Challenge Methods".

```
   code_challenge_methods_supported
      OPTIONAL.  JSON array containing a list of Proof Key for Code
      Exchange (PKCE) [RFC7636] code challenge methods supported by this
      authorization server.  Code challenge method values are used in
      the "code_challenge_method" parameter defined in Section 4.3 of
      [RFC7636].  The valid code challenge method values are those
      registered in the IANA "PKCE Code Challenge Methods" registry
      [IANA.OAuth.Parameters].  If omitted, the authorization server
      does not support PKCE.
```

aeneasr:
I see, thank you! As far as I can tell, it appears that the linked RFC would assume providing this information at a different endpoint (not `/.well-known/openid-connect` but instead `/.well-known/oauth-authorization-server`).

So I think it would make sense to open a new endpoint to follow this RFC and then add the values there?

gbolo:
> I see, thank you! As far as I can tell, it appears that the linked RFC would assume providing this information at a different endpoint (not `/.well-known/openid-connect` but instead `/.well-known/oauth-authorization-server`).
> 
> So I think it would make sense to open a new endpoint to follow this RFC and then add the values there?

I believe that we can use the existing URL as many other providers do it (such as google: https://accounts.google.com/.well-known/openid-configuration)

aeneasr:
I checked the RFC and it looks to me as they strictly require the document to be discoverable at `/.well-known/oauth-authorization-server`. See also [Section 3 of RFC 8414](https://datatracker.ietf.org/doc/html/rfc8414#section-3).

However, as you said, this also appears to be exposed at `openid-configuration`. Do you know any client libraries which would respect this field at `openid-configuration`?

gbolo:
> I checked the RFC and it looks to me as they strictly require the document to be discoverable at `/.well-known/oauth-authorization-server`. See also [Section 3 of RFC 8414](https://datatracker.ietf.org/doc/html/rfc8414#section-3).
> 
> However, as you said, this also appears to be exposed at `openid-configuration`. Do you know any client libraries which would respect this field at `openid-configuration`?

Hi @aeneasr, I haven't used any OIDC sdks. However, I have written my own **for educational purposes ONLY** (https://github.com/gbolo/muggle-oidc/blob/main/oidc/discovery.go#L21) which respects this parameter at `openid-configuration`

I quickly looked around for any popular openid sdks and i found this one for js which seems to do something similar: https://github.com/panva/node-openid-client

If you would prefer to make a separate endpoint, I can try to update the PR to accomplish that

aeneasr:
As it has come up before for other endpoints (e.g. revokation) I think it would make sense to implement RFC 8414 according to spec so we can address the other things as well then :) I think that would be the most straight forward way & spec compliant. If there is indeed a popular library expecting this at openid-configuration we could still add it there (as i. this PR:) )

gbolo:
> As it has come up before for other endpoints (e.g. revokation) I think it would make sense to implement RFC 8414 according to spec so we can address the other things as well then :) I think that would be the most straight forward way & spec compliant. If there is indeed a popular library expecting this at openid-configuration we could still add it there (as i. this PR:) )

hey @aeneasr  after reading the section 3 you pointed out, it seems like we are fine keeping it the same:
```
   Some OAuth applications will choose to use the well-known URI suffix
   "openid-configuration".  As described in Section 5, despite the
   identifier "/.well-known/openid-configuration", appearing to be
   OpenID specific, its usage in this specification is actually
   referring to a general OAuth 2.0 feature that is not specific to
   OpenID Connect.
```

aeneasr:
I finally found the relevant line in the spec (RFC 8414):

> Additional authorization server metadata parameters MAY also be used.
   Some are defined by other specifications, such as OpenID Connect
   Discovery 1.0 [OpenID.Discovery].

And [OpenID Provider Metadata](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata) says

> Additional OpenID Provider Metadata parameters MAY also be used. Some are defined by other specifications, such as OpenID Connect Session Management 1.0 [OpenID.Session].

So adding RFC 8414 to OpenID PM should be allowed :) Thus not requiring a new endpoint!

Issue #2561, opened 2021-06-04
Janitor does not clean up the sessions


mohsen3:
I noticed that `janitor $DSN --requests` cleans up `hydra_oauth2_authentication_request` and `hydra_oauth2_consent_request` tables, but not `hydra_oauth2_authentication_session`. That latter one is the second largest table in our database right now. It does not seem to be a reason to keep those rows around.

**Describe the solution you'd like**

Remove the rows from `hydra_oauth2_authentication_session` that are no longer needed.

- We should not delete a row if there is a foreign key to it from either `hydra_oauth2_authentication_request` or `hydra_oauth2_consent_request`
- We should not delete a row if the session has not yet expired

**Additional context**

The two queries to clean up `hydra_oauth2_authentication_request` and `hydra_oauth2_consent_request` tables are placed [here](https://github.com/ory/hydra/blob/master/persistence/sql/persister_consent.go#L465). It seems fine to me to follow them by a new query to clean up the sessions as well:

```go
	var ls consent.LoginSession
	err = p.Connection(ctx).RawQuery(fmt.Sprintf(`
		DELETE
		FROM %[1]s
		WHERE NOT EXISTS
			(
			SELECT NULL
			FROM %[2]s
			WHERE %[2]s.login_session_id = %[1]s.id
			)
		AND NOT EXISTS
			(
			SELECT NULL
			FROM %[3]s
			WHERE %[3]s.login_session_id = %[1]s.id
			)
		AND authenticated_at < ?
		AND authenticated_at < ?
		`,
		(&ls).TableName(),
		(&lr).TableName(),
		(&cr).TableName()),
		time.Now().Add(-p.config.ConsentRequestMaxAge()),
		notAfter).Exec()
```

I think `time.Now().Add(-p.config.ConsentRequestMaxAge())` may not be the right constraint for sessions since they may be useable beyond the lifespan of consent challenges (or am I wrong?) but the rest should be fine.

mohsen3:
It seems to me that the values in `hydra_oauth2_authentication_session` are used for two different reasons:

1. When doing a logout, this is use as id_token hint. It doesn't seem to be a big deal if hydra get's a request for an already deleted row here since the user is already technically logged out.
2. When a user has `rememberFor` option set, this table is looked up to see if there is an active session with the corresponding session_id stored in cookie. If the session is not in the table, the user is required to reauthenticate. 

[...]

PR #2567, opened 2021-06-14
feat: clean up session on janitor run

mohsen3:
## Related issue

Closes https://github.com/ory/hydra/issues/2561
@aeneasr 

## Proposed changes

Clean up the `hydra_oauth2_authentication_session` table when `hydra janitor $DSN --requests` is run.

## Checklist

      the changes.
      works.

## Further comments

`hydra_oauth2_authentication_session` is quite a large table and requires clean up. It is the second largest table in our database right now.

### DIFF
--- a/persistence/sql/persister_consent.go
+++ b/persistence/sql/persister_consent.go
@@ -447,8 +447,9 @@ func (p *Persister) FlushInactiveLoginConsentRequests(ctx context.Context, notAf
 	var cr consent.ConsentRequest
 	var crh consent.HandledConsentRequest
 
-	return p.transaction(ctx, func(ctx context.Context, c *pop.Connection) error {
+	var ls consent.LoginSession
 
+	return p.transaction(ctx, func(ctx context.Context, c *pop.Connection) error {
 		// Delete all entries (and their FK)
 		// where hydra_oauth2_authentication_request were timed-out or rejected
 		// - hydra_oauth2_authentication_request_handled
@@ -515,6 +516,32 @@ func (p *Persister) FlushInactiveLoginConsentRequests(ctx context.Context, notAf
 				(&cr).TableName(),
 				(&crh).TableName()),
 			time.Now().Add(-p.config.ConsentRequestMaxAge()),
+			notAfter).Exec()
+
+		if err != nil {
+			return sqlcon.HandleError(err)
+		}
+
+		err = p.Connection(ctx).RawQuery(fmt.Sprintf(`
+			DELETE
+			FROM %[1]s
+			WHERE NOT EXISTS
+				(
+				SELECT NULL
+				FROM %[2]s
+				WHERE %[2]s.login_session_id = %[1]s.id
+				)
+			AND NOT EXISTS
+				(
+				SELECT NULL
+				FROM %[3]s
+				WHERE %[3]s.login_session_id = %[1]s.id
+				)
+			AND authenticated_at < ?
+			`,
+			(&ls).TableName(),
+			(&lr).TableName(),
+			(&cr).TableName()),
 			notAfter).Exec()
 
 		return sqlcon.HandleError(err)


aeneasr:
Thank you for the PR! The checkbox `I have added tests that prove my fix is effective or that my feature
works.` is checked but I can't seem to find any tests ;)

@Benehiko can probably assist you to find the right place to add tests!

persistence/sql/persister_consent.go
```diff
@@ -517,6 +518,32 @@ func (p *Persister) FlushInactiveLoginConsentRequests(ctx context.Context, notAf
...
+				SELECT NULL
+				FROM %[3]s
+				WHERE %[3]s.login_session_id = %[1]s.id
+				)
+			AND authenticated_at < ?
+			`,
+			(&ls).TableName(),
+			(&lr).TableName(),
+			(&cr).TableName()),
+			notAfter).Exec()
```
aeneasr, line 545: I think this is problematic because the `notAfter` has had a different context before. Now, this could remove sessions where users are signed in and that are not expired. I think not having any references is already a good indicator if this should be removed or not but I think we probably need to move this into it's own function and then add here

https://github.com/ory/hydra/blob/be8de37caaf8060cc7e56ccc45ab7d4df7a75d28/cmd/cli/handler_janitor.go#L131-L136

as an additional case.

aeneasr (commented):
Sorry for the long review times, kinda hectic at the moment :/

Benehiko:
Hi @mohsen3 

This looks great! I believe since you are deleting from a new table you would have to write some additional tests to verify if the rows are being deleted or kept in `hydra_oauth2_authentication_session`. 

You can add your tests inside `TestJanitorHandler_PurgeLoginConsent`,  `TestJanitorHandler_PurgeLoginConsentNotAfter` here [handler_janitor_test.go](https://github.com/ory/hydra/blob/f039ebbdf315715deb44fc20fb3fdef3f4fa7b51/cmd/cli/handler_janitor_test.go#L92-L193)

As well as the helper tests inside [janitor_test_helper.go](https://github.com/ory/hydra/blob/e3d9158aebb750386c4dd2ebed0dfdaf5b374805/internal/testhelpers/janitor_test_helper.go#L36)

I believe it might be better to add a new flag to this routine since this is deleting a different aspect of the database which might not always need to be run with `--requests`. So in my opinion I think you would need to add a flag called `--sessions` and then have its own purge routines, which means adding your own test functions following the same pattern as the current janitor tests.
You would need to add a cli flag [here](https://github.com/ory/hydra/blob/f039ebbdf315715deb44fc20fb3fdef3f4fa7b51/cmd/cli/handler_janitor.go#L27-L28) and the function call [here](https://github.com/ory/hydra/blob/f039ebbdf315715deb44fc20fb3fdef3f4fa7b51/cmd/cli/handler_janitor.go#L114-L138) as well as the routine [here](https://github.com/ory/hydra/blob/f039ebbdf315715deb44fc20fb3fdef3f4fa7b51/x/fosite_storer.go#L43-L50)

persistence/sql/persister_consent.go
mohsen3: @aeneasr You are right! The `NOT EXISTS` conditions should be sufficient. The `notAfter` condition could be an extra peace of mind not to delete unintended records. It also could potentially help with the performance of the query since the engine may require to do fewer row scans this way (I didn't do any sort of benchmarking, I am just guessing).

I can remove it if you think that's misleading.

mohsen3:
> Hi @mohsen3
> 
> This looks great! I believe since you are deleting from a new table you would have to write some additional tests to verify if the rows are being deleted or kept in `hydra_oauth2_authentication_session`.
> 
> You can add your tests inside `TestJanitorHandler_PurgeLoginConsent`, `TestJanitorHandler_PurgeLoginConsentNotAfter` here [handler_janitor_test.go](https://github.com/ory/hydra/blob/f039ebbdf315715deb44fc20fb3fdef3f4fa7b51/cmd/cli/handler_janitor_test.go#L92-L193)
> 
> As well as the helper tests inside [janitor_test_helper.go](https://github.com/ory/hydra/blob/e3d9158aebb750386c4dd2ebed0dfdaf5b374805/internal/testhelpers/janitor_test_helper.go#L36)
> 
> I believe it might be better to add a new flag to this routine since this is deleting a different aspect of the database which might not always need to be run with `--requests`. So in my opinion I think you would need to add a flag called `--sessions` and then have its own purge routines, which means adding your own test functions following the same pattern as the current janitor tests.
> You would need to add a cli flag [here](https://github.com/ory/hydra/blob/f039ebbdf315715deb44fc20fb3fdef3f4fa7b51/cmd/cli/handler_janitor.go#L27-L28) and the function call [here](https://github.com/ory/hydra/blob/f039ebbdf315715deb44fc20fb3fdef3f4fa7b51/cmd/cli/handler_janitor.go#L114-L138) as well as the routine [here](https://github.com/ory/hydra/blob/f039ebbdf315715deb44fc20fb3fdef3f4fa7b51/x/fosite_storer.go#L43-L50)

Hi @Benehiko,

Please bear with me since this is my first contribution to this repo :)

1. I'll add more tests
2. I am not sure if adding an extra `--sessions` switch is very useful (or at least we in our service don't have a use for it). The `--request` switch is already deleting consent and session requests. So these sessions are just dangling around with no use. Also, deleting sessions and keeping session/consent requests cannot happen since it has foreign keys that cascades deletes.

Do you still think having a separate switch is helpful? We can start from a single switch and separate it later if we found a use for it?

Benehiko:
> > Hi @mohsen3
> > 
> > This looks great! I believe since you are deleting from a new table you would have to write some additional tests to verify if the rows are being deleted or kept in `hydra_oauth2_authentication_session`.
> > 
> > You can add your tests inside `TestJanitorHandler_PurgeLoginConsent`, `TestJanitorHandler_PurgeLoginConsentNotAfter` here [handler_janitor_test.go](https://github.com/ory/hydra/blob/f039ebbdf315715deb44fc20fb3fdef3f4fa7b51/cmd/cli/handler_janitor_test.go#L92-L193)
> > 
> > As well as the helper tests inside [janitor_test_helper.go](https://github.com/ory/hydra/blob/e3d9158aebb750386c4dd2ebed0dfdaf5b374805/internal/testhelpers/janitor_test_helper.go#L36)
> > 
> > I believe it might be better to add a new flag to this routine since this is deleting a different aspect of the database which might not always need to be run with `--requests`. So in my opinion I think you would need to add a flag called `--sessions` and then have its own purge routines, which means adding your own test functions following the same pattern as the current janitor tests.
> > You would need to add a cli flag [here](https://github.com/ory/hydra/blob/f039ebbdf315715deb44fc20fb3fdef3f4fa7b51/cmd/cli/handler_janitor.go#L27-L28) and the function call [here](https://github.com/ory/hydra/blob/f039ebbdf315715deb44fc20fb3fdef3f4fa7b51/cmd/cli/handler_janitor.go#L114-L138) as well as the routine [here](https://github.com/ory/hydra/blob/f039ebbdf315715deb44fc20fb3fdef3f4fa7b51/x/fosite_storer.go#L43-L50)
> 
> Hi @Benehiko,
> 
> Please bear with me since this is my first contribution to this repo :)
> 
> 1. I'll add more tests
> 2. I am not sure if adding an extra `--sessions` switch is very useful (or at least we in our service don't have a use for it). The `--request` switch is already deleting consent and session requests. So these sessions are just dangling around with no use. Also, deleting sessions and keeping session/consent requests cannot happen since it has foreign keys that cascades deletes.
> 
> Do you still think having a separate switch is helpful? We can start from a single switch and separate it later if we found a use for it?

That's no problem just ping me if you need help :)

I see, yes you are right! Let's keep it then under the same switch `--requests`

aeneasr:

Thank you!

Issue #2549, opened 2021-05-31
Separate client management API from the API used for dynamic client registration and configuration.


dadrus:
As of today the endpoints for the dynamic client registration and configuration are implemented as administrative end points and not as special purpose public ones. This has at least the following downsides:
* It is not possible to store additional information about the client, which shall not be available to the client itself. Best example is configuring per client valid ttls for the issued tokens as discussed in #1529.
* It is not possible to define which scopes a "dynamic" client is allowed to use and which not.
* Even the authentication of a "dynamic" client could happen in an infrastructure component, like a proxy, or an API-Gateway, it is very hard to do a proper authorization of requests to the configuration endpoint
* One is limited in products to be able to expose currently available `/clients` endpoints to support dynamic client registration and configuration.

**Describe the solution you'd like**

A pure admin API for client management purposes and public endpoints for dynamic client registration and configuration purposes. Configuration of requirements for the clients, which will register dynamically, shall be possible as well. The API for the public endpoints shall support proper authentication and authorization.

fjvierap:
@dadrus @aeneasr a new PR has been created for this... Please review when you have sometime

PR #2568, opened 2021-06-15
feat: Public api for OIDC dynamic client registration

fjvierap:
## Related issue

#2549

## Proposed changes

Create a separate endpoint for create, get, update and delete clients form public api. The new endpoint will only be available when `SERVE_PUBLIC_ALLOW_DYNAMIC_REGISTRATION` is set to true or when `allow_dynamic_registration` configuration is set to true.

## Checklist

- [ x] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [ x] I have read the [security policy](../security/policy).
- [ x] I confirm that this pull request does not address a security
      the changes.
- [x ] I have added tests that prove my fix is effective or that my feature
      works.
- [x ] I have added or changed [the documentation](docs/docs).

## Further comments

Please let me know if something else needs to be done in order to build successfully without changes mentioned above.


aeneasr:
Ready for 👀 ? :) If so you can always request a review from me so it shows up on my radar :)

fjvierap:
> Ready for 👀 ? :) If so you can always request a review from me so it shows up on my radar :)

Yes it is ready for review ;-)... Please check when you have some time

client/handler.go
```diff
@@ -221,6 +297,64 @@ func (h *Handler) updateClient(ctx context.Context, c *Client) error {
...
+	if err := json.NewDecoder(r.Body).Decode(&c); err != nil {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(err))
+		return
+	}
+
+	var secret string
+	if len(c.Secret) > 0 {
+		secret = c.Secret
+	}
+	if _, err := h.r.ClientManager().Authenticate(r.Context(), ps.ByName("id"), []byte(secret)); err != nil {
```
aeneasr, line 331: This will not allow us to update the client secret because `authenticate` will always fail. I guess it would be more common to require Basic Authorization or an access token with `sub === c.ID` here instead.

client/handler.go
```diff
@@ -122,6 +134,70 @@ func (h *Handler) Create(w http.ResponseWriter, r *http.Request, _ httprouter.Pa
...
+//
+//     Responses:
+//       201: oAuth2Client
+//       400: genericError
+//       409: genericError
+//       500: genericError
+func (h *Handler) CreateDynamicRegistration(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	var c Client
+
+	if err := json.NewDecoder(r.Body).Decode(&c); err != nil {
```
aeneasr, line 162: Looks like a lot of code is copy & paste. Could you please reuse the existing logic by extracting it into common functions or something similar?

client/handler.go
```diff
@@ -307,6 +441,64 @@ func (h *Handler) Get(w http.ResponseWriter, r *http.Request, ps httprouter.Para
...
+//
+//     Schemes: http, https
+//
+//     Responses:
+//       200: oAuth2Client
+//       401: genericError
+//       500: genericError
+func (h *Handler) GetDynamicRegistration(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	var id = ps.ByName("id")
+	var secret = r.URL.Query()["secret"][0]
```
aeneasr, line 467: Having passwords or secrets in the URL query is bad security practice as it will be exposed in e.g. HTTP logs. Please use Basic Authorization instead

client/handler.go
```diff
@@ -337,3 +529,42 @@ func (h *Handler) Delete(w http.ResponseWriter, r *http.Request, ps httprouter.P
...
+//
+//     Schemes: http, https
+//
+//     Responses:
+//       204: emptyResponse
+//       404: genericError
+//       500: genericError
+func (h *Handler) DeleteDynamicRegistration(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	var id = ps.ByName("id")
+	var secret = r.URL.Query()["secret"][0]
```
aeneasr, line 555: Use Basic Authorization here also

client/validator.go
```diff
@@ -186,6 +188,33 @@ func (v *Validator) Validate(c *Client) error {
 	return nil
 }
 
+func (v *Validator) ValidateDynamicRegistration(c *Client) error {
+
+	if c.Metadata != nil {
```
aeneasr, line 193: I think we should just disallow any metadata

client/validator.go
```diff
@@ -186,6 +188,33 @@ func (v *Validator) Validate(c *Client) error {
...
+		}
+
+		_, ok = jsonParsed.Path("id_token_ttl").Data().(float64)
+		if ok {
+			return errorsx.WithStack(ErrInvalidClientMetadata.
+				WithHint(`id_token_ttl cannot be set for dynamic client registration'`),
+			)
+		}
+	}
+	return nil
```
aeneasr, line 215:
```suggestion
	return v.Validate(c)
```

client/handler.go
```diff
@@ -122,6 +134,70 @@ func (h *Handler) Create(w http.ResponseWriter, r *http.Request, _ httprouter.Pa
 	h.r.Writer().WriteCreated(w, r, ClientsHandlerPath+"/"+c.GetID(), &c)
 }
 
+// swagger:route POST /dyn-clients public createDynOAuth2Client
```
aeneasr, line 137: OIDC specifies this as ` /connect/register` in its non-normatize examples. I think we should follow that approach

https://openid.net/specs/openid-connect-registration-1_0.html

```suggestion
// swagger:route POST /connect/register public createOAuth2ClientPublic
```

client/handler.go
```diff
@@ -221,6 +297,64 @@ func (h *Handler) updateClient(ctx context.Context, c *Client) error {
 	return nil
 }
 
+// swagger:route PUT /dyn-clients/{id} public updateDynOAuth2Client
```
aeneasr, line 300: Make the client ID part of the query string as explained shown here: https://openid.net/specs/openid-connect-registration-1_0.html#ReadRequest

```suggestion
// swagger:route PUT /connect/register public updateOAuth2ClientPublic
```

client/handler.go
```diff
@@ -307,6 +441,64 @@ func (h *Handler) Get(w http.ResponseWriter, r *http.Request, ps httprouter.Para
 	h.r.Writer().Write(w, r, c)
 }
 
+// swagger:route GET /dyn-clients/{id} public getDynOAuth2Client
```
aeneasr, line 444: Make the client ID part of the query string as explained shown here: https://openid.net/specs/openid-connect-registration-1_0.html#ReadRequest

```suggestion
// swagger:route GET /connect/register public getOAuth2ClientPublic
```

client/handler.go
```diff
@@ -122,6 +134,70 @@ func (h *Handler) Create(w http.ResponseWriter, r *http.Request, _ httprouter.Pa
...
+//     Schemes: http, https
+//
+//     Responses:
+//       201: oAuth2Client
+//       400: genericError
+//       409: genericError
+//       500: genericError
+func (h *Handler) CreateDynamicRegistration(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	var c Client
+
```
aeneasr, line 161: The OIDC spec requires authorization to perform dynamic client registration. It's not mandatory, but having this endpoint public could lead to DoS attacks on the DB by creating thousands or millions of clients.

Any ideas how we could/should protect this endpoint?

Also, in the OIDC spec they return a `registration_access_token` which can then be used to perform updates/reads on the client.

https://openid.net/specs/openid-connect-registration-1_0.html#RegistrationResponse

aeneasr (commented):
Thank you! This is a great starting point! I have some ideas how to improve the PR.

One thing noteworthy is the question if we want to implement [OpenID Connect Dynamic Client Registration](https://openid.net/specs/openid-connect-registration-1_0.html) OR [OAuth2 Dynamic Registration](https://datatracker.ietf.org/doc/html/rfc7591) or both. Depending on that, the authentication mechanisms and payloads might change.

I think it makes more sense for us to implement the OIDC spec because that's what we've implemented already, plus we can easily certify it.

client/handler.go
```diff
@@ -307,6 +441,64 @@ func (h *Handler) Get(w http.ResponseWriter, r *http.Request, ps httprouter.Para
...
+	c, err := h.r.ClientManager().Authenticate(r.Context(), id, []byte(secret))
+	if err != nil {
+		err = herodot.ErrUnauthorized.WithReason("The requested OAuth 2.0 client does not exist or you did not provide the necessary credentials")
+		h.r.Writer().WriteError(w, r, err)
+		return
+	}
+
+	c.Secret = ""
+	if c.Metadata != nil {
```
aeneasr, line 477: Metadata has a different meaning. It's purpose is not to store configuration for the client, but to store any arbitrary data. Adding `access_token_ttl` here is not metadata, imo, but configuration. Therefore I think that it would be better to have, for example, on the admin port:

```
POST /clients/<id>/settings
```

which would allow to change the config of clients - so for example the access token TTL.

aeneasr: Repurposing this field for some other functionality makes Ory Hydra vulnerable. Say you have Hydra 1.10 running. Someone scans you and sees that you use Ory Hydra 1.10 and have this endpoint exposed. He/she adds config options to the client in the metadata field. One day, you upgrade to Ory Hydra 1.11. Now, the attacker was able to change configuration which he was not supposed to change.

Therefore, this has to go somewhere else! It's generally a good idea to never use untrusted data as trusted data. Validation might help but it often fails to cover all areas :)

client/handler.go
fjvierap: @aeneasr  regarding how to prevent DoS attacks...  What I have seem in other situations is to rate limit the endpoint and when it is exceed the all next request return an error response during some time. Usually this is done in a higher level (Load balancer, api gateway or some routing engines) where the can implement this reate limit based on the referrer to do not completely block the endpoint for all users.

I think we can implement something similar. Something like add 2 configuration variables.. 1 For the rate limit and another for the timeframe to wait between the service is blocked until it accept requests again. Makes sense for you?

client/handler.go
fjvierap: I understand.....  How about have and `access_token_ttl` or `access_token_ttl` as specific fields in the root level? As now we will different endpoints for Dynamic Client registration and the `/client `ones in the admin API. Then we can have different client representations and we can include those `access_token_ttl` or `access_token_ttl` only in the one for admin API... Then hydra master can keep that configuration in sync... Does it make sense?

dadrus: Javier, I'm somehow lost by your question:
> Separate the client representation for Dynamic Client registration and the admin api maybe can be done in separate PR?

client/handler.go
aeneasr: Rate limiting is a difficult approach because it usually works on an IP basis and would require to be set up for every hydra installation as this endpoint is now exposed publicly. Alternatively, we could disable this endpoint per default and enable it per config which would mean that people have to add proper authorization themselves.

client/handler.go
fjvierap: @dadrus I mean to have `access_token_ttl` or `access_token_ttl` as dedicated fields fields for the client entity and not as part of metadata...

For this PR we will separate the public and admin endpoints... making public ones for Dynamic Client Registration to do no accept any metadata.

Does it make sense?

client/handler.go
fjvierap: @aeneasr Great... We have already a `SERVE_PUBLIC_ALLOW_DYNAMIC_REGISTRATION` configuration. Which is false be default.

fjvierap:
@aeneasr  I added the adjustments required. Please have a look when you can ;)

aeneasr (commented):
Thank you! A few more things but this is definitely going in the right direction :)

Sorry for being so slow with reviews these past days...

fjvierap:
> Thank you! A few more things but this is definitely going in the right direction :)
> 
> Sorry for being so slow with reviews these past days...

@aeneasr  I did mentioned changes :) Please review when you have some time

aeneasr (commented):
Oh man, I had this review started but apparently never clicked on submit. Sorry :(

aeneasr:
Could you please enable [write access for maintainers](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)? 🤔 

Thank you! 🙏

aeneasr:
I think one problem also is that this PR is against your fork's master branch, maybe that's why  I get push denied. Do you have some branch protection enabled in your fork?

```

To github.com:fjvierap/hydra.git
 ! [remote rejected]     fjvierap/master -> fjvierap/master (permission denied)
error: Fehler beim Versenden einiger Referenzen nach 'github.com:fjvierap/hydra.git'
```

fjvierap:
@aeneasr I did the mentioned changed and merger ory/hydra master into my repo... But It fails because a [vulnerability issue](https://app.circleci.com/pipelines/github/ory/hydra/2741/workflows/10e2e65d-aa05-4f4c-babe-56041d0aa495/jobs/32504) It also fails in my local in ory hydra repository but it worked in the [pipeline](https://app.circleci.com/pipelines/github/ory/hydra/2733/workflows/fac90d82-cd0f-441e-81a1-3cd47478cc0c/jobs/32482/parallel-runs/0/steps/0-110)

I think this [issue](https://ossindex.sonatype.org/vulnerability/1a11ab24-7bd1-4056-8fb0-450a7e4904b6?component-type=golang&component-name=github.com%2Fgobuffalo%2Fpackr&utm_source=nancy-client&utm_medium=integration&utm_content=1.0.22) has been fixed in https://github.com/gobuffalo/packr v1.26.0 I will create and additional PR in https://github.com/ory/x to upgrade to v1.26.0 and then adjust this one.

fjvierap:
@aeneasr Can you please review when you have some time?

aeneasr:
Hello! Sorry again for the long review time - I did not have any time to look into open source the past 10 days. I will be able to review this next week. My goal is to merge it by then!

aeneasr:
It looks like my changes to the PR got lost. Have you tried enabling write access for me?

https://github.com/ory/hydra/pull/2568#issuecomment-879726198 

I will have to redo all the changes again but first we need to ensure that this is working. :/

aeneasr:
Looks like I still don't have write rights :/ I will have to figure out another way to complete and review this PR.

aeneasr:
@dadrus @fjvierap how can we proceed with this? Unfortunately, I am still lacking the rights to push to the repository which would greatly help in merging this PR.

fjvierap:
> @dadrus @fjvierap how can we proceed with this? Unfortunately, I am still lacking the rights to push to the repository which would greatly help in merging this PR.

Hi @aeneasr I checked you have access to this repo... So I am wandering what the problem is..
Screen Shot 2021-09-23 at 13 15 54

Screen Shot 2021-09-23 at 13 19 20

BTW I merged from hydra repo into for fork repository

raman-nbg:
Somehow this PR got stuck. Is there anything where I could help to finish this?

aeneasr:
See: https://github.com/ory/hydra/pull/2909

aeneasr:
@fjvierap thank you for your hard work on this PR!

I am continuing the work in #2909 . I think it makes sense to follow OAuth2 Dynamic Client Registration Protocol as well as OpenID Connect Dynamic Client Registration as closely as possible and add a large set of tests to ensure everything works as needed. This means that there still needs to be quite a lot of work done, so it will take more time for this to get merged.

I will be closing this PR and you can follow the work in #2909

PR #2384, opened 2021-03-05
feat: Support for urn:ietf:params:oauth:grant-type:jwt-bearer grant type RFC 7523

Xopek:
## Related issue

#2229 @aeneasr 

## Proposed changes

This changes adds Support for JSON Web Token (JWT) Profile for OAuth 2.0 Authorization Grants (RFC7523). 
Users of Hydra will be able to grant permission for OAuth 2.0 client to act on behalf of some resource owner using jwt bearer assertions.

## Checklist

- [x] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [x] I have read the [security policy](../security/policy).
- [x] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security. vulnerability, I
      confirm that I got green light (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [x] I have added tests that prove my fix is effective or that my feature
      works.
- [ ] I have added or changed [the documentation](docs/docs).

PR #2625, opened 2021-07-07
feat: Hardware Security Module support

aarmam:
This change introduces support for Hardware Security Modules, a physical computing device that safeguards and manages digital keys, performs encryption and decryption functions for digital signatures, strong authentication, and other cryptographic functions.

If enabled, the Hardware Security Module is used to look up any keys. If no key is found, the software module is used as a fallback for lookup. This allows you to use the HSM for privileged keys, and the software module to manage lifecycle keys (e.g. for Token Exchange).

For more information, please [read the guide](https://www.ory.sh/hydra/docs/next/guides/hsm-support).

Thank you to [aarmam](https://github.com/aarmam) for this great contribution!

---

## Proposed changes

Hardware Security Module support for keys hydra.openid.id-token, hydra.jwt.access-token using PKCS#11 Cryptographic Token Interface Standard

## Checklist

- [x] I am following the
      the changes.
      works.

## Further comments

Related PR in [ory/fosite](https://github.com/ory/fosite/pull/611)


jwk/handler.go
```diff
@@ -91,21 +92,37 @@ func (h *Handler) WellKnown(w http.ResponseWriter, r *http.Request) {
...
-		if err != nil {
-			h.r.Writer().WriteError(w, r, err)
-			return
+		if h.c.HsmEnabled() {
+			keyPair, err := h.r.HardwareSecurityModule().FindKeyPair(nil, []byte(set))
+			if err != nil {
+				h.r.Writer().WriteError(w, r, err)
+				return
+			}
+			keyId, _ := h.r.HardwareSecurityModule().GetAttribute(keyPair, crypto11.CkaId)
```
aeneasr, line 101: Can we really ignore the error here?

jwk/jwt_strategy.go
```diff
@@ -109,49 +111,60 @@ func (j *RS256JWTStrategy) GetPublicKeyID(ctx context.Context) (string, error) {
...
-		j.Lock()
-		j.privateKey = k
-		j.RS256JWTStrategy.PrivateKey = k
-		j.Unlock()
-	}
-
-	if k, ok := public.Key.(*rsa.PublicKey); !ok {
-		return errors.New("unable to type assert key to *rsa.PublicKey")
+	if j.c.HsmEnabled() {
+		if keyPair, err := j.r.HardwareSecurityModule().FindKeyPair(nil, []byte(j.rs())); err != nil {
```
aeneasr, line 115: Same here, would it be possible to abstract this into `KeyManager`? That way we can avoid `j.c.HsmEnabled` checks wherever we need cryptographic keys

docs/docs/hsm-support.md
```diff
@@ -0,0 +1,91 @@
+---
+id: hsm
+title: Hardware Security Module support for JSON Web Key Sets
+---
+
+The PKCS#11 Cryptographic Token Interface Standard, also known as Cryptoki, is one of the Public Key Cryptography Standards developed by RSA Security. PKCS#11 defines the interface between an application and a cryptographic device.
```
aeneasr, line 6: This is great documentation, thank you!

jwk/handler.go
```diff
@@ -91,21 +92,37 @@ func (h *Handler) WellKnown(w http.ResponseWriter, r *http.Request) {
 	var jwks jose.JSONWebKeySet
 
 	for _, set := range stringslice.Unique(h.c.WellKnownKeys()) {
-		keys, err := h.r.KeyManager().GetKeySet(r.Context(), set)
-		if err != nil {
-			h.r.Writer().WriteError(w, r, err)
-			return
+		if h.c.HsmEnabled() {
+			keyPair, err := h.r.HardwareSecurityModule().FindKeyPair(nil, []byte(set))
```
aeneasr, line 96: Is it maybe possible to abstract this into `KeyManager` instead?

aeneasr (commented):
Thank you! This looks nice already. I think the three big things we need to address are:

- How do we write tests for HSM that work cross-platform
- Implementing HSM as a different `KeyManager` as opposed to having `if c.HSMEnabled` everywhere
- Which library to choose for HSM

Another topic will probably be supporting HSM from cloud vendors (e.g. https://cloud.google.com/kms/docs/hsm) as probably most companies today rely on cloud HSM versus having their own HSM deployed.

aeneasr:

Thank you!

aarmam:
> While the PR is being worked on I will mark it as a draft. That declutters our review backlog :)
Back from vacation! Working on it! :)

driver/registry_sql.go
```diff
@@ -81,6 +84,14 @@ func (m *RegistrySQL) Init(ctx context.Context) error {
 			return err
 		}
 
+		// TODO: Some keys might need option to use software key manager. Ideas how to refactor this?
+		// 	Current solution ignores HSM configuration for TLS keys and always uses software key manager for persistence.
+		if m.C.HsmEnabled() {
+			m.defaultKeyManager = hsm.NewKeyManager(m.HsmContext())
```
aarmam, line 90: I'm not sure if this is the right place to initialize hsm key manager or should key manager logic be refactored out of persistence logic. Problem here is that key manager is for all keys at once, so for example there is no option to exempt TLS keys from HSM and that might be overhead for some HSM (that do not support CKM_TLS_MASTER_KEY_DERIVE mechanism? I have to admit i'm not an expert in all possibilites of HSM) Thats why current solution ignores HSM key manager for TLS keys:
https://github.com/ory/hydra/blob/5298b1c29b0c0ab9fe149cf29057f633d271d127/cmd/server/helper_cert.go#L66-L87

jwk/handler.go
```diff
@@ -137,6 +135,12 @@ func (h *Handler) GetKey(w http.ResponseWriter, r *http.Request, ps httprouter.P
 		return
 	}
 
+	keys, err = FilterOpaqueKeys(keys)
```
aarmam, line 138: I would like to highlight, that opaque keys (private keys of assymetric key pairs and symmetric secret keys) cannot be returned like it was possible with software key store. How does this impact admin API? Only public keys of asymetric key pairs are returned. What about symmetric keys - should GetKey return error? What about GetKeySet where set can contain both symmetric/assymetric keys?

hsm/manager_hsm.go
```diff
@@ -0,0 +1,261 @@
...
+			return nil, err
+		}
+		return createKeySet(key, kid, alg, use)
+	case alg == "ES512":
+		key, err := m.GenerateECDSAKeyPairWithAttributes(publicAttrSet, privateAttrSet, elliptic.P521())
+		if err != nil {
+			return nil, err
+		}
+		return createKeySet(key, kid, alg, use)
+	// TODO: HS256, HS512. Makes sense only if shared HSM is used between hydra and authenticating application?
```
aarmam, line 72: HS256, HS512 on HSM makes sense only when HSM is shared between application. Can this be a use case? Also when I look fosite/jwt then there is only RS256JWTStrategy/ES256JWTStrategy so how are HS256, HS512 used at all?

aarmam:
> Thank you! This looks nice already. I think the three big things we need to address are:
> 
> * How do we write tests for HSM that work cross-platform
> * Implementing HSM as a different `KeyManager` as opposed to having `if c.HSMEnabled` everywhere
> * Which library to choose for HSM
> 
> Another topic will probably be supporting HSM from cloud vendors (e.g. https://cloud.google.com/kms/docs/hsm) as probably most companies today rely on cloud HSM versus having their own HSM deployed.



> Thank you! This looks nice already. I think the three big things we need to address are:
> 
> * How do we write tests for HSM that work cross-platform
> * Implementing HSM as a different `KeyManager` as opposed to having `if c.HSMEnabled` everywhere
> * Which library to choose for HSM
> 
> Another topic will probably be supporting HSM from cloud vendors (e.g. https://cloud.google.com/kms/docs/hsm) as probably most companies today rely on cloud HSM versus having their own HSM deployed.

aarmam:
> * How do we write tests for HSM that work cross-platform
Added quickstart-hsm for testing https://github.com/ory/hydra/blob/5298b1c29b0c0ab9fe149cf29057f633d271d127/Makefile#L88-L90
https://github.com/ory/hydra/blob/5298b1c29b0c0ab9fe149cf29057f633d271d127/.docker/Dockerfile-build#L17-L43
but I guess we should add circle-ci step with similar setup?

> * Implementing HSM as a different `KeyManager` as opposed to having `if c.HSMEnabled` everywhere
Implemented hsm key manager. See other comments for questions.

> * Which library to choose for HSM
Library is HSM vendor specific. So you install/setup the HSM client 
https://thalesdocs.com/gphsm/luna/7.4/docs/pci/Content/sdk/using/libraries_and_applications.htm
and point your pkcs11 conf. library to /usr/safenet/lunaclient/lib/libCryptoki2_64.so

> Another topic will probably be supporting HSM from cloud vendors (e.g. https://cloud.google.com/kms/docs/hsm) as probably most companies today rely on cloud HSM versus having their own HSM deployed.

Any cloud HSM that supports pkcs11 works exactly the same. For example https://docs.aws.amazon.com/cloudhsm/latest/userguide/pkcs11-library.html

aarmam:
@aeneasr, how can I help moving this pull request forward?

aeneasr:
Hey @aarmam - due to the security sensitive nature and my lack of knowledge around HSM I will probably need a full week to work on this. It might be helpful if we jump on a call to review it together. I am available again next week - you can ping me on Slack :)

.circleci/config.yml
```diff
@@ -72,6 +72,55 @@ jobs:
...
+      - prettier/install
+      - prettier/check
+      - prettier/install:
+          dir: docs/
+      - prettier/check:
+          dir: docs/
+      - goreleaser/install
+      - goreleaser/check
+      - golangci/install
+      - golangci/lint
```
aarmam, line 117: I guess some of these steps are duplicated in test job can be removed here

aeneasr:
@aarmam is this good for review again?

aarmam:
> @aarmam is this good for review again?
Yes

driver/registry_sql_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/ory/hydra/driver/config"
+	"github.com/ory/hydra/persistence/sql"
+	"github.com/ory/x/configx"
+	"github.com/ory/x/logrusx"
+)
+
+func TestDefaultKeyManager_HsmDisabled(t *testing.T) {
```
aeneasr, line 15: Could you also add a test for HSM enabled?

driver/registry_base.go
```diff
@@ -370,7 +379,7 @@ func (m *RegistryBase) newKeyStrategy(key string) (s jwk.JWTStrategy) {
 }
 
 func (m *RegistryBase) AccessTokenJWTStrategy() jwk.JWTStrategy {
-	if m.atjs == nil {
+	if m.atjs == nil && m.C.IsUsingJWTAsAccessTokens() {
```
aeneasr, line 382: This could panic if we request the method but forget to set `IsUsingJWTAsAccessTokens` to true. Do we need this check?

persistence/sql/persister_jwk.go
```diff
@@ -62,6 +81,30 @@ func (p *Persister) AddKeySet(ctx context.Context, set string, keys *jose.JSONWe
 	})
 }
 
+func (p *Persister) UpdateKey(ctx context.Context, set string, key *jose.JSONWebKey) error {
```
aeneasr, line 84: UpdateKey does not yet have a test.

I think it's necessary to run this command as a transaction, so that we always DELETE and INSERT the keys (commit), or do neither (rollback).

persistence/sql/persister_jwk.go
```diff
@@ -62,6 +81,30 @@ func (p *Persister) AddKeySet(ctx context.Context, set string, keys *jose.JSONWe
...
+
+	if err := p.AddKey(ctx, set, key); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (p *Persister) UpdateKeySet(ctx context.Context, set string, keySet *jose.JSONWebKeySet) error {
+	if err := p.DeleteKeySet(ctx, set); err != nil {
```
aeneasr, line 97: Similar to UpdateKeySet this needs to be in a transaction and it needs a test :)

aeneasr (changes requested):
Nice job! I still need to review JWK and HSM modules, but submitting this preliminary review right away to help move this along :)

jwk/helper.go
```diff
@@ -112,53 +82,63 @@ func First(keys []jose.JSONWebKey) *jose.JSONWebKey {
...
 			keys.Keys = append(keys.Keys, k)
 		}
 	}
+	return keys
+}
 
-	if len(keys.Keys) == 0 {
-		return nil, errors.Errorf("Unable to find key with prefix %s in JSON Web Key Set", prefix)
-	}
+func ExcludeOpaquePrivateKeys(set *jose.JSONWebKeySet) *jose.JSONWebKeySet {
```
aeneasr, line 133: Add a test for this? :)

jwk/handler.go
```diff
@@ -196,24 +197,12 @@ func (h *Handler) Create(w http.ResponseWriter, r *http.Request, ps httprouter.P
...
-	generator, found := h.r.KeyGenerators()[keyRequest.Algorithm]
-	if !found {
-		h.r.Writer().WriteErrorCode(w, r, http.StatusBadRequest, errors.Errorf("Generator %s unknown", keyRequest.Algorithm))
-		return
-	}
-
-	keys, err := generator.Generate(keyRequest.KeyID, keyRequest.Use)
-	if err != nil {
+	if keys, err := h.r.KeyManager().GenerateKeySet(r.Context(), set, keyRequest.KeyID, keyRequest.Algorithm, keyRequest.Use); err == nil {
+		keys = ExcludeOpaquePrivateKeys(keys)
```
aeneasr, line 201: What would happen if these were included?

jwk/handler_test.go
```diff
@@ -53,6 +53,9 @@ func TestHandlerWellKnown(t *testing.T) {
 	JWKPath := "/.well-known/jwks.json"
 
 	t.Run("Test_Handler_WellKnown/Run_public_key_With_public_prefix", func(t *testing.T) {
```
aeneasr, line 55: Is it possible to create a similar test with HSM enabled? Or is `Run_public_key_Without_public_prefix` testing this sufficiently?

jwk/helper.go
```diff
@@ -38,71 +38,41 @@ import (
...
-}
-
-func GetOrCreateKey(ctx context.Context, r InternalRegistry, g KeyGenerator, set, prefix string) (*jose.JSONWebKey, error) {
-	keys, err := r.KeyManager().GetKeySet(ctx, set)
+func GetOrGenerateKeys(ctx context.Context, r InternalRegistry, m Manager, set, kid, alg string) (public, private *jose.JSONWebKey, err error) {
+	keys, err := m.GetKeySet(ctx, set)
 	if errors.Is(err, x.ErrNotFound) || keys != nil && len(keys.Keys) == 0 {
 		r.Logger().Warnf("JSON Web Key Set \"%s\" does not exist yet, generating new key pair...", set)
-		keys, err = createKey(ctx, r, g, set)
+		keys, err = m.GenerateKeySet(ctx, set, kid, alg, "sig")
```
aeneasr, line 50: Does this method also store the keys? If so, it might make sense to name it: `GenerateAndPersistKeySet`

jwk/helper.go
```diff
@@ -38,71 +38,41 @@ import (
...
-
-		key, err = FindKeyByPrefix(keys, prefix)
+		keys, err = m.GenerateKeySet(ctx, set, kid, alg, "sig")
 		if err != nil {
-			return nil, err
+			return nil, nil, err
 		}
+		pubKey, _ = FindPublicKey(keys)
+		privKey, _ = FindPrivateKey(keys)
+		return pubKey, privKey, nil
```
aeneasr, line 74: Are we sure the errors from Find* can never be nil? I think we should return them if they do happen, even if it's unlikely, so e.g.:

```suggestion
		pubKey, err := FindPublicKey(keys)
		if err != nil {
			return nil, nil, err
		}
		privKey, err := FindPrivateKey(keys)
		if err != nil {
			return nil, nil, err
		}
		return pubKey, privKey, nil
```

jwk/jwt_strategy.go
```diff
@@ -114,12 +116,12 @@ func (j *RS256JWTStrategy) refresh(ctx context.Context) error {
 		return err
 	}
 
-	public, err := FindKeyByPrefix(keys, "public")
+	public, err := FindPublicKey(keys)
```
aeneasr, line 119: I really love that you removed this public private key prefix stuff. Thank you!

jwk/jwt_strategy_test.go
```diff
@@ -38,15 +38,12 @@ import (
 func TestRS256JWTStrategy(t *testing.T) {
```
aeneasr, line 38: Could you add a test for this strategy using HSM?

Makefile
```diff
@@ -85,6 +85,10 @@ e2e: node_modules test-resetdb
 quicktest:
 		go test -failfast -short -tags sqlite ./...
 
+.PHONY: quicktest-hsm
+quicktest-hsm:
```
aeneasr, line 89: This builds the docker image but does not execute tests?

.circleci/config.yml
aeneasr: Yes, we only need the things for HSM here, no need for goreleaser checks etc!

hsm/manager_hsm_test.go
```diff
@@ -0,0 +1,607 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/square/go-jose.v2"
+	"gopkg.in/square/go-jose.v2/cryptosigner"
+
+	"github.com/ory/hydra/hsm"
+	"github.com/ory/hydra/x"
+)
+
+func TestKeyManager_GenerateKeySet(t *testing.T) {
```
aeneasr, line 30: As far as I understand, these tests all use mocks. Would it be possible to test the real integration with libhsm as well?

aeneasr (changes requested):
Done! This looks amazing! There is only a few things which I noticed, and it's primarily about adding a few tests.

As far as I can tell, we do not have any tests which test the actual integration with an HSM provider (e.g. libhsm)? Would it be possible to add such an integration test?

driver/registry_sql_test.go
aarmam: Yes this test was indeed missing thanks! HsmEnabled test now tests default keymanager to be hardware key manager.

driver/registry_base.go
aarmam: Good that you noticed that change I should have highlighted that change! The reasoning behind this was that if jwt access token strategy is not used, then hydra.jwt.access-token key should not be generated. Hardware Security Module might have policy, that restricts generating new keys meaning keys are generated by HSM admin beforehand.  So if opaque access key stradegy is used we are forcing to generate (or pre generate by HSM admin) hydra.jwt.access-token even if its not used. Maybe cloud HSM provider will charge you extra for this also? So I removed this check for now but it is point to consider.

persistence/sql/persister_test.go
```diff
@@ -42,21 +43,35 @@ func TestManagers(t *testing.T) {
 		t.Run("package=consent/janitor="+k, testhelpers.JanitorTests(m.Config(), m.ConsentManager(), m.ClientManager(), m.OAuth2Storage()))
 
 		t.Run("package=jwk/manager="+k, func(t *testing.T) {
-			testGenerators := new(driver.RegistryBase).KeyGenerators()
-			for algo, testGenerator := range testGenerators {
-				t.Run("TestManagerKey", func(t *testing.T) {
-					ks, err := testGenerator.Generate("TestManagerKey", "sig")
-					require.NoError(t, err)
-
-					jwk.TestHelperManagerKey(m.KeyManager(), algo, ks, uuid.New())
```
aarmam, line 51: These tests were not actually executed and I had to change them because how Run subtest function was called. This led me to change the test to exclude HS256/HS512 key generators because TestHelperManagerKey, TestManagerKeySet were implemented for symmetric keys. Additionally I had to skip TestHelperManagerKey, TestManagerKeySet when HSM enabled because they test adding key/keysets and therefore are not aplicable for HSM.

persistence/sql/persister_jwk.go
aarmam: Tests are added to TestManagerKey/TestHelperManagerKeySet. Also see https://github.com/ory/hydra/pull/2625/files#r760934347

jwk/handler.go
aarmam: You would get "json: error calling MarshalJSON for type jose.JSONWebKey: square/go-jose: unknown key type '*cryptosigner.cryptoSigner'"
Private key is opaque and would not contain any info about private key so does it make sense to return only alg, use, key values? Or do you have idea how to indicate that it is opaque key when not using this filter?

jwk/jwt_strategy.go
aarmam: Thanks! :) It really helped to get rid of some logic errors because key pairs on hsm are always with matching id's so they cannot contain prefixes.

Makefile
aarmam: Actually it does by specifying `--target test-hsm`
`docker build --progress=plain -f .docker/Dockerfile-hsm --target test-hsm .`
that will execute https://github.com/ory/hydra/pull/2625/files#diff-ae54bef08e3587b28ad8e93eb253a9a5cd9ea6f4251977e35b88dc6b42329e25R20-R38

hsm/manager_hsm_test.go
aarmam: `hsm/manager_hsm_test.go` is a unit test that uses mock hsm context. 

https://github.com/ory/hydra/pull/2625/files#diff-309bfb3906e51c2911792e29568a83de4dea61a39e80bbcaee1aa6558b608cb1R15-R21

https://github.com/ory/hydra/blob/0f0eb71d9ecb1fa637feba573b0405e9b2f57624/hsm/hsm_mock_test.go

SoftHSM is used in tests as follows

`make quicktest` executes all tests without SoftHSM and HSM conf disabled.

`make quicktest-hsm` 

- [builds source with hsm build tag ](https://github.com/ory/hydra/pull/2625/files#diff-ae54bef08e3587b28ad8e93eb253a9a5cd9ea6f4251977e35b88dc6b42329e25R18)
- [sets hsm related env variables](https://github.com/ory/hydra/pull/2625/files#diff-ae54bef08e3587b28ad8e93eb253a9a5cd9ea6f4251977e35b88dc6b42329e25R21-R24)
- [initializes token named `hydra` that env variable `HSM_TOKEN_LABEL` references](https://github.com/ory/hydra/pull/2625/files#diff-ae54bef08e3587b28ad8e93eb253a9a5cd9ea6f4251977e35b88dc6b42329e25R26-R28) so that keypairs can be requested/generated correctly
- [and finaly executes tests](https://github.com/ory/hydra/pull/2625/files#diff-ae54bef08e3587b28ad8e93eb253a9a5cd9ea6f4251977e35b88dc6b42329e25R28)

[Also CircleCI test-hsm job executes all tests with SoftHSM and HSM configuration enabled.](https://github.com/ory/hydra/pull/2625/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R78-R113)

There are tests that are skipped when HSM is enabled because keys cannot be added/updated and only generated on HSM
https://github.com/ory/hydra/pull/2625/files#diff-de166095f49463ddaa5fae245ee0cfc49c99a28b69f619b1e1b9301327cd3a82R26-R28

https://github.com/ory/hydra/pull/2625/files#diff-558a3da4e777c67907421d3d7da4d33b537fa8fba7c600b2938bb2d39fdb59faR123-R144

https://github.com/ory/hydra/pull/2625/files#diff-acff70380aa55cc1a224211e29c8c9f530c8db2274f1b57456b793d79d7d9e6cR98-R100

https://github.com/ory/hydra/pull/2625/files#diff-acff70380aa55cc1a224211e29c8c9f530c8db2274f1b57456b793d79d7d9e6cR159-R161

https://github.com/ory/hydra/pull/2625/files#diff-3c4f36e62fc194cb1f350aa6cbbed40cf9407d560478c878d742de3f2a070854R56-R58

Or tests that had to be modified to minimally change the inital test

https://github.com/ory/hydra/pull/2625/files#diff-3c4f36e62fc194cb1f350aa6cbbed40cf9407d560478c878d742de3f2a070854R83-R85

https://github.com/ory/hydra/pull/2625/files#diff-881d683e78d6015e1c6ea6e60c21e442a5dea7e0c7ffc2cacae076e7d498e8d1R76-R79

https://github.com/ory/hydra/pull/2625/files#diff-acff70380aa55cc1a224211e29c8c9f530c8db2274f1b57456b793d79d7d9e6cR57-R59

https://github.com/ory/hydra/pull/2625/files#diff-acff70380aa55cc1a224211e29c8c9f530c8db2274f1b57456b793d79d7d9e6cR70-R75

https://github.com/ory/hydra/pull/2625/files#diff-acff70380aa55cc1a224211e29c8c9f530c8db2274f1b57456b793d79d7d9e6cR129-R133

https://github.com/ory/hydra/pull/2625/files#diff-acff70380aa55cc1a224211e29c8c9f530c8db2274f1b57456b793d79d7d9e6cR145-R149

jwk/jwt_strategy_test.go
aarmam: Test is executed using HSM when executing `make quicktest-hsm` or with CircleCI `test-hsm` job

aarmam:
@aeneasr ready for review

jwk/handler.go
aeneasr: That's good! :)

Makefile
aeneasr: Ah got it, that is some arcane Docker wizzardry :D

hsm/manager_hsm.go
```diff
@@ -0,0 +1,311 @@
...
+			return "", "", "", err
+		}
+		kid = ckaId.Value
+	}
+
+	var alg string
+	switch k := key.Public().(type) {
+	case *rsa.PublicKey:
+		alg = "RS256"
+		// TODO: Should we validate minimal key length by checking CKA_MODULUS_BITS?
```
aeneasr, line 218: @aarmam do you want to address this one? Otherwise we can remove the TODO item I think

aeneasr:
@aarmam this looks basically good to merge! However, I still have one question left. As far as I can tell, the Go code does not actually test HSM support with integration? I saw that we use quite a lot of mocking (which is awesome 😎 ) but I realized that we do not actually use the HSM library in the Go test. I think this is ok as you have covered all the cases, but I wanted to check in if this is intentional or not.

I also saw `quicktest-hsm` in the Dockerfile which appears to be using all the HSM dependencies, but it's not actually using them during testing because it is mocking the HSM context?

Another thing I noticed is that we introduce the `hsm` build flag. However, I can't seem to find any issue when we would not have that build tag available. As far as I can tell, the library `github.com/ThalesIgnite/crypto11` does not need any special build parameters as the HSM libraries are included at runtime? I'm not sure if I am correct here. Do we actually need this separate build tag?

From what I can tell, Ory Hydra would also build without that tag normally on all systems. I don't have HSM support on my mac installed yet, but it runs fine.

Otherwise, from a code perspective, this is good to go. Sorry for not reviewing earlier :( But I have time these next days to review the PR and get it merged before NYE :)

hsm/manager_hsm.go
aarmam: When HSM mode is enabled and no keys are found then RSA keys are generated with key length 4096. But if keys are generated on HSM beforehand then there is no key limit set. We could throw error here if key length is too small. I'm on vacation right now and not able to write serious code :) So I would remove this TODO right now if you think its not important or leave this in and I will do separate PR for this.

aarmam:
> @aarmam this looks basically good to merge! However, I still have one question left. As far as I can tell, the Go code does not actually test HSM support with integration? I saw that we use quite a lot of mocking (which is awesome 😎 ) but I realized that we do not actually use the HSM library in the Go test. I think this is ok as you have covered all the cases, but I wanted to check in if this is intentional or not.
> 
> I also saw `quicktest-hsm` in the Dockerfile which appears to be using all the HSM dependencies, but it's not actually using them during testing because it is mocking the HSM context?

Unit tests that use mocks are executed always (HSM enabled/disabled).  `quicktest-hsm` executes all tests in HSM mode using SoftHSM that emulates HSM. **So all tests are actually run using HSM when executed with  `quicktest-hsm` ** (that's why some test had to be modified or even disabled, because keys cannot be added to HSM only generated). **Also CircleCI HSM enabled test job was added to ensure all future features are compatible with HSM.**

> Another thing I noticed is that we introduce the `hsm` build flag. However, I can't seem to find any issue when we would not have that build tag available. As far as I can tell, the library `github.com/ThalesIgnite/crypto11` does not need any special build parameters as the HSM libraries are included at runtime? I'm not sure if I am correct here. Do we actually need this separate build tag?
> 
> From what I can tell, Ory Hydra would also build without that tag normally on all systems. I don't have HSM support on my mac installed yet, but it runs fine.

I used the cross compilation tool you recommended and it seemed that ARM based goarch targets did not compile. I might be completely wrong but It seemed so? So I had to use build tags -> https://github.com/ory/hydra/blob/2fab67a444051ca15b04b35941bec1fcb0941d91/.goreleaser.yml

I'm on vacation and might respond slowly, but I will rebase this PR later in the evening! And yes if possible lets try to merge it before NYE :) Let me know if you have more questions!

aeneasr:
Cross compile on ARM might indeed be an issue, I will check it out. Awesome! I’m also on vacation, I hope you enjoy yours and have time to spend with your family & loved ones and don’t get too busy reading GitHub 😇

aeneasr:
You're right, cross compilation does indeed not work out of the box. So yes, the HSM build tags do make sense. Unfortunately my internet is incredibly slow, but I'll try to check wether cross compile works in our cross compile image. If not, we have a small problem 😅

hsm/manager_hsm.go
aeneasr: Ok, that sounds good to me! I will create a follow up issue for it and link it here :)

aeneasr (approved):
Thank you for the clarifications and very, very hard work. I fixed the tests and ensured cross compilation works (it seems to work as far as I can tell). This is now good to merge if the CI passes!

aeneasr:
Hm, not sure what's going on, but the HSM tests fail with an unrelated test:

https://app.circleci.com/pipelines/github/ory/hydra/3327/workflows/1b5910a6-a18f-4828-863a-5f6660d3b2d0/jobs/34338?invite=true#step-110-1612

This test though passes on master and also in the main pipeline (without HSM enabled). Could you maybe take a look when you have a bit of time @aarmam ? Once resolved, we can merge this!

aarmam:
> Hm, not sure what's going on, but the HSM tests fail with an unrelated test:
> 
> https://app.circleci.com/pipelines/github/ory/hydra/3327/workflows/1b5910a6-a18f-4828-863a-5f6660d3b2d0/jobs/34338?invite=true#step-110-1612
> 
> This test though passes on master and also in the main pipeline (without HSM enabled). Could you maybe take a look when you have a bit of time @aarmam ? Once resolved, we can merge this!

These tests are related to [#2384](https://github.com/ory/hydra/pull/2384) that was just merged. I'm not sure how to resolve this right now. As I understand #2384 involves trusting public keys. 

If HSM is enabled hardware key manager is the default key manager. [Keys API](https://github.com/aarmam/hydra/blob/feature/hsm/jwk/handler.go) uses hardware key manager to Find/Generate/Delete keys, but the newly added [Trust API](https://github.com/aarmam/hydra/blob/feature/hsm/oauth2/trust/handler.go) uses software key manager. It uses [AddKey method](https://github.com/aarmam/hydra/blob/feature/hsm/persistence/sql/persister_grant_jwk.go#L29-L31) to store the public key. This means that this key is not found when using Keys API. 

There is a possibility to [store x509 certificates to HSM using crypto11](https://github.com/ThalesIgnite/crypto11/blob/d334790e12893aa2f8a2c454b16003dfd9f7d2de/certificates.go#L198) so that public key could be stored and [retrieved](https://github.com/ThalesIgnite/crypto11/blob/d334790e12893aa2f8a2c454b16003dfd9f7d2de/certificates.go#L110).  Importing public key only is not supported by crypto11 at this moment.

How many keys are expected to be trusted with #2384 - hundreds/thousands/more? 

If nr. of keys are in a range of HSM capabilities (I don't know what's reasonable) I could figure something with crypto11 import certificate functionality. 

Otherwise I'm not sure at the moment how to resolve this. And this might be problem with future features also, that need to import public keys or keypairs. I guess it would be possible to refactor code so that software/hardware key manager can be selected when accessing Keys API etc. If you have ideas please let me know!

aeneasr:
I see, I am glad we were able to catch this!

> How many keys are expected to be trusted with #2384 - hundreds/thousands/more?

Depending on the use case it could be up to millions as this feature will be used similar to SAML assertions. Basically you exchange a JWT for an access token.

> Otherwise I'm not sure at the moment how to resolve this. And this might be problem with future features also, that need to import public keys or keypairs. I guess it would be possible to refactor code so that software/hardware key manager can be selected when accessing Keys API etc. If you have ideas please let me know!

Wouldn't it suffice to use the software key manager in this scenario, and use HSM only for system-level keys?

aeneasr:
An alternative could also be to have two lookup strategies. When HSM is enabled, we try to look up in HSM first, then in software. When keys are added, we always add it to software. So it's not an either/or question, but something "on top".

aarmam:
> An alternative could also be to have two lookup strategies. When HSM is enabled, we try to look up in HSM first, then in software. When keys are added, we always add it to software. So it's not an either/or question, but something "on top".

Thanks I implemented this variant! I hope I was not too inattentive, I'm still in vacation mood :)

aeneasr (approved):

aeneasr:
Great work man!

vinckr:
Hello @aarmam 
Congrats on merging your first PR in Ory 🎉 !
Your contribution will soon be helping secure millions of identities around the globe 🌏 
As a small token of appreciation we send all our first time contributors a gift package to welcome them to the community. 
[Please drop me an email to claim your Ory swag!](mailto:vincent@ory.sh)

aarmam:
> Great work man!

Back from vacation and really nice news to kickstart the work mode!:tada::relaxed: Thank you @aeneasr and thanks for all the guidance!

Issue #1485, opened 2019-07-03
Add a way to filter/sort the list of clients


ngrigoriev:
I believe it would be a good idea to be able to specify the filter for the client ID list (List OAuth 2.0 Clients method)

Thus, I feel that sooner or later we will need to be able to search for the client by the client ID. At least for the reasons of maintenance. I expect that the client IDs in our multi-tenant system would have some sort of structure (URN-like probably) and at least the ability to search them by a prefix or mask would be great.

It could be also useful to do filtering based on:

- scopes
- redirect URIs
- owner
- client name
- client URI
- audience

And where is filtering, there is sorting - right? :)

It would be also important to add a way to get the shorter version of the client object in the response - through a filter or just "verbose=true|false".

**Solution description**

It could be:

- OData-like filter on the client ID or even some other
- custom simple query format

The reason why I like OData is because, I believe, it is possible to convert it to SQL query where clause relatively easily.

**Alternatives**

I do not think there are any alternatives except looking directly in Hydra's database, which is something I would like to avoid.

**Additional context**

I am not sure how practical is to implement it. For runtime, obviously, GET with known ID is just enough. Any sort of developer portal that integrates with Hydra should also be happy about CRUD interface already available. My concern is really about the maintenance.

I am sure this kind of feature will be more important if one day Hydra supports custom client attributes as part of the client data.

aeneasr:
Yeah I think that's generally a good idea. We're not entirely sure yet how querying should look like. I think there are some (sort of) standards for querying REST APIs. However, I also like the idea of maybe adding a GraphQL API?

ngrigoriev:
I think GrapQL's query capabilities are overkill for Hydra's relatively simple data model. This is why I was thinking that OData may be more appropriate - OData is simpler and does, I think, what is needed. It is also an old standard, rooted in OASIS (although it does smell Microsoft ;) ). There are many OData parsers out there and there are some that compile OData query and build SQL. 

But even OData - in order to implement it, one needs additional indexes in most of the cases. Which may be problematic for non-relational data stores. And those are very appropriate for using with the OAuth/OIDC server. This may potentially limit how far you can go with sorting and filtering.

[...]

PR #2637, opened 2021-07-13
feat: add owner/name filter to list clients

pike1212:
This adds client name and owner filters to the client list endpoint.

## Related issue(s)

#1485

## Checklist

- [x] I am following the
      the changes.
      works.

## Further Comments

Still need to update documentation, looking for any initial feedback first


client/client.go
```diff
@@ -200,6 +200,13 @@ type Client struct {
 	Metadata sqlxx.JSONRawMessage `json:"metadata,omitempty" db:"metadata"`
 }
 
+type ClientFilters struct {
```
aeneasr, line 203: This needs to be added to the swagger spec: 

https://github.com/ory/hydra/blob/b597c88e5ad7e6c00be352c8c641850e5e25f3dc/client/doc.go#L90

You can also probably re-use / re-purpose that struct as `ClientFilter`

persistence/sql/persister_client.go
```diff
@@ -76,13 +76,21 @@ func (p *Persister) DeleteClient(ctx context.Context, id string) error {
...
+	query := p.Connection(ctx).
+		Paginate(filters.Offset/filters.Limit+1, filters.Limit).
+		Order("id")
+
+	if filters.Name != "" {
+		query.Where("client_name = ?", filters.Name)
+	}
+	if filters.Owner != "" {
+		query.Where("owner = ?", filters.Owner)
+	}
```
aeneasr, line 91: This probably means that we need to index these fields? Otherwise this will get very slow when you have lots of clients. Then again, if we just have like 1000 clients, it probably wouldn't matter too much (maybe 5ms with index, 200ms without). What do you think?

aeneasr (changes requested):

client/client.go
pike1212: I added the name/owner params to that struct, is that all that needs done?

persistence/sql/persister_client.go
pike1212: For our use case, the index won't make much of a difference

aeneasr: Ok, we can keep the filter omitted for now and add it later.

client/client.go
aeneasr: Yes, that's all that's needed - I think this can be removed then!

aeneasr (commented):
Hey there, sorry for the slow review. Was quite a busy few days at Ory as we were working on the cloud platform. Next week I have much more time for reviews and merging things, which is also where I want to merge this PR!

client/client.go
pike1212: The ClientFilters struct is used internally to get passed to the client manager, so new filers could be easily added...

aeneasr:
This is what I had in mind :)

https://github.com/ory/hydra/pull/2637/commits/a4492df5cb62a0883014b88fd6fd34b0a69c2cdb

aeneasr (dismissed):

PR #2650, opened 2021-07-26
docs: add long flag --grant-types in 5min tutorial

Delapouite:
Hi

Short flag like `-g` are handy in an interactive context when a veteran user has to type long commands again and again.

Here, this tutorial is a entry point for someone discovering the `hydra` CLI and it's not obvious yet that it refers the to `--grant-types`.
Also it's more consistent with the following snippet on the same page that uses `--grant-types` explicitly.

Thanks


aeneasr:

- `docs/docs/path/to/file`

Delapouite:
Thanks for the hint. Code has been adjusted in the 2 locations now.

aeneasr (approved):

PR #2719, opened 2021-09-06
Updated Examples section part of the page

cherrymu:
Below are the changes made under Examples section

1. from hydraAdminUrl=http://hydra-example-admin:4445/'  to hydraAdminUrl=http://hydra-example-admin/' as we don't expose port number when using URL and ingress and will take care of the mapping the port behind scenes

2. Changed from --set 'hydra.config.secrets.system=$(LC_ALL=C tr -dc 'A-Za-z0-9' < /dev/urandom | base64 | head -c 32)' to --set 'hydra.config.secrets.system={'$(LC_ALL=C tr -dc 'A-Za-z0-9' < /dev/urandom | base64 | head -c 32)'}' as hydra.config.secrets.system expects a list of strings, so embraced with curly braces so a random list of strings will be generated.

If we don't add the above 2nd change then it will result in error as `Error: execution error at (hydra/templates/secrets.yaml:16:22): Expected hydra.config.secrets.system to be a list of strings` so in order to make the example deployment work without issues we need to execute the command with curly braces around with quote marks which depicts a list of strings and the deployment will be a successful one as tested in local environment and made changes.

### DIFF
--- a/docs/versioned_docs/version-v1.10/guides/kubernetes-helm-chart.mdx
+++ b/docs/versioned_docs/version-v1.10/guides/kubernetes-helm-chart.mdx
@@ -143,14 +143,14 @@ Let's install the Login and Consent App first
 
 ```bash
 $ helm install hydra-example-idp ory/example-idp \
-    --set 'hydraAdminUrl=http://hydra-example-admin:4445/' \
+    --set 'hydraAdminUrl=http://hydra-example-admin/' \
     --set 'hydraPublicUrl=http://public.hydra.localhost/' \
     --set 'ingress.enabled=true'
 ```
 
 with hostnames
 
-- `http://hydra-example-admin:4445/` corresponding to deployment name
+- `http://hydra-example-admin` corresponding to deployment name
   `--name hydra-example` (see next code sample) with suffix `-admin` which is
   the hostname of the ORY Hydra Admin API Service.
 - `https://public.hydra.localhost/` which is the default value for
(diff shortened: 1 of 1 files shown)


docs/versioned_docs/version-v1.10/guides/kubernetes-helm-chart.mdx
```diff
@@ -163,7 +163,7 @@ the ORY Hydra Helm Chart
 
 ```bash
 $ helm install hydra-example ory/hydra \
-    --set 'hydra.config.secrets.system=$(LC_ALL=C tr -dc 'A-Za-z0-9' < /dev/urandom | base64 | head -c 32)' \
+    --set 'hydra.config.secrets.system={'$(LC_ALL=C tr -dc 'A-Za-z0-9' < /dev/urandom | base64 | head -c 32)'}' \
```
aeneasr, line 166: Shouldn’t this be []? Could you maybe point to docs which explain this? :)

aeneasr (changes requested):

- `docs/docs/path/to/file`

docs/versioned_docs/version-v1.10/guides/kubernetes-helm-chart.mdx
cherrymu: Thanks a lot @aeneasr for your review and help. I have tested this with your suggestion [] and it errors out as earlier `"Expected hydra.config.secrets.system to be a list of strings"` .  So, the reason I added {} around the command is because Lists can be expressed by enclosing values in { and } when passing using `--set` flag in helm.  For example, --set name={a, b, c} translates to:

```
name:
  - a
  - b
  - c
```

Below is helm doc[1] I referred, suggested and tested out the changes in my local setup

[1] https://helm.sh/docs/intro/using_helm/

cherrymu:
I am pretty new to contributing and say myself as noob here, so please accept my apologies in advance if I am doing something wrong. As per your comment I have made the change at location `docs/docs/guides/kubernetes-helm-chart.mdx` and raised a new PR https://github.com/ory/hydra/pull/2720 . Please let me know if I have done something wrong and guide me :)

docs/versioned_docs/version-v1.10/guides/kubernetes-helm-chart.mdx
aeneasr: Awesome, thank you!

aeneasr:

PR #2758, opened 2021-09-20
feat: add method to detect public keys without prefixing

aeneasr:
Closes #2459


jwk/helper.go
```diff
@@ -49,13 +49,6 @@ func AsymmetricKeypair(ctx context.Context, r InternalRegistry, g KeyGenerator,
 	if err != nil {
 		return nil, nil, err
 	}
-	keys, err := r.KeyManager().GetKeySet(ctx, set)
```
aeneasr: I removed this because it served no purpose / did not make sense in this context and had no tests

jwk/jwt_strategy.go
```diff
@@ -113,7 +114,7 @@ func (j *RS256JWTStrategy) refresh(ctx context.Context) error {
 		return err
 	}
 
-	public, err := FindPublicKey(keys)
+	public, err := FindKeyByPrefix(keys, "public")
```
aeneasr: I reverted this because we still rely on the private: pattern to be available for private keys.

Issue #1831, opened 2020-04-28
Refresh tokens need a grace period to deal with network errors and similar issues


NavindrenBaskaran:
So basically, on Hydra if the refresh token is already used to request for a new set of tokens, that refresh token is invalidated and can't be used to request for a new set of tokens. It's a one time use only.

At the event, where the tokens were actually issued by Hydra but the response was not sent to the users because of some network issues, the users are generally stuck. Is there a way for us to not invalidate the refresh token immediately and have a refresh token grace period to perform a duplicate refresh requests with the same refresh token.

aeneasr:
> At the event, where the tokens were actually issued by Hydra but the response was not sent to the users because of some network issues, the users are generally stuck. Is there a way for us to not invalidate the refresh token immediately and have a refresh token grace period to perform a duplicate refresh requests with the same refresh token.

This is currently not possible, the client would need to talk to `/oauth2/auth` again. A response getting lost in the network can of course happen but is very unlikely and does not justify issuing multiple tokens as this would not only potentially exponentially explode the amount of tokens issued but also enable replay attacks.

imranismail:
@aeneasr does it help to mitigate the replay attack if we set a low default before the old token is completely unusable.

And to avoid exploding the amount of token issued, we could return the same token that was lost during the network error.

Please, just correct me if I'm wrong.

aeneasr:
> @aeneasr does it help to mitigate the replay attack if we set a low default before the old token is completely unusable.

Lowering TTL of things is typically a good guard against replayability, but doesn't resolve it completely of course :)

> And to avoid exploding the amount of token issued, we could return the same token that was lost during the network error.

While this is a valid strategy and also supported by some servers, it does not work in ORY Hydra. We do not store the full access tokens in the database but instead generate it from a strong `/dev/urandom` reader. What we store in the database is it's HMAC-SHA256 hash which we can use to look up a token and see if it still valid.

This feature prevents anyone with database access from issuing access and refresh tokens without also having access to the ORY Hydra `secrets.system`.

aeneasr:
https://github.com/ory/hydra/issues/1928#issuecomment-678119025

aeneasr:
Closed in error

mitar:
[...]

PR #2827, opened 2021-10-29
feat: Refresh token expiration window

bill-robbins-ss:
- Introduce optional configuration value `oauth2.refresh_token_rotation.grace_period` that can be set as a duration (`1m`, 
- Add a new migration: boolean `used` SQL column in `hydra_oauth2_refresh` 

When `grace_period` has positive duration the act of using a refresh token _will not invalidate_ the token and instead mark it as used in the database and set its expiration time to the UTCNOW + the value of the grace period. This will allow a client to continue to use a refresh token for the duration of the grace period. 

## Related issue(s)
#1831 
[Fosite 255](https://github.com/ory/fosite/issues/255)

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

This change includes a modification to Fosite in [PR 634](https://github.com/ory/fosite/pull/635)


scripts/create-migration.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
```
aeneasr, line 1: You can use the [ory cli](https://github.com/ory/cli) to achieve this:

```
ory dev pop migration create 
Error: accepts 2 arg(s), received 0
Usage:
  ory dev pop migration create [destination] [name] [flags]

Flags:
  -d, --dialect string   Choose a dialect from: [sqlite cockroach mysql postgres]
  -h, --help             help for create

accepts 2 arg(s), received 0
```

:)

aeneasr (commented):
Looking good :) Could you please also add some docs for this? :)

persistence/sql/persister_oauth2.go
```diff
@@ -247,6 +292,31 @@ func (p *Persister) findSessionBySignature(ctx context.Context, rawSignature str
 	})
 }
 
+func (p *Persister) findRequesterBySignature(ctx context.Context, signature string, tableName tableName) (fosite.Requester, error) {
```
aeneasr, line 295: Please add tests for this new method

persistence/sql/persister_oauth2.go
```diff
@@ -280,13 +350,27 @@ func (p *Persister) deactivateSessionByRequestID(ctx context.Context, id string,
...
 			RawQuery(
-				fmt.Sprintf("UPDATE %s SET active=false WHERE request_id=?", OAuth2RequestSQL{Table: table}.TableName()),
+				fmt.Sprintf("UPDATE %s SET active=false, used=true WHERE request_id=?", OAuth2RequestSQL{Table: table}.TableName()),
 				id,
 			).
 			Exec(),
 	)
 }
 
+func (p *Persister) getRefreshTokenUsedStatusBySignature(ctx context.Context, signature string) (bool, error) {
```
aeneasr, line 360: Please add tests for this new method

persistence/sql/persister_oauth2.go
```diff
@@ -353,8 +437,40 @@ func (p *Persister) DeletePKCERequestSession(ctx context.Context, signature stri
 	return p.deleteSessionBySignature(ctx, signature, sqlTablePKCE)
 }
 
-func (p *Persister) RevokeRefreshToken(ctx context.Context, id string) error {
-	return p.deactivateSessionByRequestID(ctx, id, sqlTableRefresh)
+func (p *Persister) RevokeRefreshToken(ctx context.Context, requestId string) error {
+	return p.deactivateSessionByRequestID(ctx, requestId, sqlTableRefresh)
+}
+
+func (p *Persister) RevokeRefreshTokenMaybeGracePeriod(ctx context.Context, requestId string, signature string) error {
```
aeneasr, line 444: Please add tests for this new functionality

aeneasr (changes requested):

scripts/create-migration.sh
bill-robbins-ss: Cool! Was not aware this existed. Removed my bash script and re-created those files w/ the CLI.

persistence/sql/persister_oauth2.go
bill-robbins-ss: Figured out how to remove this method :)

persistence/sql/persister_oauth2.go
bill-robbins-ss: this is a private method, do you have a suggestion on how I can test it ? Alternatively testing `RevokeRefreshTokenMaybeGracePeriod` will test this method indirectly.

persistence/sql/persister_oauth2.go
bill-robbins-ss: I can't figure out how to write a test in `fosite_store_helpers.go` that changes the configuration value that this depends on. Can you point me in the right direction?

docs/docs/guides/token-expiration.mdx
```diff
@@ -75,6 +75,44 @@ for refresh tokens to never expire.
...
+allows a refresh token to be reused for the duration of the grace period. Note
+that a new refresh token is still generated and sent back in the response;
+clients **must** store and use the new refresh token.
+
+**WARNING** Using the refresh token grace period is an increased security risk,
+as an intercepted refresh token may be reused by a bad actor. Use this feature
+with appropriate consideration.
+
+```
+  ## refresh_token_rotation
```
aeneasr, line 98: This is missing the top level key I think :)

```suggestion
oauth2:
  ## refresh_token_rotation
```

docs/docs/guides/token-expiration.mdx
```diff
@@ -75,6 +75,44 @@ for refresh tokens to never expire.
     #....
 ```
 
+### Refresh Token Rotation
+
+When a refresh token is used it is deactivated, which is known as Refresh Token
+Rotation. By default Hydra deactivates the refresh token it receives and issues
```
aeneasr, line 81: Maybe not that reusing the token will also cause that all tokens related to that refresh token are invalidated!

persistence/sql/migrations/20211101093405_add_refresh_token_used_flag.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE hydra_oauth2_refresh ADD used bool DEFAULT false;
```
aeneasr, line 1: I would suggest a clearer naming, such as `in_grace_period`

persistence/sql/models/models.go
```diff
@@ -0,0 +1 @@
+package models
```
aeneasr, line 1: Delete this file :)

persistence/sql/persister_oauth2.go
```diff
@@ -219,6 +240,30 @@ func (p *Persister) createSession(ctx context.Context, signature string, request
 	return nil
 }
 
+func (p *Persister) updateRefreshSession(ctx context.Context, requestId string, session fosite.Session, used bool) error {
+	_, ok := session.(*oauth2.Session)
+	if !ok && session != nil {
+		return errors.Errorf("Expected session to be of type *Session, but got: %T", session)
```
aeneasr, line 246:
```suggestion
		return errors.Errorf("expected session to be of type *oauth2.Session but got: %T", session)
```

persistence/sql/persister_oauth2.go
aeneasr: You can export it if need be :)

persistence/sql/persister_oauth2.go
```diff
@@ -353,8 +412,41 @@ func (p *Persister) DeletePKCERequestSession(ctx context.Context, signature stri
...
+	if gracePeriod <= 0 {
+		return p.RevokeRefreshToken(ctx, requestId)
+	}
+
+	var requester fosite.Requester
+	var err error
+	session := new(oauth2.Session)
+	if requester, err = p.GetRefreshTokenSession(ctx, signature, session); err != nil {
+		p.l.Errorf("signature: %s not found. grace period not applied", signature)
+		return errors.WithStack(err)
```
aeneasr, line 430: Could you add this to the error context instead? Without a request context, it will be difficult to trace the log in a noisy environment :)

persistence/sql/persister_oauth2.go
```diff
@@ -353,8 +412,41 @@ func (p *Persister) DeletePKCERequestSession(ctx context.Context, signature stri
...
+	}
+
+	var used bool
+	if used,err = p.getRefreshTokenUsedStatusBySignature(ctx, signature); err != nil {
+		p.l.Errorf("signature: %s used status not found. grace period not applied", signature)
+		return errors.WithStack(err)
```
aeneasr, line 436: Could you add this to the error context instead? Without a request context, it will be difficult to trace the log in a noisy environment :)

driver/config/provider.go
```diff
@@ -428,3 +429,7 @@ func (p *Provider) CGroupsV1AutoMaxProcsEnabled() bool {
 func (p *Provider) GrantAllClientCredentialsScopesPerDefault() bool {
 	return p.p.Bool(KeyGrantAllClientCredentialsScopesPerDefault)
 }
+
+func(p *Provider) RefreshTokenRotationGracePeriod() time.Duration {
+	return p.p.DurationF(KeyRefreshTokenRotationGracePeriod, 0)
```
aeneasr, line 434: Should we limit to prevent this form being set to e.g. 1 year?

persistence/sql/persister_oauth2.go
```diff
@@ -108,6 +104,31 @@ func (p *Persister) sqlSchemaFromRequest(rawSignature string, r fosite.Requester
 	}, nil
 }
 
+func (p *Persister) marshalSession(session fosite.Session) ([]byte, error) {
+	sessionBytes, err := json.Marshal(session)
+	if err != nil {
+		return nil, err
+	}
+
+	if sessionBytes, err = p.maybeEncryptSession(sessionBytes); err != nil {
```
aeneasr, line 113: Could you make this explicit? A simple `:=` here would disable encryption, a significant security feature. Thanks!

persistence/sql/persister_oauth2.go
```diff
@@ -353,8 +412,41 @@ func (p *Persister) DeletePKCERequestSession(ctx context.Context, signature stri
...
+		p.l.Errorf("signature: %s used status not found. grace period not applied", signature)
+		return errors.WithStack(err)
+	}
+
+	if ! used {
+		session := requester.GetSession()
+		session.SetExpiresAt(fosite.RefreshToken, time.Now().UTC().Add(gracePeriod))
+		if err = p.updateRefreshSession(ctx, requestId, session, true); err != nil {
+			p.l.Errorf("failed to update session with signature: %s", signature)
+			return errors.WithStack(err)
```
aeneasr, line 444: Same as above

persistence/sql/persister_oauth2.go
```diff
@@ -353,8 +412,41 @@ func (p *Persister) DeletePKCERequestSession(ctx context.Context, signature stri
...
+		}
+	} else {
+		p.l.Debugf("request_id: %s has already been used and is in the grace period", requestId)
```
aeneasr, line 447: This will pollute the log relatively quickly as it is being triggered on every refresh, I would suggest to remove it or maybe make it a "trace"

persistence/sql/persister_oauth2.go
```diff
@@ -280,13 +325,27 @@ func (p *Persister) deactivateSessionByRequestID(ctx context.Context, id string,
 	return sqlcon.HandleError(
 		p.Connection(ctx).
 			RawQuery(
-				fmt.Sprintf("UPDATE %s SET active=false WHERE request_id=?", OAuth2RequestSQL{Table: table}.TableName()),
+				fmt.Sprintf("UPDATE %s SET active=false, used=true WHERE request_id=?", OAuth2RequestSQL{Table: table}.TableName()),
```
aeneasr, line 328: Do we actually need `used` when we have `active` already?

persistence/sql/persister_oauth2.go
aeneasr: Hm yeah that looks a bit iffy. I think you can call `store.Config().MustSet(key, value)` but you most likely need to reset it to the original value on `t.Cleanup` so other tests don't fail.

oauth2/fosite_store_helpers.go
```diff
@@ -408,6 +410,68 @@ func testHelperRevokeAccessToken(x InternalRegistry) func(t *testing.T) {
...
+			_, err = m.GetRefreshTokenSession(ctx, refreshTokenSession, *tmpSession)
+
+			// a revoked refresh token returns an error when getting the token again
+			assert.Error(t, err)
+			assert.True(t, errors.Is(err, fosite.ErrInactiveToken))
+		})
+
+		t.Run("refresh token enters grace period when configured,", func(t *testing.T) {
+
+			/* TODO: figure out how to change config values and get a new/udpated OAuth2Storage instance
```
aeneasr, line 441: Try:

```
x.Config().MustSet(config.KeyEncryptSessionData, tc.enableSessionEncrypted)
```

persistence/sql/persister_oauth2.go
```diff
@@ -353,8 +412,41 @@ func (p *Persister) DeletePKCERequestSession(ctx context.Context, signature stri
...
+
+	var used bool
+	if used,err = p.getRefreshTokenUsedStatusBySignature(ctx, signature); err != nil {
+		p.l.Errorf("signature: %s used status not found. grace period not applied", signature)
+		return errors.WithStack(err)
+	}
+
+	if ! used {
+		session := requester.GetSession()
+		session.SetExpiresAt(fosite.RefreshToken, time.Now().UTC().Add(gracePeriod))
```
aeneasr, line 441: So to understand this a bit better - when refreshing a token we do not deactivate it (`active: false`) immediately. Instead, if grace is enabled, we extend the expiry time by X from NOW, and we also set `used: true`. The next time we refresh the token, and we are still in the refresh grace period, nothing is updated.

When we refresh the token again, and NOW + x has passed, fosite will see the key as expired, implying that it is no longer active.

I think this has some serious implications. I am not 100% any more how token reuse detection works, but I think it doesn't trigger on expired tokens, only on inactive ones. Because an expired token has not been reused, it just expired! And since it's expired, you can't use it. A used (active=false) token though has been used, thus, reuse means invalidation of all other tokens.

aeneasr (changes requested):
Great work, I've got a couple of questions and ideas :)

driver/config/provider.go
bill-robbins-ss: Yes. I consider `<=1h`  extremely reasonable, what do you think ? In our deployment I would set this to 5-10m

persistence/sql/persister_oauth2.go
bill-robbins-ss: I'm not sure what you mean here, is there code elsewhere that does this?

persistence/sql/persister_oauth2.go
bill-robbins-ss: I don't think so, `findSessionBySignature` will return errors when a session is marked inactive.

oauth2/fosite_store_helpers.go
bill-robbins-ss: `x` does not seem to have `Config()` :(

aeneasr: I need to check but you can probably do:

```
func testHelperRevokeRefreshTokenMaybeGracePeriod(x interface {
Config() ...
InternalRegistry
}) ...

persistence/sql/migrations/20211101093405_add_refresh_token_used_flag.up.sql
bill-robbins-ss: done

persistence/sql/persister_oauth2.go
bill-robbins-ss: set to trace

persistence/sql/persister_oauth2.go
bill-robbins-ss: I think you're saying
* this will work but the error will be wrong (expired not reused)
* since it isn't detected as reused, other tokens are not revoked

I mentioned in an earlier comment that setting `active=false` won't work for grace periods, since calls to `findSessionBySignature` will not succeed. To properly detect that case (`active==true, in_grace_period==true, expired`) something could be modified to deal with it, I'm worried that changing `findSessionBySignature` to throw an expired error may have undesirable repercussions. From what I can tell token validation, which includes expiration checking, is done at a different layer and is part of a Strategy. The persistence layer does not have access to token strategies.

aeneasr: That was what I was trying to say :)

The best way to prove/disprove this hypothesis is to introduce a test which covers this. If there's no regression on revokation, this should be fine! It would theoretically also be fine to revoke the token chain on expiration also, not just on reuse.

persistence/sql/persister_oauth2.go
bill-robbins-ss: @aeneasr I used reflection to update the config, and enabled the test that exercises this function

aeneasr:
Is this good for another review @bill-robbins-ss ? :)

bill-robbins-ss:
@aeneasr yes it's ready :)

aeneasr:
@bill-robbins-ss I fixed the conflicts and got parts of the CI to pass again, however there seems to be an linting issue. Would it be possible for you to take a look?

https://app.circleci.com/pipelines/github/ory/hydra/3404/workflows/bd1ecd98-8d78-4ea0-835a-5b45ff9a7aae/jobs/34581

ps: For motivation, we have merged several large-scale changes to Ory Hydra so there's a good chance this will land very quickly as well :)

bill-robbins-ss:
@aeneasr this is great, I'll be checking it out this week hopefully getting those rules to pass. Do you know how releasing will be done? We'll want to upgrade our instance to take advantage of the great new things!

aeneasr:
Once this is merged we'll do a new minor release

bill-robbins-ss:
hey @aeneasr I fixed the linting errors that came from me, but he other 3 are on code that I didn't touch. Please advise.

driver/config/provider.go
aeneasr: Done!

aeneasr (changes requested):
Thank you for the work and sorry for the late review time. I did fix some things in the PR but while working on it, I realized that there is still some way to go. In particular:

1. The new `in_grace_period` column breaks all tables that do not have this column as they all re-use the same struct to load stuff from the database
2. The feature is not end-to-end tested
3. Documentation is missing

Given the sensitivity of this feature we need to ensure that everything works end-to-end. So:

1. Refreshing a token within the grace period works
2. Revoking a token revokes also all grace tokens
3. Rotating a grace token works and does not affect other tokens
4. Token reuse still works after grace period and revokes all of those tokens also

Let me know if you need any help with this!

bill-robbins-ss:
Hey @aeneasr thank you for checking in on this. 

1. I can introduce a new struct to handle the new table schema, does that seem reasonable?
2. I would be delighted to add e2e tests, I think they belong [here](https://github.com/bill-robbins-ss/hydra/tree/refresh-token-expiration-window/test/e2e)?
3. It looks like you added documentation with [this](https://github.com/ory/docs/commit/37f172283f41dd8b14d9432d7c4dc41b36216510) change. If there is more that needs to be added just let me know where and I'll add it ! 

Also, I saw that you removed the `SetConfig` method - I introduced it for the test `refresh token enters grace period when configured`. Is that same method available somewhere else ?

aeneasr:
Thank you @bill-robbins-ss for the fast response and keeping engaged. I have this PR on my prio list now and will be more active reviewing it. You can also ping me on Slack if I am too slow :)

Regarding your questions:

> I can introduce a new struct to handle the new table schema, does that seem reasonable?

It does! I would encourage you though to maybe rebase this PR right away with https://github.com/ory/hydra/pull/2796 to avoid any big conflicts during merge.

> I would be delighted to add e2e tests, I think they belong [here](https://github.com/bill-robbins-ss/hydra/tree/refresh-token-expiration-window/test/e2e)?

Exactly, I would suggest to cover the following use cases:

1. [ ] Refreshing a refresh token without grace period (I believe we have this already)
1. [ ] Refreshing a refresh token without grace period that is expired (I believe we have this already)
2. [ ] Refreshing a refresh token with grace period
2. [ ] Refreshing a refresh token with grace period that is expired
2. [ ] Refreshing a refresh token 1 (before expiry) 2 (after expiry) 3 (after expiry) with grace period

We also need to cover revokation *and* deleting the consent:

1. [ ] Revoke a refresh token that was granted before expiry
2. [ ] Revoke consent of refresh token that was granted before expiry
1. [ ] Revoke a refresh token that was granted before expiry with grace period
2. [ ] Revoke consent of refresh token that was granted before expiry with grace period
1. [ ] Revoke a refresh token that was granted before expiry with grace period that was expired
1. [ ] Revoke consent of refresh token that was granted before expiry with grace period that was expired
1. [ ] Revoke a 1 (before expiry) 2 (after expiry) 3 (after expiry) refresh token grace period or consent
1. [ ] Revoke consent of 1 (before expiry) 2 (after expiry) 3 (after expiry) refresh token grace period or consent
2. [ ] For the token revokation and consent deletion we need to check whether all tokens have been revoked. It should not be possible to revoke just one refresh token "chain".

Checking whether we can refresh graceful tokens is important too:

1. [ ] Refresh a 1 (before expiry) 2 (after expiry) 3 (after expiry) with grace period or consent that was granted normally with grace period. We should refresh at least 2 or 3 times to ensure that the "refresh token tree" works correctly.

As well as introspection:

1. [ ] Refreshing a 1 (before expiry) 2 (after expiry) 3 (after expiry) times with grace period and introspecting their access tokens

I believe most functions are there already to test this, so it should not be too much work hopefully.

Besides e2e tests, it probably also makes sense to start with some Go test first, as that will make debugging so much easier for you!

https://github.com/ory/hydra/blob/8a966e8866d0a054f4d6cd0581db42690ba78805/oauth2/oauth2_auth_code_test.go#L277

> It looks like you added documentation with [this](https://github.com/ory/docs/commit/37f172283f41dd8b14d9432d7c4dc41b36216510) change. If there is more that needs to be added just let me know where and I'll add it !

Apologies, I got in review mode and am not fully used to the new docs system yet 😓 

> Also, I saw that you removed the SetConfig method - I introduced it for the test refresh token enters grace period when configured. Is that same method available somewhere else ?

Yes, it's now defined on the interface level so no reflection is needed!

aeneasr:
@bill-robbins-ss please let me know when you think that this is good for another review :)

yuekun0707:
Hi, any update here?  Is configuration described here https://www.ory.sh/docs/hydra/guides/token-expiration#refresh-token-rotation ready for using?

aeneasr:
Looks like that made it to prod on mistake, the feature is not yet usable!

yuekun0707:
@aeneasr  thank you for your reply! I really need feature like this. We use hydra in our project and it works well. But now we want to connect to google assistant by account linking.  There is a problem with the refresh token rotation. The error from google like this:
"Refresh token has been rotated. This is not forbidden, however we do not see much benefit that rotating the refresh token can provide but the potential problem it has. We also tried to refresh token with the old refresh token after it has been rotated. Refresh didn't work, this means partner invalidated the old refresh token right after the rotation. Partners shall only invalidate the old refresh token after seeing we use the new one to ensure we got it successfully."
So, is there any way I can try for this?

misl6:
Hi @aeneasr and @bill-robbins-ss !
As other users pointed out, this is pretty much a requirement to pass the Google Assistant Actions certification.

Is this feature something scheduled and that we may expect in a near future?
There's anything we can do on our side (as users) to help you to make that happen?

StarAurryon:
Hi @aeneasr, I am willing to continue the work on the coming months as we are currently experiencing network token loss when Windows goes to sleep with our app and also with our android / ios app.

Did @bill-robbins-ss signed the Ory Code Agreement so that I can reuse his code and pursue the implementation ?

Kind regards,

joshuaflanagan:
The issue was closed because the original fork was mistakenly deleted. I created a new fork and pushed the branch, but it would not let me push to the same branch name, so it is now at https://github.com/shipstation/hydra/tree/refresh-token-expiration-window-2.  Unfortunately, there does not appear to be a way to change the branch for this PR to reference the new branch. Either way, it looks like the code is still accessible in this PR - maybe an ory/hydra project owner can re-open the PR

PR #2836, opened 2021-11-03
feat: introduce the concept of a flow in the persistence layer

grantzvolsky:
@aeneasr This is the first step towards merging the four login and consent tables.

## Commit Message

This patch changes the internal data structure and reduces four (sort of redundant) tables into one. As part of this change, a few new tools have been added:

* Introduce the `hydra sql gen` command and a convenience Make target with autocompletion. The command reads migration templates from a source directory and produces migration files in a target directory. Its main function is to split a single source file into multiple files using split marks.

* Introduce the `hack/db-diff.sh` command to generate database schema diffs at different commits. This script is used to view and review the impact of migrations on the database schema.

BREAKING CHANGES: This patch merges four SQL Tables into a new table, deleting the old tables in the process. The migrations in this patch are expected to be applied offline. Please be aware that *there are no down migrations*, and if something goes wrong, data loss is possible. Always back up your database before applying migrations. For more information, see [Hydra 2.x Migration Guide](https://www.ory.sh/hydra/docs/guides/migrate-v2).

Rows with NULL login_challenge in `hydra_oauth2_consent_request` and corresponding `hydra_oauth2_consent_request_handled` are deleted as a side effect of the merge migration. This is done with the assumption that only a very small number of sessions, issued by pre-1.0 Hydra, will be affected. Please contact us if this assumption doesn't apply or if the deletion adversely affects your deployment.

## Reviewers Notes

* [ ] Verify persister logic, especially clean up

## TODO

* Hydra 2.x includes a migration that deletes all consent requests (including _handled) that have a NULL login_challenge.
* * [ ] Makre sure that this is communicated in the changelog and in migration warnings.
* * [ ] Include a sanity check query that reports the number of affected rows by month.
* [ ] e2e: Make sure that users remain signed in after consent revocation (edit: Does [this](https://github.com/ory/hydra/blob/master/cypress/integration/oauth2/consent.js#L54) suffice, @aeneasr ?)

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.


flow/flow.go
```diff
@@ -0,0 +1,291 @@
...
+	//if f.ForceSubjectIdentifier != h.ForceSubjectIdentifier {
+	//	return errors.New(fmt.Sprintf("Flow ForceSubjectIdentifier %s does not match the HandledLoginRequest ForceSubjectIdentifier  %s", f.ForceSubjectIdentifier, h.ForceSubjectIdentifier))
+	//}
+	//if h.Error != nil {
+	//	return errors.New("Refusing to process HandledLoginRequest " + h.ID + " with error")
+	//}
+
+	f.State = FlowStateLoginUnused
+	f.ID = h.ID
+	f.Subject = h.Subject
```
grantzvolsky, line 205: Do we need Subject and ForceSubjectIdentifier in the LoginRequest if we overwrite it here?

persistence/sql/migrations/20211019000001000000_merge_authentication_request_tables.sqlite.up.sql
```diff
@@ -0,0 +1,187 @@
...
+     context
+FROM hydra_oauth2_consent_request;
+
+DROP TABLE hydra_oauth2_consent_request;
+ALTER TABLE hydra_oauth2_consent_request_tmp RENAME TO hydra_oauth2_consent_request;
+
+CREATE INDEX hydra_oauth2_consent_request_client_id_idx ON hydra_oauth2_consent_request (client_id);
+CREATE INDEX hydra_oauth2_consent_request_subject_idx ON hydra_oauth2_consent_request (subject);
+CREATE INDEX hydra_oauth2_consent_request_login_session_id_idx ON hydra_oauth2_consent_request (login_session_id);
+CREATE INDEX hydra_oauth2_consent_request_login_challenge_idx ON hydra_oauth2_consent_request (login_challenge);
```
aeneasr, line 187: Are we missing any keys here and if not where can we find the reference to existing keys at the time of this migration?

persistence/sql/persister_consent.go
```diff
@@ -153,18 +154,35 @@ func (p *Persister) GetConsentRequest(ctx context.Context, challenge string) (*c
 }
 
 func (p *Persister) CreateLoginRequest(ctx context.Context, req *consent.LoginRequest) error {
-	return errorsx.WithStack(p.Connection(ctx).Create(req))
+	f := flow.NewFlow(req)
+	return errorsx.WithStack(p.Connection(ctx).Create(f))
```
aeneasr, line 158: We have a helper `github.com/ory/x/sqlcon.HandleError` which will properly wrap the different database errors and return a common error, e.g. for conflicts. Please use that one instead :)

aeneasr (commented):
Looking really good! I found a few things but most of them are related to code style or our tooling :) I think this is the right approach to take here!

persistence/sql/migrations/20211019000001000000_merge_authentication_request_tables.sqlite.up.sql
```diff
@@ -0,0 +1,187 @@
+-- login_used_at
+
+CREATE TABLE hydra_oauth2_flow
```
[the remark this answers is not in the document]
grantzvolsky: Would this be good time to introduce fizz, or should I just split it manually (with awk)?

persistence/sql/migrations/20211019000001000000_merge_authentication_request_tables.sqlite.up.sql
grantzvolsky: The way I make these sqlite migrations is that I dump the current schema, copy the relevant parts, and make the necessary changes. Maybe we could automatically create schema diffs for each migration and version these diffs to make the migrations better reviewable?

grantzvolsky: Here's a bash oneliner to diff the sqlite schema of two commits: `__dump() { P="/tmp/sqlitest/$1.sqlite"; rm $P; yes | go run -tags sqlite . migrate sql "sqlite://$P?_fk=true"; echo '.dump' | sqlite3 $P > /tmp/sqlitest/$1.dump; }; __diff_sqlite() { mkdir -p /tmp/sqlitest; git checkout $1; __dump $1; git checkout $2; __dump $2; git diff --no-index /tmp/sqlitest/$1.dump /tmp/sqlitest/$2.dump; }`. Usage: `__diff_sqlite c0f38431 08a1aa35`

flow/flow.go
```diff
@@ -0,0 +1,291 @@
...
+	f.Error = h.Error
+
+	f.Remember = h.Remember
+	f.RememberFor = h.RememberFor
+	f.ACR = h.ACR
+	f.AMR = h.AMR
+	f.Context = h.Context
+	f.WasHandled = h.WasHandled
+	f.LoginAuthenticatedAt = h.AuthenticatedAt
+	return nil
```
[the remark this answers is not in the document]
grantzvolsky: see the updated TestFlow_HandleLoginRequest

grantzvolsky:
@aeneasr Please review the updated PR. These comments from your previous review remain unresolved:

1. Query optimization using pop foreign keys - Unresolved because replacing `db:"-"` with `belongs_to:"client"` on the Flow.Client field causes the following error; it is probably something trivial, but I couldn't work it out with my limited knowledge of pop: `reflect.Set: value of type uuid.UUID is not assignable to type string`.
2. "A DROP TABLE is probably ok here" - I'll add the down migrations later. Just to be sure, drop table is fine, but we still need to ensure that the down migration restores the original schema, right?
3. "Each instruction must be in an individual file." - See my question on the conversation.
4. "Are we missing any keys here and if not where can we find the reference to existing keys at the time of this migration?" - See my answer in the thread.

persistence/sql/migrations/20211019000001000000_merge_authentication_request_tables.sqlite.down.sql
```diff
@@ -0,0 +1 @@
+-- TODO
```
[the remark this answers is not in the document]
aeneasr: ping

persistence/sql/migrations/20211019000001000000_merge_authentication_request_tables.sqlite.up.sql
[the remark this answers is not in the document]
aeneasr: let's use awk for now, It hink this migration is too complex for fizz anyways :)

persistence/sql/migrations/20211019000001000000_merge_authentication_request_tables.sqlite.up.sql
aeneasr: Ok, I see! That is a good approach :) Maybe add some docs in the PR on the final commit so we can ensure the SQL schemas do indeed match as much as possible?

aeneasr (commented):
Nice! This can definitely continue like that!

persistence/sql/migrations/20211019000001000000_merge_authentication_request_tables.sqlite.down.sql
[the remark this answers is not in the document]
grantzvolsky: I'll add the down migrations later. Just to be sure, even if we can just drop the table, we still want to ensure that the down migration restores the original schema, right?

persistence/sql/migrations/20211019000001000000_merge_authentication_request_tables.sqlite.up.sql
grantzvolsky: Will do :)

persistence/sql/migrations/20211019000001000000_merge_authentication_request_tables.sqlite.up.sql
```diff
@@ -0,0 +1,286 @@
...
+            ch_remember_for IS NOT NULL AND
+            ch_error IS NOT NULL AND
+            ch_session_access_token IS NOT NULL AND
+            ch_session_id_token IS NOT NULL AND
+            ch_was_used IS NOT NULL
+        ))
+    )
+);
+
+INSERT INTO hydra_flow (
```
aeneasr, line 105: I would prefer to name this `hydra_oauth2_flow`, wdyt?

consent/manager_test_helpers.go
```diff
@@ -41,7 +41,7 @@ import (
 
 func MockConsentRequest(key string, remember bool, rememberFor int, hasError bool, skip bool, authAt bool) (c *ConsentRequest, h *HandledConsentRequest) {
```
aeneasr, line 42: Is there a particular reason for changing `challenge` to `consent-challenge`, `fk-login-challenge` to `login-challenge`, and `time.Minute` to `time.Hour`?

persistence/sql/migrations/20211019000001000000_merge_authentication_request_tables.sqlite.down.sql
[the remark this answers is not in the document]
aeneasr: If that's possible, then yes!

driver/registry_base.go
```diff
@@ -431,6 +432,24 @@ func (m *RegistryBase) SubjectIdentifierAlgorithm() map[string]consent.SubjectId
 	return m.sia
 }
 
+func (s *RegistryBase) ObfuscateSubjectIdentifier(cl fosite.Client, subject, forcedIdentifier string) (string, error) {
```
aeneasr, line 435: I would suggest to move this to `ConsentStrategy` because the registry is responsible for instantiation, not for business logic. That way, you can also just pass the consent request which has both `subject` as well as `forcedIdentifier`

oauth2/handler.go
```diff
@@ -682,6 +682,10 @@ func (h *Handler) AuthHandler(w http.ResponseWriter, r *http.Request, _ httprout
 	}
 
 	session, err := h.r.ConsentStrategy().HandleOAuth2AuthorizationRequest(w, r, authorizeRequest)
+	var obfuscatedSubject string
+	if err == nil {
+		obfuscatedSubject, err = h.r.ObfuscateSubjectIdentifier(authorizeRequest.GetClient(), session.ConsentRequest.Subject, session.ConsentRequest.ForceSubjectIdentifier)
+	}
```
aeneasr, line 688: I suggest to move this after the `var accessTokenKeyID string` block just before https://github.com/ory/hydra/pull/2836/files#diff-dc5158991c1ce571f0a59c32b86d6fc64c55fbe11d67cdee44b1b0ae436569dfR727 with explicit error handling.

consent/sdk_test.go
```diff
@@ -61,8 +61,8 @@ func TestSDK(t *testing.T) {
 		Subject: "subject1",
```
aeneasr, line 61: Is there a particular reason for updating these tests? Shouldn't they pass as-is?

consent/strategy_default.go
```diff
@@ -609,11 +591,6 @@ func (s *DefaultStrategy) verifyConsent(w http.ResponseWriter, r *http.Request,
 		return nil, err
 	}
 
-	pw, err := s.obfuscateSubjectIdentifier(req.GetClient(), session.ConsentRequest.Subject, session.ConsentRequest.ForceSubjectIdentifier)
-	if err != nil {
-		return nil, err
-	}
-
```
aeneasr, line 616: Note to self: this is ok because it is handled in the outer function

flow/flow.go
```diff
@@ -0,0 +1,412 @@
...
+		Verifier:               r.Verifier,
+		CSRF:                   r.CSRF,
+		LoginAuthenticatedAt:   r.AuthenticatedAt,
+		RequestedAt:            r.RequestedAt,
+		State:                  FlowStateLoginInitialized,
+	}
+}
+
+func (f *Flow) HandleLoginRequest(h *consent.HandledLoginRequest) error {
+	// TODO Validate HandledLoginRequest. This will require updating tests that expect the flow to fail at a later point.
```
aeneasr, line 204: I think we can remove this TODO item?

flow/flow.go
```diff
@@ -0,0 +1,412 @@
...
+		return errors.Errorf("flow ID %s does not match HandledLoginRequest ID %s", f.ID, h.ID)
+	}
+	if f.Subject != "" && h.Subject != "" && f.Subject != h.Subject {
+		return errors.Errorf("flow Subject %s does not match the HandledLoginRequest Subject %s", f.Subject, h.Subject)
+	}
+	if f.ForceSubjectIdentifier != "" && h.ForceSubjectIdentifier != "" && f.ForceSubjectIdentifier != h.ForceSubjectIdentifier {
+		return errors.Errorf("flow ForceSubjectIdentifier %s does not match the HandledLoginRequest ForceSubjectIdentifier %s", f.ForceSubjectIdentifier, h.ForceSubjectIdentifier)
+	}
+
+	f.State = FlowStateLoginUnused // TODO FlowStateError if h.Error != nil
```
aeneasr, line 218: TODO?

flow/flow.go
```diff
@@ -0,0 +1,412 @@
...
+
+	CHGrantedScope       sqlxx.StringSlicePipeDelimiter `json:"ch_grant_scope" db:"ch_granted_scope"`
+	CHGrantedAudience    sqlxx.StringSlicePipeDelimiter `json:"ch_grant_access_token_audience" db:"ch_granted_at_audience"`
+	CHRemember           bool                           `json:"ch_remember" db:"ch_remember"`
+	CHRememberFor        int                            `json:"ch_remember_for" db:"ch_remember_for"`
+	CHHandledAt          sqlxx.NullTime                 `json:"ch_handled_at" db:"ch_handled_at"`
+	CHWasHandled         bool                           `json:"-" db:"ch_was_used"`
+	CHError              *consent.RequestDeniedError    `json:"-" db:"ch_error"`
+	CHSessionIDToken     sqlxx.MapStringInterface       `json:"-" db:"ch_session_id_token" faker:"-"`
+	CHSessionAccessToken sqlxx.MapStringInterface       `json:"-" db:"ch_session_access_token" faker:"-"`
```
aeneasr, line 178: Can you please add code docs for these fields? Additionally, it's a bit confusing if we have `CHWasHandled` for consent, but `WasHandled` for login. I'd suggest to either prefix both or neither

persistence/sql/migrations/20211019000001000000_merge_authentication_request_tables.sqlite.up.sql
```diff
@@ -0,0 +1,286 @@
...
+
+    ch_granted_scope         TEXT        NOT NULL,
+    ch_granted_at_audience   TEXT        NULL DEFAULT '',
+    ch_remember              INTEGER     NOT NULL,
+    ch_remember_for          INTEGER     NOT NULL,
+    ch_handled_at            TIMESTAMP   NULL,
+    ch_was_used              INTEGER     NOT NULL,
+    ch_error                 TEXT        NOT NULL,
+    ch_session_id_token      TEXT        NOT NULL,
+    ch_session_access_token  TEXT        NOT NULL,
```
aeneasr, line 42: Same as above - we have `consent_`, `` (empty), `ch_` prefixes. I would prefer to use a common language to better understand what field refers to what

persistence/sql/migrations/20211019000001000000_merge_authentication_request_tables.sqlite.up.sql
```diff
@@ -0,0 +1,286 @@
+CREATE TABLE hydra_flow
```
aeneasr, line 1: TODO: Move into separate files as discussed on slack

persistence/sql/persister_consent.go
```diff
@@ -16,30 +16,28 @@ import (
...
+		fs := make([]*flow.Flow, 0)
 		if err := c.
 			Where(whereStmt, whereArgs...).
-			Select("r.challenge").
-			Join(
-				fmt.Sprintf("%s AS r", consent.ConsentRequest{}.TableName()),
-				fmt.Sprintf("r.challenge = %s.challenge", consent.HandledConsentRequest{}.TableName())).
-			All(&hrs); err != nil {
+			Select("challenge").
+			All(&fs); err != nil {
```
aeneasr, line 40: No more joins, yay 🥳

persistence/sql/persister_consent.go
```diff
@@ -48,21 +46,20 @@ func (p *Persister) revokeConsentSession(whereStmt string, whereArgs ...interfac
...
-			if err := p.RevokeRefreshToken(ctx, hr.ID); errors.Is(err, fosite.ErrNotFound) {
+			if err := p.RevokeRefreshToken(ctx, f.ID); errors.Is(err, fosite.ErrNotFound) {
 				// do nothing
 			} else if err != nil {
 				return err
 			}
 
-			// Since we ON DELETE CASCADE, hydra_oauth2_consent_request_handled will be removed automagically.
-			localCount, err := c.RawQuery("DELETE FROM hydra_oauth2_consent_request WHERE challenge = ?", hr.ID).ExecWithCount()
+			localCount, err := c.RawQuery("DELETE FROM hydra_flow WHERE challenge = ?", f.ID).ExecWithCount()
```
aeneasr, line 62: 😎

aeneasr (commented):
I will have to take a deeper dive into the persister logic, high level this looks very good! A major question for me is wether we really need to update the existing SDK tests? IMO everything should pass without modification to any existing tests (except for the persister tests of course)

persistence/sql/migrations/20211019000001000000_merge_authentication_request_tables.sqlite.up.sql
grantzvolsky: > I will have to take a deeper dive into the persister logic, high level this looks very good! A major question for me is wether we really need to update the existing SDK tests? IMO everything should pass without modification to any existing tests (except for the persister tests of course)

There are two reasons why I made these changes:
1. The original tests used the same ID for the consent request and for the login request. I made the IDs different to ensure the new code doesn't mistake one for another, and also to make debugging the tests easier by ensuring it is apparent from the id whether it is for login or for consent. These changes were useful for development and are no longer necessary.
2. Some test changes are necessary because we now do slightly more validation, which means that some shortcuts used in tests are no longer possible. e.g. it is no longer possible to create a consent request without first creating a login request.

Should I go over the test changes and for each one either remove it or clarify why it is necessary?

> I would prefer to name this hydra_oauth2_flow, wdyt?

I don't have a preference, so let's make it hydra_oauth2_flow)

flow/flow.go
grantzvolsky: Correct

flow/flow.go
grantzvolsky: Sure. I'll prefix both

persistence/sql/migrations/20211019000001000000_merge_authentication_request_tables.sqlite.up.sql
grantzvolsky: Agree. How about this? The fields in the Flow struct would be renamed accordingly, including WasHandled -> WasUsed.
```
challenge                ->  login_challenge
requested_scope
verifier                 ->  login_verifier
csrf                     ->  login_csrf
subject
request_url
skip                     ->  login_skip
client_id
requested_at
login_initialized_at
oidc_context
login_session_id
requested_at_audience
state
remember                 ->  login_remember
remember_for             ->  login_remember_for
error                    ->  login_error
acr
login_authenticated_at
was_used                 ->  login_was_used
forced_subject_identifier
context
amr
consent_challenge_id
consent_skip
consent_verifier
consent_csrf
ch_granted_scope         ->  granted_scope
ch_granted_at_audience   ->  granted_at_audience
ch_remember              ->  consent_remember
ch_remember_for          ->  consent_remember_for
ch_handled_at            ->  consent_handled_at
ch_was_used              ->  consent_was_used
ch_error                 ->  consent_error
ch_session_id_token      ->  session_id_token
ch_session_access_token  ->  session_access_token

```

consent/manager_test_helpers.go
grantzvolsky: (1/n) regarding the RequestedAt 'time.Hour' -> 'time.Minute':
I changed this in two places, 1) in the mocked consent request, and 2) in the mocked login request. In the original tests, these values were ignored and the tests passed no matter the time, because the value that was actually used (in tests, not in application code) was the one from HandledConsentRequest, which always was 'time.Minute'. There is only one RequestedAt column now, and it is assigned from the LoginRequest as opposed to HandledConsentRequest (this has always been the case in application code). It follows that changing RequestedAt in the mocked consent request is unnecessary, so I reverted this change in the latest commit. The 'time.Hour' -> 'time.Minute' in MockAuthRequest is necessary for the above reasons.

persistence/sql/migrations/20211019000001000000_merge_authentication_request_tables.sqlite.up.sql
aeneasr: > Should I go over the test changes and for each one either remove it or clarify why it is necessary?

I will take another look and mark the ones where I have questions - your explanation makes sense!

consent/manager_test_helpers.go
aeneasr: Thank you!

consent/sdk_test.go
grantzvolsky: I've updated the PR to make minimal changes to the tests, see explanation in the other thread

driver/registry_base.go
grantzvolsky: Alright, I moved in commit bb34087. I haven't changed the interface to take a consent request because the function is called from a place where we'd then need to create a request stub just to call the function

aeneasr (commented):
Only a very few minor things that once addressed I think we can merge here. I still need to review the persisted logic itself though.

flow/flow.go
```diff
@@ -0,0 +1,459 @@
...
+	FlowStateLoginUsed = int16(3)
+
+	// FlowStateConsentInitialized applies while Hydra waits for a consent request
+	// to be accepted or rejected.
+	FlowStateConsentInitialized = int16(4)
+
+	FlowStateConsentUnused = int16(5)
+	FlowStateConsentUsed   = int16(6)
+	FlowStateLoginError    = int16(128)
+	FlowStateConsentError  = int16(129)
```
aeneasr, line 51: It appears these two are unused in the code. Should we keep them?

consent/types.go
```diff
@@ -166,79 +166,47 @@ type RequestWasHandledResponse struct {
...
-}
+	ConsentRequest  *ConsentRequest     `json:"-"`
+	Error           *RequestDeniedError `json:"-"`
+	RequestedAt     time.Time           `json:"-"`
+	AuthenticatedAt sqlxx.NullTime      `json:"-"`
 
-func (_ HandledConsentRequest) TableName() string {
-	return "hydra_oauth2_consent_request_handled"
+	SessionIDToken     sqlxx.MapStringInterface `json:"-" faker:"-"`
+	SessionAccessToken sqlxx.MapStringInterface `json:"-" faker:"-"`
```
aeneasr, line 203: These two are not properly checked in `flow/flow_test.go` because faker sets them to zero.

flow/flow_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+func TestFlow_GetHandledLoginRequest(t *testing.T) {
+	t.Run("GetHandledLoginRequest should set all fields on its return value", func(t *testing.T) {
+		f := Flow{}
+		expected := consent.HandledLoginRequest{}
+		assert.NoError(t, faker.FakeData(&expected))
+		f.setHandledLoginRequest(&expected)
+		actual := f.GetHandledLoginRequest()
+		expected.LoginRequest = nil
+		actual.LoginRequest = nil
```
aeneasr, line 103: I think it would be better to check for a set value here instead of an unset value?

flow/flow_test.go
```diff
@@ -0,0 +1,198 @@
...
+	})
+}
+
+func TestFlow_HandleLoginRequest(t *testing.T) {
+	t.Run(
+		"HandleLoginRequest should ignore RequestedAt in its argument and copy the other fields",
+		func(t *testing.T) {
+			f := Flow{}
+			assert.NoError(t, faker.FakeData(&f))
+			f.State = FlowStateLoginInitialized
```
aeneasr, line 123: Consider adding & testing

```
f.SessionIDToken = ...
f.SessionAccessToken = ...
```

flow/flow_test.go
```diff
@@ -0,0 +1,198 @@
...
+	})
+}
+
+func TestFlow_GetHandledConsentRequest(t *testing.T) {
+	t.Run("GetHandledConsentRequest should set all fields on its return value", func(t *testing.T) {
+		f := Flow{}
+		expected := consent.HandledConsentRequest{}
+		assert.NoError(t, faker.FakeData(&expected))
+		expected.Session = &consent.ConsentRequestSessionData{}
+		expected.ConsentRequest = nil
```
aeneasr, line 192: Missing

Consider adding & testing

```
expected.SessionIDToken = ...
expected.SessionAccessToken = ...
```

hack/db-diff.sh
```diff
@@ -0,0 +1,136 @@
+#!/bin/bash
```
aeneasr, line 1: Move this to scripts/ maybe and add a short explanation what it does at the top?

persistence/sql/persister_consent.go
```diff
@@ -135,59 +133,107 @@ func (p *Persister) GetForcedObfuscatedLoginSession(ctx context.Context, client,
...
+		req.Verifier,
+		req.CSRF,
+		req.LoginChallenge.String(),
+	).ExecWithCount()
+	if err != nil {
+		return sqlcon.HandleError(err)
+	}
+	if c != 1 {
+		return errorsx.WithStack(x.ErrNotFound)
+	}
```
aeneasr, line 162: Wouldn't this be misleading when c == 2?

persistence/sql/persister_consent.go
```diff
@@ -135,59 +133,107 @@ func (p *Persister) GetForcedObfuscatedLoginSession(ctx context.Context, client,
...
 }
 
 func (p *Persister) GetLoginRequest(ctx context.Context, challenge string) (*consent.LoginRequest, error) {
-	var lr consent.LoginRequest
-	return &lr, p.transaction(ctx, func(ctx context.Context, c *pop.Connection) error {
-		if err := (&lr).FindInDB(c, challenge); err != nil {
+	var lr *consent.LoginRequest
+	return lr, p.transaction(ctx, func(ctx context.Context, c *pop.Connection) error {
+		f := &flow.Flow{}
+		if err := f.FindInDB(c, challenge); err != nil {
```
aeneasr, line 211: To be honest, for readability, I think you can just do `c.Find(f, challenge)` here and delete `func (f *Flow) FindInDB(c *pop.Connection, id string) error`

aeneasr (commented):
All reviewed, just some cosmetics left and it's good to go!

x/clean_sql.go
```diff
@@ -20,6 +20,7 @@ func CleanSQL(t *testing.T, db *sqlx.DB) {
 		"hydra_oauth2_consent_request",
 		"hydra_oauth2_authentication_request_handled",
 		"hydra_oauth2_authentication_request",
```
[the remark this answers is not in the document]
grantzvolsky: Right, I removed their deletion from the cleanup, and I also deleted the unused non-pop version of the function.

persistence/sql/migrations/20211019000001000002_merge_authentication_request_tables.cockroach.up.sql
```diff
@@ -0,0 +1,10 @@
+-- Migration generated by the command below; DO NOT EDIT.
+-- hydra:generate hydra migrations gen
+
+CREATE INDEX hydra_oauth2_flow_client_id_subject_idx ON public.hydra_oauth2_flow USING btree (client_id, subject);
```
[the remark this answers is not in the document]
grantzvolsky: Alright, thanks for the clarification. I wasn't sure where exactly to draw the line. Fortunately adding split marks in the source file is trivial now :)

persistence/sql/persister_consent.go
```diff
@@ -455,25 +505,30 @@ func (p *Persister) FlushInactiveLoginConsentRequests(ctx context.Context, notAf
...
 	queryFormat := `
-	SELECT %[1]s.challenge
-	FROM %[1]s
-	LEFT JOIN %[2]s ON %[1]s.challenge = %[2]s.challenge
+	SELECT login_challenge
+	FROM hydra_oauth2_flow
 	WHERE (
-		(%[2]s.challenge IS NULL)
-		OR (%[2]s.error IS NOT NULL AND %[2]s.error <> '{}' AND %[2]s.error <> '')
+		(state = ` + fmt.Sprint(flow.FlowStateLoginInitialized) + `)
```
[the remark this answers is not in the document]
grantzvolsky: Done. I also simplified the clause by inverting the state condition

persistence/sql/persister_consent.go
grantzvolsky: login_challenge is the primary key and we're talking about a situation where it isn't unique. I was deciding between
1. not handling this case at all
2. handling it with a misleading error
3. handling it properly
I picked 2) because 1) felt wrong and 3) redundant. wdyt?

flow/flow_test.go
grantzvolsky: I'll add the not nil check, but we still have to unset the value because faker fills expected.LoginRequest with a LoginRequest that isn't consistent with the parent HandledLoginRequest

Makefile
```diff
@@ -109,6 +109,19 @@ sdk: .bin/ory
 		swagger generate client -f ./spec/api.json -t internal/httpclient -A Ory_Hydra
 		make format
 
+MIGRATIONS_SRC_DIR = ./persistence/sql/src/
```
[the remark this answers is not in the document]
grantzvolsky: Is this alright?

flow/flow_test.go
grantzvolsky: I added it to `TestFlow_GetHandledConsentRequest`. Not sure about `TestFlow_HandleLoginRequest` - there should be no tokens yet?

flow/flow.go
grantzvolsky: In order to benefit from the error states (rather than adding yet another thing that needs to be checked when working with errors), we'd need to change a lot of the error handling logic. But this is PR is about refactoring the database, so I decided to leave it as a TODO. Should I remove the error states completely and just add a note that they could be introduced in the future?

flow/flow.go
grantzvolsky: Copying my comment from the other thread: "In order to benefit from the error states (rather than adding yet another thing that needs to be checked when working with errors), we'd need to change a lot of the error handling logic. But this is PR is about refactoring the database, so I decided to leave it as a TODO. Should I remove the error states completely and just add a note that they could be introduced in the future?"

consent/types.go
grantzvolsky: I added them manually to TestFlow_GetHandledConsentRequest manually

aeneasr:
Not sure what's going on with golangci-lint? Could you take a look? :)

flow/flow.go
aeneasr: Adding a note could be really helpful for code readers :)

flow/flow.go
aeneasr: Yes, add a short note :)

flow/flow_test.go
aeneasr: Ah, that makes sense!

aeneasr (approved):
Good to go, just the two code comments and figuring out why the CI fails :)

grantzvolsky:
> Good to go, just the two code comments and figuring out why the CI fails :)

All done! I resolved the CI issue by merging in master

aeneasr:
Awesome! Will merge later :)

Issue #1690, opened 2020-01-13
Auth session cannot be prolonged even if the user is active


jimmytheneutrino:
**Describe the bug**

The fix for #1557 changed the logic of when the session is killed. The hard timeout makes it now impossible to prolong the session even if the user is active and keeps refreshing tokens via silent refresh.

Could we reintroduce the possibility for the old behavior while keeping the new behaviour possible, too? In particular, it would be enough if setting rememberFor to a positive value on skipped login would still trigger the recreation of the session cookie. 
See the relevant line: https://github.com/ory/hydra/pull/1564/files#diff-daf069930bf5aee3faa0136e399ebaa2R433

This way, omitting the rememberFor on skipped login would result in the current behaviour (the old cookie is kept, max-age is not changed). But providing positive rememberFor would update the cookie with the new max-age, thus prolonging its life.

aeneasr:
I think it would be better to set `RememberFor` to something large the first time the user authenticates that you can live with, such as a month, 6 months, or a year. It's typically a good idea to purge the cookie at some point. Wouldn't that too resolve your issue?

jimmytheneutrino:
Not exactly. I want the session to die after a period of inactivity of, say, 4-6 hours. But not to die when the user is active. If I could reset rememberFor on skipped request, this could be easily achieved.

aeneasr:
Right, that makes total sense!

The reason we disabled this behavior was because people might accidentally set `RememberFor` to some value always but do actually intend to not refresh the session every time.

Therefore, I think we should add another payload, something along the lines of `RefreshRememberTTL uint64 (or bool)` which would have the effect you want. Still trying to figure out if it should be the number of seconds as well, or if it should be a boolean which basically re-enables the `RememberFor` flag. Not sure at the moment about the naming, appreciate ideas!

jimmytheneutrino:
I think, a more-or-less standard name for such a behavior is session `timeout` (e.g., server.servlet.session.timeout in spring boot). This could, in theory, be different from the total max-age of the session set by rememberFor.

aeneasr:
Ok, tracking this - contributions are welcomed

PR #2848, opened 2021-11-10
feat: option to update session cookie expiry time on session refresh

aarmam:
This pull request introduces feature to update session cookie expiry time on session refresh request.

**Use case**: We want to keep session duration quite short (15 minutes) and force client applications to periodically extend the session by performing authentication requests with `prompt=none`. Each subsequent authentication request produces a new identity token with lifetime of 15 minutes. But as a security measure we want that browser session cookie would not be kept alive any longer than necessary - therefore browser session cookie duration should be periodically extended, each time by 15 minutes (the same lifetime as each new identity token).

**Current situation**: Browser session cookie (`oauth2_authentication_session`) expiration is set from first [acceptLoginRequest](https://www.ory.sh/hydra/docs/reference/api/#operation/acceptLoginRequest)'s `remember_for` value.
When performing subsequent session update requests (authentication requests with `prompt=none`), then browser session cookie expiration cannot be changed.

**Proposed solution**: Add `refresh_remember_for` parameter for [PUT /oauth2/auth/requests/login/accept](https://www.ory.sh/hydra/docs/reference/api/#operation/acceptLoginRequest) request body. When `refresh_remember_for=true`, session cookie expiry will be reset.

## Related issue(s)
https://github.com/ory/hydra/issues/1690
https://github.com/ory/hydra/issues/1557
https://github.com/ory/hydra/discussions/2246

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments
**Tests and documentation will be commited after inital acceptance of the proposed feature.**

### DIFF
--- a/consent/types.go
+++ b/consent/types.go
@@ -265,6 +265,10 @@ type HandledLoginRequest struct {
 	// authorization will be remembered for the duration of the browser session (using a session cookie).
 	RememberFor int `json:"remember_for"`
 
+	// RefreshRememberFor, if set to true, session cookie expiry time will be updated when session is
+	// refreshed (login skip=true).
+	RefreshRememberFor bool `json:"refresh_remember_for"`
+
 	// ACR sets the Authentication AuthorizationContext Class Reference value for this authentication session. You can use it
 	// to express that, for example, a user authenticated using two factor authentication.
 	ACR string `json:"acr"`
--- a/consent/strategy_default.go
+++ b/consent/strategy_default.go
@@ -425,7 +425,7 @@ func (s *DefaultStrategy) verifyAuthentication(w http.ResponseWriter, r *http.Re
 		}
 	}
 
-	if !session.Remember || session.LoginRequest.Skip {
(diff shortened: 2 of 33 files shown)


consent/types.go
```diff
@@ -265,6 +265,10 @@ type HandledLoginRequest struct {
 	// authorization will be remembered for the duration of the browser session (using a session cookie).
 	RememberFor int `json:"remember_for"`
 
+	// RefreshRememberFor, if set to true, session cookie expiry time will be updated when session is
+	// refreshed (login skip=true).
+	RefreshRememberFor bool `json:"refresh_remember_for"`
```
sgal, line 270: To be consistent with the existing values, it is better to name this property `refresh_remember` since it is a boolean value.

Existing props are following this format:
- `remember` - boolean
- `remember_for` - number

aeneasr: Agree, maybe even clearer `extend_session_lifespan`. Could also be a time value alternatively (although it would be conflicting with `remember_for`)

aarmam: I'm on vacation and will come back to this in few weeks! :)

PR #2844, opened 2021-11-08
feat: revoke consent by session id. trigger back channel logout.

aarmam:
This pull request introduces feature to revoke consent by session id and option to trigger back channel logout.

Use case description:

1. User logs in from device/browser 1 to client application A. Hydra has created login session 1 and authenticated user. 
2. User logs in from device/browser 1 to client application B by accepting consent.
3. User performs steps 1, 2 in device/browser 2. Hydra has created login session 2 for the same user.
4. User initiates log out from client application A using device/browser 1.
5. Logout UI with message "You have been logged out from Application A. You have active sessions in following applications: Application B" options 5.1) Logout 5.2) Resume sessions
6. User selects 5.2
6.1 Logout provider performs PUT "/oauth2/auth/requests/logout/reject"
6.2 Logout provider performs DELETE "/oauth2/auth/sessions/consent"?subject=user1&client=clientA&login_session_id=session1&trigger_backchannel_logout=true"

In result **only application A consent from session 1** is revoked and backchannel logout is triggered as requested. 

Triggering backchannel logout is a separate feature and can be used without _login_session_id_ or with _all=true_ parameter.

## Related issue(s)

https://github.com/ory/hydra/issues/2666

## Checklist

- [x] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [x] I have referenced an issue containing the design document if my change
      introduces a new feature.
- [x] I am following the
      [contributing code guidelines](../blob/master/CONTRIBUTING.md#contributing-code).
- [x] I have read the [security policy](../security/policy).
- [x] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security. vulnerability, I
      confirm that I got green light (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [ ] I have added tests that prove my fix is effective or that my feature
      works.
- [ ] I have added or changed [the documentation](docs/docs).

## Further Comments

Tests and documentation will be commited after inital acceptance of the proposed feature.

PR #2849, opened 2021-11-10
feat: async backchannel logout

aarmam:
This pull request introduces feature to execute backchannel logout asynchronously.

**Use case**: User should not have to wait for backchannel requests to finish (4x30sec, with potential connection timeout 1min on each request), when performing logout from all applications using GET /oauth2/sessions/logout or from specific application by using DELETE /oauth2/auth/sessions/consent?trigger_backchannel_logout=true (https://github.com/ory/hydra/pull/2844)

Will add configuration option if needed.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

**Tests and documentation will be commited after inital acceptance of the proposed feature.**

### DIFF
--- a/consent/strategy_default.go
+++ b/consent/strategy_default.go
@@ -26,7 +26,6 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/ory/hydra/driver/config"
@@ -696,13 +695,9 @@ func (s *DefaultStrategy) executeBackChannelLogout(ctx context.Context, r *http.
 		tasks = append(tasks, task{url: c.BackChannelLogoutURI, clientID: c.OutfacingID, token: t})
 	}
 
-	var wg sync.WaitGroup
 	hc := httpx.NewResilientClient()
-	wg.Add(len(tasks))
 
 	var execute = func(t task) {
-		defer wg.Done()
-
 		res, err := hc.PostForm(t.url, url.Values{"logout_token": {t.token}})
 		if err != nil {
 			s.r.Logger().WithRequest(r).WithError(err).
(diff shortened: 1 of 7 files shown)


consent/strategy_default.go
```diff
@@ -696,13 +695,9 @@ func (s *DefaultStrategy) executeBackChannelLogout(ctx context.Context, r *http.
...
 	}
 
-	var wg sync.WaitGroup
 	hc := httpx.NewResilientClient()
-	wg.Add(len(tasks))
 
 	var execute = func(t task) {
-		defer wg.Done()
-
 		res, err := hc.PostForm(t.url, url.Values{"logout_token": {t.token}})
```
aarmam, line 701: Retry could take 5 x 60sec request timeout + backoff time in between. We could generate updated logout token for each request so `iat` does not expire.
```golang
var body = retryablehttp.ReaderFunc(func() (io.Reader, error) {			
	return strings.NewReader(url.Values{"logout_token": {generateToken}}.Encode()), nil
})
req, err := retryablehttp.NewRequest("POST", c.BackChannelLogoutURI, body)
if err != nil {
	s.r.Logger().WithRequest(r).WithError(err).
		WithField("client_id", c.OutfacingID).
		WithField("backchannel_logout_url", c.BackChannelLogoutURI).
		Error("Unable to create OpenID Connect Back-Channel Logout Request")
	return
}
req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
res, err := hc.Do(req)
if err != nil {
	s.r.Logger().WithRequest(r).WithError(err).
		WithField("client_id", c.OutfacingID).
		WithField("backchannel_logout_url", c.BackChannelLogoutURI).
		Error("Unable to execute OpenID Connect Back-Channel Logout Request")
	return
}
```

aeneasr (changes requested):
It looks like this PR is somewhat broken? There is only one file that has changes unrelated to the PR's title

aeneasr (changes requested):
I think this makes sense! Could you please add some documentation that explains that this runs async?

aeneasr (approved):

aeneasr (approved):

PR #2875, opened 2021-11-29
feat: backchannel logout request client tls configuration

aarmam:
This pull request introduces feature to configure backchannel logout request client TLS min/max versions and supported cipher suites.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.


spec/config.json
```diff
@@ -421,6 +421,61 @@
         }
       }
     },
+    "client.tls": {
```
aeneasr, line 424: Can you please move `tls` to a separate sub key? :) Dots in key names are not handled very well by the tooling.

consent/strategy_default.go
```diff
@@ -697,7 +701,13 @@ func (s *DefaultStrategy) executeBackChannelLogout(ctx context.Context, r *http.
 	}
 
 	var wg sync.WaitGroup
-	hc := httpx.NewResilientClient()
+	hc := httpx.NewResilientClient(
+		httpx.ResilientClientWithClient(&http.Client{
+			Timeout: time.Minute,
+			Transport: &http.Transport{
+				TLSClientConfig: s.tlsClientConfig,
```
aeneasr, line 708: Is it problematic if tlsClientConfig is nil?

aeneasr (changes requested):
Looking great! :) Could you please add a guide that explains how to use this feature in this doc: https://github.com/ory/hydra/blob/master/docs/docs/guides/logout.mdx ?

Thank you!

aeneasr:
Thank you, this looks great! The CI is failing because some files are formatted incorrectly. To format them, run:

```
$ make format
$ git commit -a -m "styles: format code"
$ git push
```

Thank you!

spec/config.json
aeneasr: Thinking about this a bit more. I think the configuration key should be more explicit. There are two options:

- Support mTLS globally, so use it in all the clients (e.g. webhooks, backchannel, ...)
- Have mTLS configured per feature

In the first case, the current structure is probably ok. Maybe it makes sense to have a structure like:

```
http:
  client:
    default:
      tls:
      # in the future:
      # timeout ....
   backchannel_logout:
     tls:
       # ...
   refresh_token_webhook:
     # ...
```

WDYT?

aarmam: @aeneasr implemented per feature configuration with default fallback

consent/strategy_default.go
aarmam: `tlsClientConfig` cannot be nil

https://github.com/ory/hydra/pull/2875/files#diff-265e72885dd37a9d8c6f81e2c928b2bfad035cecd98051ef2f04ea88a56ed540R77

spec/config.json
```diff
@@ -452,6 +513,22 @@
...
     },
+    "client": {
+      "default": {
+        "tls": {
+          "$ref": "#/definitions/client_tls_config"
+        }
+      },
+      "back_channel_logout": {
+        "tls": {
+          "allOf": [
```
aeneasr, line 524: I don't think you need the `allOf` here, just using `ref` like above should be enough.

aeneasr (changes requested):
Thank you! I've got one question. This PR is currently targeting exactly one backchannel logout URL. However, most environments will have 1..n backchannel logout URLs, depending on the clients that define those URLs. To me it appears that setting this TLS configuration at a global level will be problematic if you have two or more OAuth2 clients with separate TLS configs (e.g. one special TLS config and one regular HTTPs with a public certificate).

What's your take on this?

aarmam:
> This PR is currently targeting exactly one backchannel logout URL. However, most environments will have 1..n backchannel logout URLs, depending on the clients that define those URLs. 

This is how its implemented before this PR also - global golang default for all backchannel requests? 

> To me it appears that setting this TLS configuration at a global level will be problematic if you have two or more OAuth2 clients with separate TLS configs (e.g. one special TLS config and one regular HTTPs with a public certificate).

And I did not add this problem with current implementation?

Just reiterating my thoughts so I'm not missing anything: :)

- If `client.default.tls`/`client.back_channel_logout.tls` are not set, then [golang defaults are used](https://github.com/ory/hydra/pull/2875/files#diff-cebba46feafaad907019baaa1089ca35ad328ac5bd878d9361886922757a1c73R88) like previously (?)
- If `client.default.tls` is set, then the settings are [merged with golang default settings](https://github.com/ory/hydra/pull/2875/files#diff-cebba46feafaad907019baaa1089ca35ad328ac5bd878d9361886922757a1c73R90-R115) and would be used for [all Hydra clients, that decide to use this functionality.](https://github.com/ory/hydra/pull/2875/files#diff-cebba46feafaad907019baaa1089ca35ad328ac5bd878d9361886922757a1c73R83-R85)
- If **only** `client.back_channel_logout.tls` is specified, then only [backchannel client will have custom tls configuration](https://github.com/ory/hydra/pull/2875/files#diff-265e72885dd37a9d8c6f81e2c928b2bfad035cecd98051ef2f04ea88a56ed540R77) merged with golang default settings.
- If both `client.default.tls`/`client.back_channel_logout.tls` are set, then [backchannel client will have default and fallback settings merged with golang default settings.](https://github.com/ory/hydra/pull/2875/files#diff-265e72885dd37a9d8c6f81e2c928b2bfad035cecd98051ef2f04ea88a56ed540R77)

This is functionality is MVP for us, **but it could be further extended** to check client specific settings when iterating over clients in executeBackChannelLogout() and it would not go in conflict with current implementation - in contrary would give possibility to further override these global settings when needed?

aeneasr (changes requested):
Thank you @aarmam as always for your great contributions! After thinking about this for quite some time, I don't think that we should merge this feature upstream. The reason is the following:

The backchannel logout URL is something that is configured on a per-client basis. Adding a config option which allows us only to configure a global TLS certificate does not really match the underlying principle that an authorization server can have almost unlimited backchannel logout targets.

If we were to introduce this, we would need to keep for backwards compatibility and also add to Ory Network. Therefore, we can't merge the feature as is.

One option is to add this feature to the client configuration. However, I think that the backchannel logout URL should use TLS with a trusted certificate, and not trust untrusted certificates. To work around this on local environments, one could use Caddy or something similar to add a trusted CA to the OS certificate store, or set up LE in staging environments.

So generally speaking, I don't think we should add TLS exceptions here. It will only open questions like: Why does JWKs fetching not have a custom TLS certificate? Or webhooks, or feature X, Y, Z

aarmam:
> Adding a config option which allows us only to configure a global TLS certificate does not really match the underlying principle that an authorization server can have almost unlimited backchannel logout targets.

This is actually not what this pr introduces. I'll give you an example:
```yaml
client.default.tls:
  min_version: tls12
  max_version: tls13
  cipher_suites:
    - TLS_AES_128_GCM_SHA256
client.back_channel_logout.tls:
  cipher_suites:
    - TLS_AES_128_GCM_SHA256
    - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
    - TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
```
This configuration limits all outgoing TLS connections to use min/max tls versions and specified cipher suites - thats all this pr introduces. It does not configure global TLS certificate nor trusted certificates.
In this example client.back_channel_logout.tls.cipher_suites overrides client.default.tls.cipher_suites.

Issue #2824, opened 2021-10-28
Login request not fully marked as used until browser navigates to /oauth2/auth


Conduitry:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/hydra/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/hydra/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

After accepting a login request with `PUT /oauth2/auth/requests/login/accept` but before the user's browser has navigated to the public `/oauth2/auth` URL, the login request has not yet been properly marked as "used". Calls to `GET /oauth2/auth/requests/login` succeed (seemingly indicating that this is still a valid login request), but further calls to `PUT /oauth2/auth/requests/login/accept` return a 409 error coming from the underlying database.

### Reproducing the bug

1. Follow the steps in [the quickstart tutorial](https://www.ory.sh/hydra/docs/5min-tutorial/)
2. When logging in, using some mechanism to make sure the browser doesn't then navigate to the `redirect_to` target (For example: Open browser dev tools, submit the form with incorrect credentials, grab the cURL command, edit it to fix the incorrect credentials, and then manually run it without automatically following redirects.)
3. Navigate to or refresh the login form. Observe that the `login_challenge` is accepted by `GET /oauth2/auth/requests/login`.
4. Fill out the form again and submit it. Observe the error response from `PUT /oauth2/auth/requests/login/accept`

### Relevant log output

```shell
[...]

PR #2914, opened 2022-01-03
fix: allow retries of accept login request

jonkjenn:
## Related issue(s)

#2824

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/flow/flow.go
+++ b/flow/flow.go
@@ -1,7 +1,6 @@
 package flow
 
 import (
-	"github.com/ory/x/errorsx"
 	"time"
 
 	"github.com/gofrs/uuid"
@@ -256,7 +255,7 @@ func NewFlow(r *consent.LoginRequest) *Flow {
 func (f *Flow) HandleLoginRequest(h *consent.HandledLoginRequest) error {
 
 	if f.State == FlowStateLoginUnused {
-		return errorsx.WithStack(x.ErrConflict)
+		return x.ErrConflict
 	} else if f.State != FlowStateLoginInitialized {
 		return errors.Errorf("invalid flow state: expected %d, got %d", FlowStateLoginInitialized, f.State)
 	}
--- a/flow/flow.go
+++ b/flow/flow.go
@@ -255,7 +255,7 @@ func NewFlow(r *consent.LoginRequest) *Flow {
 func (f *Flow) HandleLoginRequest(h *consent.HandledLoginRequest) error {
 
 	if f.State == FlowStateLoginUnused {
-		return x.ErrConflict
+		return nil
 	} else if f.State != FlowStateLoginInitialized {
 		return errors.Errorf("invalid flow state: expected %d, got %d", FlowStateLoginInitialized, f.State)
 	}
--- a/flow/flow.go
+++ b/flow/flow.go
@@ -1,6 +1,7 @@
 package flow
 
 import (
+	"github.com/ory/x/errorsx"
 	"time"
 
 	"github.com/gofrs/uuid"
@@ -253,7 +254,10 @@ func NewFlow(r *consent.LoginRequest) *Flow {
 }
 
 func (f *Flow) HandleLoginRequest(h *consent.HandledLoginRequest) error {
-	if f.State != FlowStateLoginInitialized {
+
+	if f.State == FlowStateLoginUnused {
+		return errorsx.WithStack(x.ErrConflict)
+	} else if f.State != FlowStateLoginInitialized {
 		return errors.Errorf("invalid flow state: expected %d, got %d", FlowStateLoginInitialized, f.State)
 	}
 	if f.ID != h.ID {
--- a/persistence/sql/persister_consent.go
+++ b/persistence/sql/persister_consent.go
@@ -266,6 +266,14 @@ func (p *Persister) HandleLoginRequest(ctx context.Context, challenge string, r 
 		if err != nil {
 			return sqlcon.HandleError(err)
 		}
+
+		if f.State == flow.FlowStateLoginUnused {
+			lr, _ = p.GetLoginRequest(ctx, challenge)
+			return errorsx.WithStack(x.ErrConflict)
+		} else if err != nil {
+			return err
+		}
+
 		err = f.HandleLoginRequest(r)
 		if err != nil {
 			return err
--- a/persistence/sql/persister_consent.go
+++ b/persistence/sql/persister_consent.go
@@ -267,12 +267,6 @@ func (p *Persister) HandleLoginRequest(ctx context.Context, challenge string, r 
 			return sqlcon.HandleError(err)
 		}
 
-		if f.State == flow.FlowStateLoginUnused {
-			return errorsx.WithStack(x.ErrConflict)
-		} else if err != nil {
-			return err
-		}
-
 		err = f.HandleLoginRequest(r)
 		if err != nil {
 			return err
--- a/persistence/sql/persister_consent.go
+++ b/persistence/sql/persister_consent.go
@@ -266,7 +266,6 @@ func (p *Persister) HandleLoginRequest(ctx context.Context, challenge string, r 
 		if err != nil {
 			return sqlcon.HandleError(err)
 		}
-
 		err = f.HandleLoginRequest(r)
 		if err != nil {
 			return err
--- a/persistence/sql/persister_consent.go
+++ b/persistence/sql/persister_consent.go
@@ -268,7 +268,6 @@ func (p *Persister) HandleLoginRequest(ctx context.Context, challenge string, r 
 		}
 
 		if f.State == flow.FlowStateLoginUnused {
-			lr, _ = p.GetLoginRequest(ctx, challenge)
 			return errorsx.WithStack(x.ErrConflict)
 		} else if err != nil {
 			return err
--- a/consent/handler.go
+++ b/consent/handler.go
@@ -357,16 +357,26 @@ func (h *Handler) AcceptLoginRequest(w http.ResponseWriter, r *http.Request, ps 
 	p.RequestedAt = ar.RequestedAt
 
 	request, err := h.r.ConsentManager().HandleLoginRequest(r.Context(), challenge, &p)
+<<<<<<< HEAD
 	if errors.Is(err, x.ErrConflict) {
 		ru, err := url.Parse(ar.RequestURL)
+=======
+	if request != nil && errors.Is(err, x.ErrConflict) {
+		ru, err := url.Parse(request.RequestURL)
+>>>>>>> Refactor
 		if err != nil {
 			h.r.Writer().WriteError(w, r, err)
 			return
 		}
 		h.r.Writer().WriteCode(w, r, http.StatusConflict, &RequestWasHandledResponse{
+<<<<<<< HEAD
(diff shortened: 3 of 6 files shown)


persistence/sql/persister_consent.go
```diff
@@ -216,11 +216,18 @@ func (p *Persister) VerifyAndInvalidateConsentRequest(ctx context.Context, verif
 
 func (p *Persister) HandleLoginRequest(ctx context.Context, challenge string, r *consent.HandledLoginRequest) (lr *consent.LoginRequest, err error) {
 	return lr, p.transaction(ctx, func(ctx context.Context, c *pop.Connection) error {
-		err := c.Create(r)
+		found, err := c.Where("challenge = ?", challenge).Exists(&consent.HandledLoginRequest{})
 		if err != nil {
 			return sqlcon.HandleError(err)
 		}
 
+		if !found {
```
aeneasr, line 224: This would only work in the case of an exact retry. It will be confusing for developers who expect the second call to override the first one, which is why we wanted to return 429 with an URL the developers can use for next action.

aeneasr (changes requested):
Happy new year @jonkjenn :)

Thank you for working on this issue and your PR! I have two asks of you:

1. Would it be possible to make your changes against the #2796 branch (in particular PR #2836) as we are changing the database layout.
2. As described in issue #2824 we originally wanted to return 429 with an URL the developer can use for further propagation. The reason for not allowing double submit here is that we want to prevent conflicting submissions and potential race conditions. It's been quite some time that we introduced this and I don't quite remember the reasoning. But I think we need a sound overview of the potential attack vectors of allowing double submission here before we can change the behavior to just override the previous request or ignore it.
3. Can you introduce this feature to both handling consent and login?

jonkjenn:
> 3. Can you introduce this feature to both handling consent and login?

It already seems to work to retry the accept consent requests. I tested v1.10.6 towards our project and this branch by modifying the Cypress e2e tests with duplicate requests and also some unit test but those gets quite convoluted quickly when you get to the consent step since there's so much setup and options.

persistence/sql/persister_consent.go
jonkjenn: Looks like consent is supported already https://github.com/ory/hydra/pull/2914#issuecomment-1008902061. Do you see the need for further work to ensure that enabling this at all is a secure solution @aeneasr? I.e. is it not realistic to get a change like this merged as more of a quick fix?

mig5:
This looks great and as far as I can tell, would make the behavior consistent with Login Requests when the token is attempted to be reused?

I'm not sure there's a security risk associated if it means that via the redirect_uri, the browser can reattempt the auth flow. Presumably the token would be different once it arrives at the consent screen again as part of that auth flow.

Would love to see this in the next release!

aeneasr:
@jonkjenn is this ready for review? :)

jonkjenn:
I changed to returning 200 on retries to stay consistent with how accept consent works also it feels more correct not to use an error for this case. The new tests shows more clearly how both retries of accept login and accept consent works.

aeneasr:
Thank you! I think the tests are awesome! I would like to follow up with a comment made in this issue:

https://github.com/ory/hydra/issues/2824#issuecomment-955816154
https://github.com/ory/hydra/issues/2824#issuecomment-956168851
https://github.com/ory/hydra/issues/2824#issuecomment-1085240321

which would basically mean that we return an error like 409/410 but include a `redirect_url` in the response:

```
{"error":"Unable to insert or update resource because a resource with that value exists already","error_description":"","redirect_to":"https://.../oauth2/..."}.
```

which would help the developer implementing this flow to identify the URL they need to return the user to. What do you think about that solution? In the issue thread, it appears that most agree :) I think it would be relatively straight forward to modify this PR to follow that behavior as all the tests and everything is already in place :)

jonkjenn:
@aeneasr So 409 for /accept/login and leave /accept/consent as 200?

aeneasr:
409 for both - or is consent currently returning 200 on conflict? Sorry I didn’t have time yet to refresh my knowledge on the issue…

mig5:
@jonkjenn @aeneasr 

> So 409 for /accept/login and leave /accept/consent as 200?
> 409 for both - or is consent currently returning 200 on conflict? Sorry I didn’t have time yet to refresh my knowledge on the issue…

Let me try and clarify:

Both Login and Consent flows can experience a '410 Gone' on `GET /oauth2/auth/requests/login?login_challenge=xxxxxxxxxxxxx` or `GET /oauth2/auth/requests/consent?consent_challenge=xxxxxxxxxxxxxxx` 

In both cases, the 410 Gone error has a `redirect_uri` parameter in the error response, which we can redirect to in order to 'restart' the flow transparently without the user getting blocked on an error page.

But a 409 Conflict on the Consent flow (`PUT /oauth2/auth/requests/consent/accept?consent_challenge=xxxxxxxxxxxxxx`) does *not* have a `redirect_uri` in the response. According to https://github.com/ory/hydra/issues/2824#issuecomment-955811384 , this is also true for the 409 on the Login flow (but I personally haven't been able to trigger that scenario, only the 410)

In other words, in both Login and Consent flows, a PUT with the challenge token throws a 409 but does not have a redirect_uri in the error response. But a 410 Gone (on the GET) does.

We are hoping for the same redirect_uri parameter to be added to the error response for a 409, in either Login or Consent case. That's all.

The 409 Conflict on the Consent form is easy to trigger, assuming you are not 'skipping' the consent:

1) get to the 'Consent' form in a tab (after completing the Login flow), e.g `https://yourloginandconsentapp.com/consent?consent_challenge=xxxxxxxxxxxxxxxxxxx`
2) load that same page in a second tab (it's got the same consent_challenge token)
3) submit the form on the first tab (flow completes normally)
4) submit the form on the second tab (gets the following 409 Conflict in Hydra):

```
time=2022-04-20T03:35:00Z level=error msg=An error occurred while handling a request audience=application error=map[debug: message:Conflict reason:The consent request was already used and can no longer be changed. status:Conflict status_code:409] http_request=map[headers:map[accept:application/json content-length:130 content-type:application/json [.......] method:PUT path:/oauth2/auth/requests/consent/accept [.....] http_response=map[status_code:409] service_name=Ory Hydra service_version=v1.11.7
```

... and there is no redirect_uri in the error response unlike a 410 .

(Obviously if you optionally 'save' the consent the first time, you tend not to hit this problem again (at least until any such consent is revoked by the user), because the redirects happen automatically and the Consent form thus never has to be submitted again.)

Ultimately the most important thing from an implementor's perspective is to be able to gracefully recover by consuming the `redirect_uri` in the error response for any error. Right now that only works for 410s but not the 409 on PUT.

I hope that clarifies things!

mig5:
Edited my comment above to note that it's named `redirect_to` in the 410 Gone error responses, not `redirect_uri`. We obviously would want this to be consistently named in the 409 too once it exists for that response :)

jonkjenn:
@mig5 You're talking about the case where the user has visited the the `redirect_to` returned from the first /consent/accept as I understand it? Just to be clear this PR does not attempt to deal with that scenario only the scenario where you "need to" perform multiple /(login|consent)/accept BEFORE the user visits the `redirect_to`.

All my tests indicate that before the user visits the `redirect_to` in the consent case the response to multiple /consent/accept requests is always 200 with the same response payload.  This test hopefully shows it directly  https://github.com/ory/hydra/pull/2914/files#diff-852c558f6e189a7d5ce22301e754abf3ef10d2785b7a708922cbd026413009c4R101 and it's what I see when I manually verify.

@aeneasr 
> 409 for both - or is consent currently returning 200 on conflict? Sorry I didn’t have time yet to refresh my knowledge on the issue…

Yes it is currently returning 200 with the same `redirect_to` payload on multiple PUT consent/accept requests BEFORE the user has visited the `redirect_to`. Which is why it feels strange to have different behavior for PUT login/accept unless it's really required. Changing the consent response of course would be a breaking change.

mig5:
@jonkjenn 

> only the scenario where you "need to" perform multiple /(login|consent)/accept BEFORE the user visits the redirect_to.

Right - I am saying, when the 'multiple' request fails on 409 (PUT), it would be nice to have the same `redirect_to` be present in the error message, for both Login and Consent. Right now in the latest release of Hydra, there is only a `redirect_to` on 410 (a failure on GET), but not on the 409.

flow/flow.go
```diff
@@ -251,7 +251,10 @@ func NewFlow(r *consent.LoginRequest) *Flow {
 }
 
 func (f *Flow) HandleLoginRequest(h *consent.HandledLoginRequest) error {
-	if f.State != FlowStateLoginInitialized {
+
+	if f.State == FlowStateLoginUnused {
+		return nil
+	} else if f.State != FlowStateLoginInitialized {
```
aeneasr, line 257: Would it make sense to use the same logic as in the consent flow?

```

	if f.State != FlowStateConsentInitialized && f.State != FlowStateConsentUnused && f.State != FlowStateConsentError {
		return errors.Errorf("invalid flow state: expected %d/%d/%d, got %d", FlowStateConsentInitialized, FlowStateConsentUnused, FlowStateConsentError, f.State)
	}
```

aeneasr (commented):
@grantzvolsky can you maybe take a look, since you wrote the state machine? Does the code change make sense in your view?

flow/flow.go
aeneasr: On a second look I am still confused though. The 409 error does not happen in this logic here, it comes from the database or explicitly returning `x.ErrConflict`.

I then executed the tests included in this PR without the changes above, and the tests pass, indicating that the change has not the desired effect and the code on this branch already behaves the way we want it to?

aeneasr (changes requested):

aeneasr:
Thank you all for the explanations! As per my review, it appears that the code is already behaving as expected in the test cases as they seem to pass without the changes to `flow.go`.

flow/flow.go
jonkjenn:
```suggestion
        if f.State != FlowStateLoginInitialized && f.State != FlowStateLoginUnused {
		return errors.Errorf("invalid flow state: expected %d/%d, got %d", FlowStateLoginInitialized, FlowStateLoginUnused, f.State)
	}
```

I don't understand exactly what you're saying is working or not but the suggestion of changing it to be more similar to the `HandleConsentRequest` function makes sense to me and if I'm not wrong it's logically equivalent.

aeneasr: Thank you for getting back :) I ran the tests you provided but changed the code in flow.go back to what it was before your changes, and the tests you wrote still pass, making me wonder whether we need this change in flow.go?

jonkjenn: Hmm they fail locally for me but I could be doing something wrong. I created https://github.com/ory/hydra/pull/3085 to attempt to run tests in the pipeline but if there is a PR pipeline I didn't manage to trigger it.

aeneasr: My bad @jonkjenn ! You're right :)

aeneasr (approved):

jonkjenn:
Oops didn't notice your PR

flow/flow.go
grantzvolsky: @jonkjenn cc @aeneasr you are right that there are some opportunities for more consistency checks. When I introduced Flow, I initially tried to validate everything. But it turned out that some tests depend on imperfections in the validation, and that updating them would be a lot of work, so I changed my strategy and just made the minimum necessary changes.

Issue #2930, opened 2022-01-10
feat: RFC 7523 - allow multiple subjects in trust relationships


jagobagascon:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/hydra/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/hydra/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

The current implementation of the RFC requires a trust relationship to exist between the JWT issuer and the authorised subject. This means that a JWT issuer that needs to authorise many users requires a trust relationship for each of them to exist in Hydra.

I understand that in some environments the way this is implemented adds an additional layer of security: you may not have control over the JWT issuer, and this lets you limit who can be authorised by it. But in others (when the JWT issuer is part of your own infrastructure) it just adds a dependency between the issuer and Hydra and doesn't add much value in terms of security. 

Our problem is that **our JWT issuer does not have access to Hydra's admin backend**.

### Describe your ideal solution

**We want a way to define trust relationships that allow the authorisation of many subjects**, either by allowing any subject or only subjects under a certain domain (`*@mydomain.com`) to be authorised.

This trust relationships are out of the scope of the RFC, so adding this would have no impact on RFC compliance.

### Workarounds or alternatives

I guess that exposing the Hydra backend to the internet (with some kind of security of course) to create a trust relationship before issuing a new JWT token would "fix" our problem.

But there's no way we are doing this. It would create a bunch of new problems, and a new attack vector to our infrastructure.

### Version

master branch (The RFC 7523 feature is yet to be released)

### Additional Context

There was some discussion about this topic in https://github.com/ory/hydra/issues/2229#issuecomment-741682310. 

There it was proposed to use the `anonymous` keyword as a subject and make it a required parameter. And I kinda like that solution, but I'm afraid it could cause problems on systems where `anonymous` is a valid user/subject name. So I think it should just be a flag on grant creation and just keep it as a null value in the database.

[...]

PR #2957, opened 2022-01-27
feat: add support for domain wide trust relationships

jagobagascon:
This PR adds support for domain wide trust relationships. This will allow setting issuers that can sign tokens for every user under the configured domain, instead registering all subjects individually.

The following changes have been made:

#### REST service
Added a new `domain` field in the REST service used to setup the trust grants. Also the `subject` field is no longer required, either one of `domain` or `subject` has to be present, never both. Validation has been updated to this.
```
POST https://<admin.ory-hydra>/trust/grants/jwt-bearer/issuers
Content-Type: application/json

{
  "issuer": "https://my-issuer.com",
  "domain": "example.com",
  "scope": ["read"],
  "jwk": { ... },
  "expires_at": "2021-04-23T18:25:43.511Z",
}
```

Single subject trust relationships can still be setup by defining setting the `subject` field in the request:
```
POST https://<admin.ory-hydra>/trust/grants/jwt-bearer/issuers
Content-Type: application/json

{
  "issuer": "https://my-issuer.com",
  "subject": "alice@example.com",
  "scope": ["read"],
  "jwk": { ... },
  "expires_at": "2021-04-23T18:25:43.511Z",
}
```

#### Database
Added a new `domain` column to the `hydra_oauth2_trusted_jwt_bearer_issuer` table, and updated the unique index to contain this new column.

#### JWT Token validation
When a token is received, if the subject looks like an email, hydra will look for public keys that exactly match the subject or the domain (`subject = ? or domain = ?`). If the subject is not an email then only the subject is checked (keeps previous behaviour).

## Related issue(s)
#2930

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/cypress/integration/oauth2/grant_jwtbearer.js
+++ b/cypress/integration/oauth2/grant_jwtbearer.js
@@ -79,7 +79,7 @@ describe('The OAuth 2.0 JWT Bearer (RFC 7523) Grant', function () {
   const gr = (subject, domain) => ({
     issuer: prng(),
     subject: subject,
-    domain: domain,
+    allowed_domain: domain,
     scope: ['foo', 'openid', 'offline_access'],
     jwk: testPublicJwk,
     expires_at: dayjs().utc().add(1, 'year').set('millisecond', 0).toISOString()
--- a/cypress/integration/oauth2/grant_jwtbearer.js
+++ b/cypress/integration/oauth2/grant_jwtbearer.js
@@ -76,9 +76,10 @@ describe('The OAuth 2.0 JWT Bearer (RFC 7523) Grant', function () {
     response_types: ['token']
   })
 
-  const gr = (subject) => ({
+  const gr = (subject, domain) => ({
     issuer: prng(),
     subject: subject,
+    domain: domain,
     scope: ['foo', 'openid', 'offline_access'],
     jwk: testPublicJwk,
     expires_at: dayjs().utc().add(1, 'year').set('millisecond', 0).toISOString()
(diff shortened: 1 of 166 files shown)


jagobagascon:
I did some changes to the initial implementation:

MySQL was failing because there's a limit on the size of an index, so I had to change the unique index on MySQL to contain only the first few characters of each column: `UNIQUE (issuer(128), subject(128), allowed_domain(128), key_id)`. It should be quite rare to use a domain or an issuer longer than 30 characters, but I didn't make it smaller because the RFC states that [domains can contain up to 255 characters](https://datatracker.ietf.org/doc/html/rfc1034#:~:text=To%20simplify%20implementations%2C%20the%20total%20number%20of%20octets%20that%20represent%20a%0Adomain%20name%20(i.e.%2C%20the%20sum%20of%20all%20label%20octets%20and%20label%20lengths)%20is%0Alimited%20to%20255.). I'm not sure if I like this so I would like to hear what others think about this.

I also renamed the `domain` field to `allowed_domain` in both, the database and the service.

Regarding the tests, I'm going to need some help. I added new end-to-end tests and also updated the request validation tests. And both are working fine. But for some reason some tests are failing in CircleCI:
```
    --- FAIL: TestManagers/package=consent/janitor=mysql (0.46s)
        --- FAIL: TestManagers/package=consent/janitor=mysql/case=flush-consent-request-limit (0.09s)
            --- FAIL: TestManagers/package=consent/janitor=mysql/case=flush-consent-request-limit/case=limit (0.02s)
                --- FAIL: TestManagers/package=consent/janitor=mysql/case=flush-consent-request-limit/case=limit/step=validate (0.00s)
                    janitor_test_helper.go:246: check login: limit_flush-login-1
                    janitor_test_helper.go:246: check login: limit_flush-login-2
                    janitor_test_helper.go:246: check login: limit_flush-login-3
                    janitor_test_helper.go:251: 
                        	Error Trace:	janitor_test_helper.go:251
                        	Error:      	An error is expected but got nil.
                        	Test:       	TestManagers/package=consent/janitor=mysql/case=flush-consent-request-limit/case=limit/step=validate
```

I've tried to run them locally but they are working fine for me.

aeneasr (changes requested):
Hello there, sorry for the very late review, I was very busy with other stuff :(

Regarding this PR I unfortunately have some reservations. It is considered bad practice to use emails as subject identifiers in all of OAuth2 and OpenID Connect for several reasons:

1. In OpenID Connect, the email can only be "leaked" if the `email` scope was requested and granted. Without it, the email **must not** be part of the ID Token. I am not sure whether this change would impact that but having this feature implemented as such will draw users of Ory Hydra to the conclusion that it is ok to use the email as the sub field.
2. In OAuth2 it is considered bad practice to use mutable identifiers as the subject field. We have a blog post about a company which did use the email/username as the sub field and ended up having a pretty serious vulnerability: https://www.ory.sh/sign-in-with-user-impersonation-oauth2-openid-connect/

Therefore, I do not think that we can accept this change as it is, and we should probably update the documentation around this feature to not use emails in the first place (so e.g. here https://github.com/ory/hydra/pull/2957/files#diff-448af211211e279215cf0452c8dec6532e51ed9a94049278d54e6f7b5ddaa364R63).

I am really sorry for all the work you've put in so far. It's my fault for having said "good to go" on the issue description. Maybe there is a way to leverage all the work here with a different approach towards allowing multiple subject IDs?

jagobagascon:
It's ok, I just had some time and made an attempt at implementing this. Thank you for your time & review, let's keep the discussion in #2930.

PR #2977, opened 2022-02-09
fix: revert back to PATs

icyphox:
Due to how branch protection works, we can't push without a token that
has permissions to bypass branch protection. Hence, we revert back to
PATs here.

However, we also want to prevent triggering recursive workflow runs
which can potentially occur in commits originating from a non-GitHub
Actions actor. One possible workaround would be to add a `[skip ci]` to
the body in autogen commits. This prevents any workflow from triggering
on the commit. See: https://github.com/ory/ci/commit/849d2713e4b24ee3540b1f752ad1a4094d634e2f

Let me know what you think.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -39,7 +39,7 @@ jobs:
     steps:
       - uses: ory/ci/sdk/generate@master
         with:
-          token: ${{ github.token }}
+          token: ${{ secrets.ORY_BOT_PAT }}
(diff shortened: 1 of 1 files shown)


aeneasr:
ci skip for autogen is an option, as long as it does not interfere with the PR status reporting and only runs on master :)

icyphox:
@aeneasr yup, it's only on master.

aeneasr:
Ok, let’s try it!

PR #2991, opened 2022-02-15
ci: use ory/ci/checkout

icyphox:
Switch to ory/ci/checkout. Ref: https://github.com/ory/ci/commit/192e56872f0d231e9df6f3b66df203e4539bf2ae


aeneasr:
Nice!

PR #3003, opened 2022-02-21
fix: pass token to render-version-schema

icyphox:
Fixes the branch protection error.


aeneasr:
@icyphox these have to be applied to the other projects too right?

icyphox:
@aeneasr not yet. Will do.

PR #3010, opened 2022-02-24
fix: Tokens issued for urn:ietf:params:oauth:grant-type:jwt-bearer grant have incomplete session / token

akrause-cel:
Tokens issued with urn:ietf:params:oauth:grant-type:jwt-bearer grant have incomplete session / token. In case of a JWT strategy they miss:

- Header Key ID
- Claims Issuer
- Client ID

A Sample is provided below:

Header:
```
{
  "alg": "RS256",
  "kid": "",
  "typ": "JWT"
}
```
Claims:
```
{
  "aud": [
    "http://localhost:4444/oauth2/token"
  ],
  "client_id": "",
  "exp": 1645700662,
  "ext": {},
  "iat": 1645697061,
  "jti": "d497e7d4-73ea-419f-938b-5dda67ed2898",
  "nbf": 1645697061,
  "scp": [
    "something"
  ],
  "sub": "5ed812dc-939e-11ec-b909-0242ac120004"
}
```
## Related issue(s)

This is a fix for a previously unknown bug. 

Steps to reproduce:
1. Launch Hydra with environment Variable `STRATEGIES_ACCESS_TOKEN` set to  `jwt`
1. Follow the guide to setup the JWT Bearer Grant: https://www.ory.sh/docs/hydra/guides/oauth2-grant-type-jwt-bearer#exchanging-jwts-for-access-tokens
2. Itrospect resulting token

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.


aeneasr:
Thank you for this PR and fix, could you please add a test for this to confirm that it is a bug and this is the fix? Thank you! :)

akrause-cel:
hi @aeneasr, I could not find any tests for this grant type and hence did not build it up from scratch to fix it for those minor changes. As it seems this would be quite some effort. Could you support with a skeleton here?

akrause-cel:
I looked into this now, afaik it is not possible to register JWK for testing purposes (unlike e.g. Clients:`require.NoError(t, reg.ClientManager().CreateClient(context.TODO(), c))`. 

Making the code testable from my PoV would be a significant rewrite of the existing codebase which is probably why the test was missing in the first place. Can we merge this and work on testability in a separate issue.

grantzvolsky:
I've been working with tests a lot recently, so I could write the test if we want this before 2.x

aeneasr:
Sorry, I missed the questions here. We do have tests for this:

https://github.com/ory/hydra/blob/master/cypress/integration/oauth2/grant_jwtbearer.js

Also, a good place to add another test on this in Go code (easier to debug) would be here:

https://github.com/ory/hydra/blob/master/oauth2/oauth2_client_credentials_test.go

You can take a look here

https://github.com/ory/hydra/blob/e3236dd9f6febbb2c147417f6c21a7285418e055/oauth2/trust/manager_test_helpers.go#L37

to see how to manage trust relationships :)

Even if these changes are considered minor, tests are a hard requirement for merging anything into Ory :) Otherwise there's no formal way to verify that the changes work and verify that they continue to work long after this PR was merged. Missing tests for this were probably the cause of the bug in the first place :)

aeneasr:
@akrause-cel let me know when this is ready for another review, or if you need any further help!

akrause-cel:
Will look into it. However, quite busy with other things at the moment.

aeneasr:
No problem, take your time :)

aeneasr:
@akrause-cel I looked into the client auth optional thing, and yes the problem is the foreign key here. I think we require the client to be set across the code base, and there isn't an easy remediation to resolve this. It requires looking into all the places where we have assumptions about the client being set, and validating whether that assumption is correct there. So it is quite some work to get this particular feature done. However, I'll add a test (it is skipped at the moment) which can be enabled to test any future implementation of this.

PR #2397, opened 2021-03-10
fix: Remove 404 on revoke consent

NikolaySl:
## Related issue

No

## Proposed changes

Do not throw 404 on revoke consent sessions of subject - https://www.ory.sh/hydra/docs/reference/api/#revokes-consent-sessions-of-a-subject-for-a-specific-oauth-20-client

The same changes and as in https://github.com/ory/hydra/issues/1168 (https://github.com/ory/hydra/pull/1173)

## Checklist

- [x ] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [ x] I have read the [security policy](../security/policy).
- [x ] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security. vulnerability, I
      confirm that I got green light (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [x ] I have added tests that prove my fix is effective or that my feature
      works.
- [ x] I have added or changed [the documentation](docs/docs).

PR #3013, opened 2022-02-28
feat: nid

grantzvolsky:
This is a draft PR for the network ID feature. The work is based on current 2.x (0376ac31) rebased on master (b1776722) with resolved merge conflicts.

## Finalized

- [x] https://github.com/ory/hydra/pull/3013#pullrequestreview-915679585

## TODO

- [x] https://github.com/ory/hydra/pull/3013#discussion_r841056028

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.


Benehiko (changes requested):
This PR looks great! :)

The only thing that still needs thorough review are the sql migration files (the core logic of this PR).

Things still that need a bit of cleanup are:
- [ ] Database migration script needs some cleanup (detatched head and  `TEST_DATABASE_COCKROACHDB` env variable).
- [ ] NID needs randomization.

aeneasr (changes requested):

client/validator.go
```diff
@@ -52,25 +53,28 @@ var (
...
 type Validator struct {
-	c    *http.Client
-	conf *config.Provider
+	c              *http.Client
+	contextualizer contextx.Contextualizer
+	defaultConf    *config.Provider
 }
 
-func NewValidator(conf *config.Provider) *Validator {
+func NewValidator(contextualizerProvider contextx.Contextualizer, defaultConf *config.Provider) *Validator {
```
aeneasr, line 61: Would it make sense to pass the registry here instead? And then do away with defaultConf and just use `reg.Config(ctx)`?

aeneasr (changes requested):

persistence/sql/persister_client.go
```diff
@@ -73,15 +80,16 @@ func (p *Persister) DeleteClient(ctx context.Context, id string) error {
 		return err
 	}
 
-	return sqlcon.HandleError(p.Connection(ctx).Destroy(&client.Client{ID: cl.ID}))
+	return sqlcon.HandleError(p.Connection(ctx).Destroy(&client.Client{ID: cl.ID, NID: p.NetworkID(ctx)}))
```
aeneasr, line 83: Destroy only uses the primary key, so the NID doesn't have an effect here. You will need to use RawQUERY instead

persistence/sql/persister_consent.go
```diff
@@ -17,30 +18,28 @@ import (
...
-		hrs := make([]*consent.HandledConsentRequest, 0)
+		fs := make([]*flow.Flow, 0)
 		if err := c.
 			Where(whereStmt, whereArgs...).
-			Select("r.challenge").
-			Join(
-				fmt.Sprintf("%s AS r", consent.ConsentRequest{}.TableName()),
-				fmt.Sprintf("r.challenge = %s.challenge", consent.HandledConsentRequest{}.TableName())).
-			All(&hrs); err != nil {
+			Select("consent_challenge_id").
```
aeneasr, line 41: Isn't this missing the nid?

persistence/sql/persister_consent.go
```diff
@@ -377,12 +446,16 @@ func (p *Persister) listUserAuthenticatedClients(ctx context.Context, subject, s
 	return cs, p.transaction(ctx, func(ctx context.Context, c *pop.Connection) error {
 		if err := c.RawQuery(
 			/* #nosec G201 - channel can either be "front" or "back" */
-			fmt.Sprintf(`SELECT DISTINCT c.* FROM hydra_client as c JOIN hydra_oauth2_consent_request as r ON (c.id = r.client_id) WHERE r.subject=? AND c.%schannel_logout_uri!='' AND c.%schannel_logout_uri IS NOT NULL AND r.login_session_id = ?`,
+			fmt.Sprintf(`
+SELECT DISTINCT c.* FROM hydra_client as c
+JOIN hydra_oauth2_flow as f ON (c.id = f.client_id)
+WHERE f.subject=? AND c.%schannel_logout_uri!='' AND c.%schannel_logout_uri IS NOT NULL AND f.login_session_id = ? AND f.nid = ?`,
```
aeneasr, line 452: Should this also have a constraint for c.NID?

persistence/sql/persister_oauth2.go
```diff
@@ -177,12 +179,9 @@ func (p *Persister) ClientAssertionJWTValid(ctx context.Context, jti string) err
...
 	return p.transaction(ctx, func(ctx context.Context, c *pop.Connection) error {
 		// delete expired
-		now := "now()"
-		if c.Dialect.Name() == "sqlite3" {
-			now = "CURRENT_TIMESTAMP"
-		}
+		nid := p.NetworkID(ctx)
 		/* #nosec G201 table is static */
-		if err := c.RawQuery(fmt.Sprintf("DELETE FROM %s WHERE expires_at < %s", oauth2.BlacklistedJTI{}.TableName(), now)).Exec(); err != nil {
+		if err := c.RawQuery(fmt.Sprintf("DELETE FROM %s WHERE expires_at < CURRENT_TIMESTAMP AND nid = ?", oauth2.BlacklistedJTI{}.TableName()), nid).Exec(); err != nil {
```
aeneasr, line 184: Is `CURRENT_TIMESTAMP` supported everywhere? It looks like we used `now` before.

aeneasr (changes requested):
I found a few more items, but otherwise this looks pretty good!

persistence/sql/src/20220210000001_nid/20220210000001000000_nid.cockroach.up.sql
```diff
@@ -0,0 +1,212 @@
...
+ALTER TABLE hydra_client ADD COLUMN "nid" UUID;
+ALTER TABLE hydra_client ADD CONSTRAINT "hydra_client_nid_fk_idx" FOREIGN KEY ("nid") REFERENCES "networks" ("id") ON UPDATE RESTRICT ON DELETE CASCADE;
+--split
+UPDATE hydra_client SET nid = (SELECT id FROM networks LIMIT 1);
+--split
+ALTER TABLE hydra_client ALTER nid SET NOT NULL;
+--split
+DROP INDEX hydra_client_id_key CASCADE;
+--split
+CREATE UNIQUE INDEX hydra_client_id_key ON hydra_client (id ASC, nid ASC);
```
aeneasr, line 11: Add a key over id,nid for fast lookup

aeneasr (changes requested):
Please add indexes for `nid` which are regular indices to speed up WHERE filters

persistence/sql/persister_oauth2.go
grantzvolsky: `CURRENT_TIMESTAMP` is ANSI SQL and it is supported by all our db engines. `now` is a common alias for `CURRENT_TIMESTAMP`, but it isn't defined in sqlite3.

consent/strategy_default.go
```diff
@@ -609,11 +591,6 @@ func (s *DefaultStrategy) verifyConsent(w http.ResponseWriter, r *http.Request,
 		return nil, err
 	}
```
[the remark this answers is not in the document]
grantzvolsky: "Note to self: this is ok because it is handled in the outer function" - https://github.com/ory/hydra/pull/2836/files#r754383563

consent/strategy_default.go
```diff
@@ -626,7 +603,6 @@ func (s *DefaultStrategy) verifyConsent(w http.ResponseWriter, r *http.Request,
 		session.Session.IDToken = map[string]interface{}{}
 	}
 
-	session.ConsentRequest.SubjectIdentifier = pw
 	session.AuthenticatedAt = session.ConsentRequest.AuthenticatedAt
```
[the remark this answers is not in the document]
grantzvolsky: "Note to self: this is ok because it is handled in the outer function" - https://github.com/ory/hydra/pull/2836/files#r754383563

client/manager_test_helpers.go
```diff
@@ -121,27 +153,48 @@ func TestHelperCreateGetUpdateDeleteClient(k string, m Storage) func(t *testing.
 			BackChannelLogoutSessionRequired:  true,
 		}
 
-		require.NoError(t, m.CreateClient(ctx, c))
-		assert.Equal(t, c.GetID(), "1234")
+		require.NoError(t, t1.CreateClient(ctx, t1c1))
+		{
+			t2c1 := *t1c1
+			require.Error(t, t2.CreateClient(ctx, &t2c1), "should not be able to create the same client in other manager/tenant; are they backed by the same database?")
```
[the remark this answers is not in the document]
grantzvolsky: Good catch! This code shouldn't work because we exclude `pk` in the Create function. The reason why it does work is that `pk` is a UUID-typed primary key, leading to [this](https://github.com/gobuffalo/pop/blob/c285511d17469aa430ab08202a6f2353c3ae19e5/dialect_common.go#L81) line that adds `pk` despite its exclusion from `cols`.

This is the only place in Hydra where we rely on `Create` to exclude a field in order to generate a random ID, so I added `c.ID = uuid.Must(uuid.NewV4())` in CreateClient persister + a test.

persistence/sql/src/20220210000001_nid/20220210000001000000_nid.sqlite.up.sql
```diff
@@ -0,0 +1,465 @@
...
+DROP TABLE "hydra_client";
+ALTER TABLE "_hydra_client_tmp" RENAME TO "hydra_client";
+
+UPDATE hydra_client SET nid = (SELECT id FROM networks LIMIT 1);
+
+CREATE UNIQUE INDEX hydra_client_id_nid_idx ON hydra_client (id, nid);
```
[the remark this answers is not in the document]
grantzvolsky: Just to be sure, this only affects sqlite, right? The documentation for postgres and cockroach both mention the '[UNIQUE]' bit as something merely creates an additional constraint.

persistence/sql/persister_consent.go
```diff
@@ -264,103 +319,117 @@ func (p *Persister) ConfirmLoginSession(ctx context.Context, id string, authenti
...
 }
 
 func (p *Persister) DeleteLoginSession(ctx context.Context, id string) error {
-	return sqlcon.HandleError(
-		p.Connection(ctx).Destroy(
-			&consent.LoginSession{ID: id},
-		))
+	count, err := p.Connection(ctx).RawQuery("DELETE FROM hydra_oauth2_authentication_session WHERE id=? AND nid = ?", id, p.NetworkID(ctx)).ExecWithCount()
+	if count == 0 {
+		return errorsx.WithStack(x.ErrNotFound)
```
aeneasr, line 337: You could introduce a new method DestroyWithNetwork() which internally uses QueryWith... + Delete()

persistence/sql/persister_consent.go
```diff
@@ -392,41 +465,46 @@ func (p *Persister) listUserAuthenticatedClients(ctx context.Context, subject, s
...
 	return p.GetLogoutRequest(ctx, challenge)
 }
 
 func (p *Persister) RejectLogoutRequest(ctx context.Context, challenge string) error {
-	return errorsx.WithStack(
-		p.Connection(ctx).
-			RawQuery("UPDATE hydra_oauth2_logout_request SET rejected=true, accepted=false WHERE challenge=?", challenge).
-			Exec())
+	count, err := p.Connection(ctx).
+		RawQuery("UPDATE hydra_oauth2_logout_request SET rejected=true, accepted=false WHERE challenge=? AND nid = ?", challenge, p.NetworkID(ctx)).
```
aeneasr, line 482: QueryWith...

persistence/sql/persister_consent.go
```diff
@@ -392,41 +465,46 @@ func (p *Persister) listUserAuthenticatedClients(ctx context.Context, subject, s
...
 	var lr consent.LogoutRequest
-	return &lr, sqlcon.HandleError(p.Connection(ctx).Where("challenge = ? AND rejected = FALSE", challenge).First(&lr))
+	return &lr, sqlcon.HandleError(p.Connection(ctx).Where("challenge = ? AND rejected = FALSE AND nid = ?", challenge, p.NetworkID(ctx)).First(&lr))
 }
 
 func (p *Persister) VerifyAndInvalidateLogoutRequest(ctx context.Context, verifier string) (*consent.LogoutRequest, error) {
 	var lr consent.LogoutRequest
 	return &lr, p.transaction(ctx, func(ctx context.Context, c *pop.Connection) error {
-		if err := c.Where("verifier=? AND was_used=FALSE AND accepted=TRUE AND rejected=FALSE", verifier).Select("challenge").First(&lr); err != nil {
+		if err := c.Where("verifier=? AND was_used=FALSE AND accepted=TRUE AND rejected=FALSE AND nid = ?", verifier, p.NetworkID(ctx)).Select("challenge").First(&lr); err != nil {
```
aeneasr, line 499: QueryWith...

persistence/sql/persister_consent.go
```diff
@@ -392,41 +465,46 @@ func (p *Persister) listUserAuthenticatedClients(ctx context.Context, subject, s
...
 			if err == sql.ErrNoRows {
 				return errorsx.WithStack(x.ErrNotFound)
 			}
 
 			return sqlcon.HandleError(err)
 		}
 
-		if err := c.RawQuery("UPDATE hydra_oauth2_logout_request SET was_used=TRUE WHERE verifier=?", verifier).Exec(); err != nil {
+		if err := c.RawQuery("UPDATE hydra_oauth2_logout_request SET was_used=TRUE WHERE verifier=? AND nid = ?", verifier, p.NetworkID(ctx)).Exec(); err != nil {
```
aeneasr, line 507: QueryWith...

persistence/sql/persister_grant_jwk.go
```diff
@@ -32,14 +32,15 @@ func (p *Persister) CreateGrant(ctx context.Context, g trust.Grant, publicKey jo
 		}
 
 		data := p.sqlDataFromJWTGrant(g)
+		data.NID = p.NetworkID(ctx)
 
 		return sqlcon.HandleError(p.Connection(ctx).Create(&data))
```
aeneasr, line 37: CreateWith

persistence/sql/persister_grant_jwk.go
```diff
@@ -32,14 +32,15 @@ func (p *Persister) CreateGrant(ctx context.Context, g trust.Grant, publicKey jo
...
 	})
 }
 
 func (p *Persister) GetConcreteGrant(ctx context.Context, id string) (trust.Grant, error) {
 	var data trust.SQLData
-	if err := p.Connection(ctx).Where("id = ?", id).First(&data); err != nil {
+	if err := p.Connection(ctx).Where("id = ? AND nid = ?", id, p.NetworkID(ctx)).First(&data); err != nil {
```
aeneasr, line 43: QueryWith...

persistence/sql/persister_grant_jwk.go
```diff
@@ -64,7 +65,10 @@ func (p *Persister) DeleteGrant(ctx context.Context, id string) error {
 func (p *Persister) GetGrants(ctx context.Context, limit, offset int, optionalIssuer string) ([]trust.Grant, error) {
```
aeneasr, line 65: ^- DELETEGRANT IS MISSING NID CHECK!

persistence/sql/persister_grant_jwk.go
```diff
@@ -64,7 +65,10 @@ func (p *Persister) DeleteGrant(ctx context.Context, id string) error {
...
 	grantsData := make([]trust.SQLData, 0)
 
-	query := p.Connection(ctx).Paginate(offset/limit+1, limit).Order("id")
+	query := p.Connection(ctx).
+		Where("nid = ?", p.NetworkID(ctx)).
```
aeneasr, line 69: QueryWith...

persistence/sql/persister_grant_jwk.go
```diff
@@ -82,7 +86,9 @@ func (p *Persister) GetGrants(ctx context.Context, limit, offset int, optionalIs
 }
 
 func (p *Persister) CountGrants(ctx context.Context) (int, error) {
-	n, err := p.Connection(ctx).Count(&trust.SQLData{})
+	n, err := p.Connection(ctx).
+		Where("nid = ?", p.NetworkID(ctx)).
```
aeneasr, line 90: QueryWith...

persistence/sql/persister_grant_jwk.go
```diff
@@ -91,7 +97,8 @@ func (p *Persister) GetPublicKey(ctx context.Context, issuer string, subject str
 	query := p.Connection(ctx).
 		Where("issuer = ?", issuer).
 		Where("subject = ?", subject).
-		Where("key_id = ?", keyId)
+		Where("key_id = ?", keyId).
+		Where("nid = ?", p.NetworkID(ctx))
```
aeneasr, line 101: QueryWith...

persistence/sql/persister_grant_jwk.go
```diff
@@ -108,7 +115,8 @@ func (p *Persister) GetPublicKeys(ctx context.Context, issuer string, subject st
 	grantsData := make([]trust.SQLData, 0)
 	query := p.Connection(ctx).
 		Where("issuer = ?", issuer).
-		Where("subject = ?", subject)
+		Where("subject = ?", subject).
+		Where("nid = ?", p.NetworkID(ctx))
```
aeneasr, line 119: QueryWith...

persistence/sql/persister_grant_jwk.go
```diff
@@ -139,7 +147,8 @@ func (p *Persister) GetPublicKeyScopes(ctx context.Context, issuer string, subje
 	query := p.Connection(ctx).
 		Where("issuer = ?", issuer).
 		Where("subject = ?", subject).
-		Where("key_id = ?", keyId)
+		Where("key_id = ?", keyId).
+		Where("nid = ?", p.NetworkID(ctx))
```
aeneasr, line 151: QueryWith...

persistence/sql/persister_grant_jwk.go
```diff
@@ -194,7 +203,8 @@ func (p *Persister) FlushInactiveGrants(ctx context.Context, notAfter time.Time,
 		deleteUntil = notAfter
 	}
 	return sqlcon.HandleError(p.Connection(ctx).RawQuery(
```
aeneasr, line 205: QueryWith...

persistence/sql/persister_jwk.go
```diff
@@ -49,6 +49,7 @@ func (p *Persister) AddKey(ctx context.Context, set string, key *jose.JSONWebKey
 	}
 
 	return sqlcon.HandleError(p.Connection(ctx).Create(&jwk.SQLData{
```
aeneasr, line 51: CreateWith...

persistence/sql/persister_jwk.go
```diff
@@ -70,6 +71,7 @@ func (p *Persister) AddKeySet(ctx context.Context, set string, keys *jose.JSONWe
 			}
 
 			if err := c.Create(&jwk.SQLData{
+				NID:     p.NetworkID(ctx),
```
aeneasr, line 74: CreateWith...

persistence/sql/persister_jwk.go
```diff
@@ -84,7 +86,7 @@ func (p *Persister) AddKeySet(ctx context.Context, set string, keys *jose.JSONWe
 
 func (p *Persister) UpdateKey(ctx context.Context, set string, key *jose.JSONWebKey) error {
 	return p.transaction(ctx, func(ctx context.Context, c *pop.Connection) error {
-		if err := p.DeleteKey(ctx, set, key.KeyID); err != nil {
+		if err := p.DeleteKey(ctx, set, key.KeyID); err != nil && errorsx.Cause(err) != x.ErrNotFound {
```
aeneasr, line 89: This check is new. Did something regress? Also, please use `errors.Is()`

persistence/sql/persister_jwk.go
```diff
@@ -96,7 +98,7 @@ func (p *Persister) UpdateKey(ctx context.Context, set string, key *jose.JSONWeb
 
 func (p *Persister) UpdateKeySet(ctx context.Context, set string, keySet *jose.JSONWebKeySet) error {
 	return p.transaction(ctx, func(ctx context.Context, c *pop.Connection) error {
-		if err := p.DeleteKeySet(ctx, set); err != nil {
+		if err := p.DeleteKeySet(ctx, set); err != nil && errorsx.Cause(err) != x.ErrNotFound {
```
aeneasr, line 101: This check is new. Did something regress? Also, please use `errors.Is()`

persistence/sql/persister_jwk.go
```diff
@@ -109,7 +111,7 @@ func (p *Persister) UpdateKeySet(ctx context.Context, set string, keySet *jose.J
 func (p *Persister) GetKey(ctx context.Context, set, kid string) (*jose.JSONWebKeySet, error) {
 	var j jwk.SQLData
 	if err := p.Connection(ctx).
-		Where("sid = ? AND kid = ?", set, kid).
+		Where("sid = ? AND kid = ? AND nid = ?", set, kid, p.NetworkID(ctx)).
```
aeneasr, line 114: QueryWith...

persistence/sql/persister_jwk.go
```diff
@@ -133,7 +135,7 @@ func (p *Persister) GetKey(ctx context.Context, set, kid string) (*jose.JSONWebK
 func (p *Persister) GetKeySet(ctx context.Context, set string) (*jose.JSONWebKeySet, error) {
 	var js []jwk.SQLData
 	if err := p.Connection(ctx).
-		Where("sid = ?", set).
+		Where("sid = ? AND nid = ?", set, p.NetworkID(ctx)).
```
aeneasr, line 138: QueryWith...

persistence/sql/persister_jwk.go
```diff
@@ -165,9 +167,19 @@ func (p *Persister) GetKeySet(ctx context.Context, set string) (*jose.JSONWebKey
 }
 
 func (p *Persister) DeleteKey(ctx context.Context, set, kid string) error {
-	return sqlcon.HandleError(p.Connection(ctx).RawQuery("DELETE FROM hydra_jwk WHERE sid=? AND kid=?", set, kid).Exec())
+	count, err := p.Connection(ctx).RawQuery("DELETE FROM hydra_jwk WHERE sid=? AND kid=? AND nid = ?", set, kid, p.NetworkID(ctx)).ExecWithCount()
```
aeneasr, line 170: QueryWith...

persistence/sql/persister_jwk.go
```diff
@@ -165,9 +167,19 @@ func (p *Persister) GetKeySet(ctx context.Context, set string) (*jose.JSONWebKey
...
+	if count == 0 {
```
aeneasr, line 171: These checks are new. This is considered a regression. DELETE in REST usually always returns 204 No Content because the deletion succeeded if the object does not exist (because nothing was there to delete in the first place). I would suggest to revert these checks with count == 0 whereever they were introduced UNLESS there is a very good reason to do this, and it is documented as a brekaing change. If it is, please link the location of that documentation in this comment.

persistence/sql/persister_jwk.go
```diff
@@ -165,9 +167,19 @@ func (p *Persister) GetKeySet(ctx context.Context, set string) (*jose.JSONWebKey
...
 }
 
 func (p *Persister) DeleteKeySet(ctx context.Context, set string) error {
-	return sqlcon.HandleError(p.Connection(ctx).RawQuery("DELETE FROM hydra_jwk WHERE sid=?", set).Exec())
+	count, err := p.Connection(ctx).RawQuery("DELETE FROM hydra_jwk WHERE sid=? AND nid = ?", set, p.NetworkID(ctx)).ExecWithCount()
+	if count == 0 {
+		return errorsx.WithStack(x.ErrNotFound)
+	} else {
+		return sqlcon.HandleError(err)
+	}
```
aeneasr, line 184: See comment above

persistence/sql/persister_oauth2.go
aeneasr, line 184: QueryWith

persistence/sql/persister_oauth2.go
```diff
@@ -203,6 +202,7 @@ func (p *Persister) GetClientAssertionJWT(ctx context.Context, j string) (*oauth
 }
```
aeneasr, line 202: Missing NID!

persistence/sql/persister_oauth2.go
```diff
@@ -203,6 +202,7 @@ func (p *Persister) GetClientAssertionJWT(ctx context.Context, j string) (*oauth
...
 
 func (p *Persister) SetClientAssertionJWTRaw(ctx context.Context, jti *oauth2.BlacklistedJTI) error {
+	jti.NID = p.NetworkID(ctx)
 	return sqlcon.HandleError(p.Connection(ctx).Create(jti))
```
aeneasr, line 206: CreateWith...

persistence/sql/persister_oauth2.go
```diff
@@ -212,6 +212,8 @@ func (p *Persister) createSession(ctx context.Context, signature string, request
 		return err
 	}
 
+	req.NID = p.NetworkID(ctx)
+
 	if err := sqlcon.HandleError(p.Connection(ctx).Create(req)); errors.Is(err, sqlcon.ErrConcurrentUpdate) {
```
aeneasr, line 217: CreateWith...

persistence/sql/persister_oauth2.go
```diff
@@ -227,7 +229,7 @@ func (p *Persister) findSessionBySignature(ctx context.Context, rawSignature str
 	var fr fosite.Requester
 
 	return fr, p.transaction(ctx, func(ctx context.Context, c *pop.Connection) error {
-		err := p.Connection(ctx).Where("signature = ?", rawSignature).First(&r)
+		err := p.Connection(ctx).Where("signature = ? AND nid = ?", rawSignature, p.NetworkID(ctx)).First(&r)
```
aeneasr, line 232: QueryWith

persistence/sql/persister_oauth2.go
```diff
@@ -254,15 +256,16 @@ func (p *Persister) deleteSessionBySignature(ctx context.Context, signature stri
 	/* #nosec G201 table is static */
 	return sqlcon.HandleError(
 		p.Connection(ctx).
-			RawQuery(fmt.Sprintf("DELETE FROM %s WHERE signature=?", OAuth2RequestSQL{Table: table}.TableName()), signature).
+			RawQuery(fmt.Sprintf("DELETE FROM %s WHERE signature=? AND nid = ?", OAuth2RequestSQL{Table: table}.TableName()), signature, p.NetworkID(ctx)).
```
aeneasr, line 259: QueryWith

persistence/sql/persister_oauth2.go
```diff
@@ -254,15 +256,16 @@ func (p *Persister) deleteSessionBySignature(ctx context.Context, signature stri
...
 			Exec())
 }
 
 func (p *Persister) deleteSessionByRequestID(ctx context.Context, id string, table tableName) error {
 	/* #nosec G201 table is static */
 	if err := p.Connection(ctx).RawQuery(
-		fmt.Sprintf("DELETE FROM %s WHERE request_id=?", OAuth2RequestSQL{Table: table}.TableName()),
+		fmt.Sprintf("DELETE FROM %s WHERE request_id=? AND nid = ?", OAuth2RequestSQL{Table: table}.TableName()),
```
aeneasr, line 266: QueryWith

persistence/sql/persister_oauth2.go
```diff
@@ -281,8 +284,9 @@ func (p *Persister) deactivateSessionByRequestID(ctx context.Context, id string,
 	return sqlcon.HandleError(
 		p.Connection(ctx).
 			RawQuery(
-				fmt.Sprintf("UPDATE %s SET active=false WHERE request_id=?", OAuth2RequestSQL{Table: table}.TableName()),
+				fmt.Sprintf("UPDATE %s SET active=false WHERE request_id=? AND nid = ?", OAuth2RequestSQL{Table: table}.TableName()),
```
aeneasr, line 287: QueryWith

persistence/sql/persister_oauth2.go
```diff
@@ -301,9 +305,12 @@ func (p *Persister) InvalidateAuthorizeCodeSession(ctx context.Context, signatur
 	return sqlcon.HandleError(
 		p.Connection(ctx).
 			RawQuery(
```
aeneasr, line 307: QueryWith

persistence/sql/persister_oauth2.go
```diff
@@ -378,9 +385,10 @@ func (p *Persister) flushInactiveTokens(ctx context.Context, notAfter time.Time,
 
 	// Select tokens' signatures with limit
 	q := p.Connection(ctx).RawQuery(
-		fmt.Sprintf("SELECT signature FROM %s WHERE requested_at < ? ORDER BY signature LIMIT %d",
+		fmt.Sprintf("SELECT signature FROM %s WHERE requested_at < ? AND nid = ? ORDER BY signature LIMIT %d",
```
aeneasr, line 388: QueryWith

persistence/sql/persister_oauth2.go
```diff
@@ -421,6 +430,11 @@ func (p *Persister) DeleteAccessTokens(ctx context.Context, clientID string) err
 	/* #nosec G201 table is static */
 	return sqlcon.HandleError(
 		p.Connection(ctx).
-			RawQuery(fmt.Sprintf("DELETE FROM %s WHERE client_id=?", OAuth2RequestSQL{Table: sqlTableAccess}.TableName()), clientID).
-			Exec())
+			RawQuery(
+				fmt.Sprintf("DELETE FROM %s WHERE client_id=? AND nid = ?", OAuth2RequestSQL{Table: sqlTableAccess}.TableName()),
```
aeneasr, line 434: QueryWith

persistence/sql/persister_oauth2.go
```diff
@@ -398,8 +406,9 @@ func (p *Persister) flushInactiveTokens(ctx context.Context, notAfter time.Time,
 
 		if i != j {
 			err = p.Connection(ctx).RawQuery(
-				fmt.Sprintf("DELETE FROM %s WHERE signature in (?)", OAuth2RequestSQL{Table: table}.TableName()),
+				fmt.Sprintf("DELETE FROM %s WHERE signature in (?) AND nid = ?", OAuth2RequestSQL{Table: table}.TableName()),
```
aeneasr, line 409: QueryWith

aeneasr (changes requested):
I found a few places where we forgot to set the NID. Also, I would like to ask you to use the approach patrik has made in Ory Keto as it is much cleaner and would allow us in the future to narrow down the interfaces to use nid at all times and prevent raw use of `Delete()` or `Create()`

flow/flow.go
```diff
@@ -0,0 +1,475 @@
...
+	if f.Client != nil {
+		f.ClientID = f.Client.OutfacingID
+	}
+	if f.State == FlowStateLoginUnused && string(f.Context) == "" {
+		f.Context = sqlxx.JSONRawMessage("{}")
+	}
+	return nil
+}
+
+// TODO Populate the client field in FindInDB and FindByConsentChallengeID in
```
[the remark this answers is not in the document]
grantzvolsky: It isn't tracked at the moment. We talked about this back when I was in Munich, and the conclusion was that there is no reasonable way to do this with Pop due to the foreign key not being a primary key, so the best we can do is to write it as raw sql. Considering the number of columns in Flow and Client, I was reluctant to do that and added a TODO instead, hoping to find a better solution. Can you think of a solution? Should I just write the massive raw query?

persistence/sql/migratest/fixtures/networks/00000000-0000-0000-0000-000000000000.json
```diff
@@ -0,0 +1,3 @@
+{
+  "id": "00000000-0000-0000-0000-000000000000"
```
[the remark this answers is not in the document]
grantzvolsky: There is an assertion in the test to make sure the NID is not zero. We can't have it in the fixture because it is randomly generated by the migration that is being tested.

persistence/sql/persister_consent.go
```diff
@@ -49,21 +48,20 @@ func (p *Persister) revokeConsentSession(whereStmt string, whereArgs ...interfac
...
-			if err := p.RevokeRefreshToken(ctx, hr.ID); errors.Is(err, fosite.ErrNotFound) {
+			if err := p.RevokeRefreshToken(ctx, f.ConsentChallengeID.String()); errors.Is(err, fosite.ErrNotFound) {
 				// do nothing
 			} else if err != nil {
 				return err
 			}
 
-			// Since we ON DELETE CASCADE, hydra_oauth2_consent_request_handled will be removed automagically.
-			localCount, err := c.RawQuery("DELETE FROM hydra_oauth2_consent_request WHERE challenge = ?", hr.ID).ExecWithCount()
+			localCount, err := c.RawQuery("DELETE FROM hydra_oauth2_flow WHERE consent_challenge_id = ? AND nid = ?", f.ConsentChallengeID, p.NetworkID(ctx)).ExecWithCount()
```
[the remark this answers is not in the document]
grantzvolsky: Does that allow us to count the affected rows?

persistence/sql/persister_consent.go
```diff
@@ -264,103 +319,117 @@ func (p *Persister) ConfirmLoginSession(ctx context.Context, id string, authenti
...
+	session.NID = p.NetworkID(ctx)
 	return sqlcon.HandleError(p.Connection(ctx).Create(session))
 }
 
 func (p *Persister) DeleteLoginSession(ctx context.Context, id string) error {
-	return sqlcon.HandleError(
-		p.Connection(ctx).Destroy(
-			&consent.LoginSession{ID: id},
-		))
+	count, err := p.Connection(ctx).RawQuery("DELETE FROM hydra_oauth2_authentication_session WHERE id=? AND nid = ?", id, p.NetworkID(ctx)).ExecWithCount()
```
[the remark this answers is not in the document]
grantzvolsky: Is there a way to use QueryWith and determine whether any rows were affected?

persistence/sql/persister_consent.go
```diff
@@ -85,7 +83,7 @@ func (p *Persister) revokeConsentSession(whereStmt string, whereArgs ...interfac
 }
 
 func (p *Persister) RevokeSubjectLoginSession(ctx context.Context, subject string) error {
-	if err := p.Connection(ctx).RawQuery("DELETE FROM hydra_oauth2_authentication_session WHERE subject = ?", subject).Exec(); errors.Is(err, sql.ErrNoRows) {
+	if err := p.Connection(ctx).RawQuery("DELETE FROM hydra_oauth2_authentication_session WHERE subject = ? AND nid = ?", subject, p.NetworkID(ctx)).Exec(); errors.Is(err, sql.ErrNoRows) {
```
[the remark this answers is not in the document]
grantzvolsky: Does the original (as well as updated) code do what it purports to do? I don't think that the code can result in `ErrNoRows`. The only way to report ErrNotFound that I'm aware of is to use ExecWithCount, which is incompatible with QueryWithNetwork.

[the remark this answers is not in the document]
grantzvolsky: I see that this was discussed in #2397. I'll remove the not found error handling because I couldn't reproduce the error, even after testing all supported databases.

persistence/sql/persister_consent.go
grantzvolsky: Is there a way to count the affected rows with Delete?

persistence/sql/persister_consent.go
```diff
@@ -392,41 +465,46 @@ func (p *Persister) listUserAuthenticatedClients(ctx context.Context, subject, s
 }
 
 func (p *Persister) CreateLogoutRequest(ctx context.Context, request *consent.LogoutRequest) error {
+	request.NID = p.NetworkID(ctx)
 	return errorsx.WithStack(p.Connection(ctx).Create(request))
 }
 
 func (p *Persister) AcceptLogoutRequest(ctx context.Context, challenge string) (*consent.LogoutRequest, error) {
-	if err := p.Connection(ctx).RawQuery("UPDATE hydra_oauth2_logout_request SET accepted=true, rejected=false WHERE challenge=?", challenge).Exec(); err != nil {
+	if err := p.Connection(ctx).RawQuery("UPDATE hydra_oauth2_logout_request SET accepted=true, rejected=false WHERE challenge=? AND nid = ?", challenge, p.NetworkID(ctx)).Exec(); err != nil {
```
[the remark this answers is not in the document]
grantzvolsky: The above comment was most likely made by mistake

persistence/sql/persister_consent.go
grantzvolsky: The above comment was most likely made by mistake (RawQuery overrides query pop's query builder)

persistence/sql/persister_consent.go
grantzvolsky: The above comment was most likely made by mistake (RawQuery overrides pop's query builder)

persistence/sql/persister_jwk.go
grantzvolsky: There are two reasons I introduced this change - to facilitate a test, and because in my experience, it is the more common practice. I thought the current implementation was a mistake, but I appreciate being informed that the current state is intentional. I'll revert my change to preserve the convention.

persistence/sql/persister_jwk.go
grantzvolsky: See my comment below (https://github.com/ory/hydra/pull/3013#discussion_r838209491)

persistence/sql/persister_jwk.go
grantzvolsky: See my comment below (https://github.com/ory/hydra/pull/3013#discussion_r838209491)

persistence/sql/persister_oauth2.go
grantzvolsky: I'm not sure if it's possible to combine RawQuery and QueryWith. Would you elaborate on this suggestion?

persistence/sql/persister_oauth2.go
grantzvolsky: I'm not sure if it's possible to combine RawQuery and QueryWith. Would you elaborate on this suggestion?

flow/flow.go
```diff
@@ -0,0 +1,475 @@
...
+	}
+}
+
+func (_ Flow) TableName() string {
+	return "hydra_oauth2_flow"
+}
+
+// FindByConsentChallengeID retrieves a flow given its consent challenge ID.
+func (f *Flow) FindByConsentChallengeID(c *pop.Connection, id string, nid uuid.UUID) error {
+	return c.Where("consent_challenge_id = ? AND nid = ?", id, nid).First(f)
```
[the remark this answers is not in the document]
grantzvolsky: I've removed this function because it doesn't seem to add any value and we'd have to pass in the persister to use QueryWithNetwork.

Furthermore I'd like to remove `challenge` from the signature of HandleConsentRequest https://github.com/ory/hydra/blob/00100a1bcb60d1836a2c3d6c6a4212e3161b1bda/consent/manager.go#L43, because we require the other argument, `r *HandledConsentRequest` to have the correct challenge id. Passing it in a separate argument is then redundant. Do you agree with this interface change?

client/validator.go
grantzvolsky: Good idea

driver/registry_sql.go
```diff
@@ -101,10 +114,34 @@ func (m *RegistrySQL) Init(ctx context.Context) error {
...
+			net, err := p.DetermineNetwork(ctx)
+			if err != nil {
+				m.Logger().WithError(err).Warnf("Unable to determine network, retrying.")
 				return err
 			}
+
+			m.persister = p.WithNetworkID(net.ID)
 		}
+
+		if m.C.HsmEnabled() {
```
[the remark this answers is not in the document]
grantzvolsky: Done. I'm just hijacking this comment to note that I have yet to review all the references to `m.C` in registry_base

driver/registry_sql.go
```diff
@@ -80,18 +99,12 @@ func (m *RegistrySQL) Init(ctx context.Context) error {
 		if err := resilience.Retry(m.l, 5*time.Second, 5*time.Minute, c.Open); err != nil {
 			return errorsx.WithStack(err)
 		}
-		m.persister, err = sql.NewPersister(ctx, c, m, m.C, m.l)
+
+		p, err := sql.NewPersister(ctx, c, m, m.C, m.l)
```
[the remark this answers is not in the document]
grantzvolsky: Note to self: TODO

persistence/sql/persister_consent.go
grantzvolsky: By the way, it's possible to reduce the number of database accesses from three to two by moving the WHERE check from the first SELECT into the UPDATE and querying the final result using verifier instead of challenge. Should I do it?

Note to self: the RawQuery UPDATE is necessary because we're updating by the verifier which isn't the primary key.

grantzvolsky: On a second thought, I can see how this could be done without RawQuery in 3 DB roundtrips. So the options are
1. Refactor the function with 3 db roundtrips and no RawQuery
2. Refactor the function with 2 db roundtrips keeping the RawQuery

flow/flow.go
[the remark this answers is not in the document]
aeneasr: SGTM

flow/flow.go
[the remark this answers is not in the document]
aeneasr: Ok makes sense, can you create an issue for this TODO item and link it in the comment? In the issue you can provide a bit of context as to what could be improved.

persistence/sql/persister_consent.go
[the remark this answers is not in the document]
aeneasr: I'm not sure, if not the current was is ok

persistence/sql/persister_oauth2.go
aeneasr: You're right, I don't think that's possible

docs/docs/guides/migrating-to-hydra2.mdx
```diff
@@ -0,0 +1,36 @@
+---
```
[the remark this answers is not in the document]
aeneasr: @grantzvolsky the document is removed but there is no PR for it in https://github.com/ory/docs/pulls - could you please create a PR there with this document?

driver/registry_sql.go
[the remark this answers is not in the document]
aeneasr: I have added this as a follow up item to: https://github.com/ory/hydra/issues/3054

aeneasr:
Note to self - all review items of https://github.com/ory/hydra/pull/3013#pullrequestreview-915679585 have been addressed. The only remaining issue from that review is:

- [ ] https://github.com/ory/hydra/pull/3013#discussion_r841056028

persistence/sql/migratest/fixtures/networks/00000000-0000-0000-0000-000000000000.json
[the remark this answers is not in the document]
aeneasr: Ok, I have looked into the test and can confirm that. I've removed the fixture and the line calling it, because I think it adds little value to the test since we just override the nid with uuid.Nil

persistence/sql/src/20220210000001_nid/20220210000001000000_nid.sqlite.up.sql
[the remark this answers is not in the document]
aeneasr: Resolved as my assumption was not correct here, UNIQUE INDEX is equivalent to INDEX in terms of lookup performance.

flow/flow.go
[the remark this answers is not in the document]
aeneasr: Actually, I don't think we need a follow up here for now, marking as resolved.

persistence/sql/persister_consent.go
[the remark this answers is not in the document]
aeneasr: Apparently it doesn't

persistence/sql/persister_consent.go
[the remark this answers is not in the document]
aeneasr: There isn't :/

persistence/sql/persister_consent.go
aeneasr: 2db roundtrips it is :)

persistence/sql/persister_grant_jwk.go
aeneasr: @grantzvolsky why was this not caught by the tests that cover the networks?

persistence/sql/persister_jwk.go
aeneasr: Thank you for the thoughtful response!

persistence/sql/persister_oauth2.go
aeneasr: Should this not have been caught by tests?

aeneasr:
@grantzvolsky I think all comments in https://github.com/ory/hydra/pull/3013#pullrequestreview-920032457 have been addressed! From a functional perspective, I think this is extremely solid.

Regarding tests, I think we're lacking nid tests for some of the functions which is why we didn't catch that they were missing in some cases. It's important though to have them so that no regressions can occurr.

Are you planning on addressing them in this PR? I don't think it will be possible to get 100% coverage - especially for JOIN queries it will be difficult to set up the test for this. But I think we can reach ~90% of functions that use nid.

aeneasr:
By the way, tests fail on my machine in the janitor due to some db.Scan issue.

docs/docs/guides/migrating-to-hydra2.mdx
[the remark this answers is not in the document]
aeneasr: It's here https://github.com/ory/docs/pull/708...

aeneasr:
@grantzvolsky I have pushed all changes from this PR to branch v2.x.

Can you please create any new tests and commtis you write as a new PR with base of v2.x? This will help us consolidate the branches and make it easier to direct users to make PRs against v2.x branch, as this branch is on a fork.

grantzvolsky:
@aeneasr will do

PR #3029, opened 2022-03-05
fix: Remove unnecessary transaction

mitar:
While resolving an issue in my own implementation of storage for Fosite which has been modeled after Hydra's, I realized:

* There is unnecessary transaction in this function. The first query (deleting expired assertions) is completely unrelated from the other one (inserting a new assertion). In fact, deletion could even be done as a background job and not inside this function (in fact, why it is not a background job, or in a goroutine).
* Transaction isolation level used here is a default one for the connection (i.e., read committed) while `BeginTX` uses `LevelRepeatableRead`. Is this on purpose or is this a bug? In my implementation I always used the same way to create a transaction so I always had repeatable read transactions, which combined with the unnecessary transaction in this function lead to commit conflicts (because multiple parallel transactions were attempting to delete same expired assertions).

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/persistence/sql/persister_oauth2.go
+++ b/persistence/sql/persister_oauth2.go
@@ -175,26 +175,26 @@ func (p *Persister) ClientAssertionJWTValid(ctx context.Context, jti string) err
 }
 
 func (p *Persister) SetClientAssertionJWT(ctx context.Context, jti string, exp time.Time) error {
-	return p.transaction(ctx, func(ctx context.Context, c *pop.Connection) error {
-		// delete expired
-		now := "now()"
-		if c.Dialect.Name() == "sqlite3" {
-			now = "CURRENT_TIMESTAMP"
-		}
-		/* #nosec G201 table is static */
-		if err := c.RawQuery(fmt.Sprintf("DELETE FROM %s WHERE expires_at < %s", oauth2.BlacklistedJTI{}.TableName(), now)).Exec(); err != nil {
-			return sqlcon.HandleError(err)
-		}
+	// delete expired
+	c := p.Connection(ctx)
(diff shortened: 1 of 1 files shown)


aeneasr:
You're right,  that should indeed be a background task. But we can merge this as is already, thank you!

mitar:
Note: are you sure it is OK that some transactions are not using `LevelRepeatableRead`? So you have to code paths here to create transactions and one has `LevelRepeatableRead` level and the other does not. Is this on purpose?

mitar:
Sorry to nitpick, but "serializable" is the highest isolation level. :-)

Also, we had issues (and this is how I discovered this code) at a pretty low scale: it was pretty common to get "transaction aborted" with repeatable read before this code change, because the delete query touches the whole table, so two parallel queries had always a conflict.

Anyway, if you know what you are doing and why is `LevelRepeatableRead` even set there in the first place, then it is OK. :-)

aeneasr:
I see, to be honest I haven't looked into the code in a while and it always requires cognitive load to do that. If this is an issue that is bothering you in production and is causing problems in delivering the service, please feel free to open an issue with a problem statement, reproducible case, and ideas on how to resolve it. If that's not the case I'll refrain from trying to think myself through this :)

mitar:
No, issues on my side has been resolved. So this is all fine now. I just wanted to put this isolation-level discrepancy on your radar so that if in the future you might learn about some issue with isolation, you might remember that parts of queries are using one isolation level and parts another.

mitar:
With some luck, I might also notice it and pitch in. :-)

Have a good night!

PR #3059, opened 2022-04-06
feat: enable simultaneous auth flows by creating client related csrf co…

### LINKED ISSUE #3019: Not able to perform simultaneous auth flows with different clients
### Preflight checklist

- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

After reading [this comment](https://github.com/ory/hydra/issues/923#issuecomment-450828062), I understand that Hydra didn't support this 4 years ago (Wow, that's a long time!). However, I feel that when multiple distinct clients are performing the auth flow, hydra should be able to support each of them individually.

### Reproducing the bug

1. Open 2 different apps, with 2 different clients
2. Initiate the login flow in different tabs of the same browser
3. Error: The CSRF value from the token does not match the CSRF value from the data store

### Version

v1.10.7

Linux

Kubernetes with Helm

### Additional Context

Here's the use case: We have multiple apps loaded in an iFrame and they all perform the authentication flow simultaneously. This is according to the SMART on FHIR specification for healthcare applications. We need to load all the iFrames simultaneously to optimize the speed of loading. Also, there are instances when multiple apps with different client ids might be opened on different tabs. We don't have control over most of these applications since they are developed by 3rd party developers.

aarmam:
This pull request enables simultaneous auth flows by creating client related csrf cookie names. 

oauth2_authentication_csrf -> oauth2_authentication_csrf_%murmur3(client_id)%
oauth2_authentication_csrf_insecure -> oauth2_authentication_csrf_%murmur3(client_id)%_insecure
oauth2_consent_csrf -> oauth2_consent_csrf_%murmur3(client_id)%
oauth2_consent_csrf_insecure -> oauth2_consent_csrf_%murmur3(client_id)%_insecure

Additionally max age is set for cookie using ttl.login_consent_request configuration property.

## Related issue(s)
Fixes #3019

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/consent/strategy_default.go
+++ b/consent/strategy_default.go
@@ -22,11 +22,14 @@ package consent
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"net/url"
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/twmb/murmur3"
 
 	"github.com/ory/hydra/driver/config"
 
@@ -247,6 +250,7 @@ func (s *DefaultStrategy) forwardAuthenticationRequest(ctx context.Context, w ht
 	}
 
 	// Set the session
+	cl := sanitizeClientFromRequest(ar)
 	if err := s.r.ConsentManager().CreateLoginRequest(
 		r.Context(),
 		&LoginRequest{
@@ -257,7 +261,7 @@ func (s *DefaultStrategy) forwardAuthenticationRequest(ctx context.Context, w ht
 			RequestedScope:    []string(ar.GetRequestedScopes()),
 			RequestedAudience: []string(ar.GetRequestedAudience()),
 			Subject:           subject,
-			Client:            sanitizeClientFromRequest(ar),
+			Client:            cl,
 			RequestURL:        iu.String(),
 			AuthenticatedAt:   sqlxx.NullTime(authenticatedAt),
 			RequestedAt:       time.Now().Truncate(time.Second).UTC(),
@@ -274,7 +278,8 @@ func (s *DefaultStrategy) forwardAuthenticationRequest(ctx context.Context, w ht
 		return errorsx.WithStack(err)
 	}
 
-	if err := createCsrfSession(w, r, s.r.Config(), s.r.CookieStore(ctx), s.r.Config().CookieNameLoginCSRF(ctx), csrf); err != nil {
+	clientSpecificCookieNameLoginCSRF := fmt.Sprintf("%s_%d", s.r.Config().CookieNameLoginCSRF(ctx), murmur3.Sum32([]byte(cl.LegacyClientID)))
+	if err := createCsrfSession(w, r, s.r.Config(), s.r.CookieStore(ctx), clientSpecificCookieNameLoginCSRF, csrf, s.c.ConsentRequestMaxAge(ctx)); err != nil {
 		return errorsx.WithStack(err)
 	}
 
@@ -335,7 +340,8 @@ func (s *DefaultStrategy) verifyAuthentication(w http.ResponseWriter, r *http.Re
 		return nil, errorsx.WithStack(fosite.ErrRequestUnauthorized.WithHint("The login request has expired. Please try again."))
 	}
 
-	if err := validateCsrfSession(r, s.r.Config(), s.r.CookieStore(ctx), s.r.Config().CookieNameLoginCSRF(ctx), session.LoginRequest.CSRF); err != nil {
+	clientSpecificCookieNameLoginCSRF := fmt.Sprintf("%s_%d", s.r.Config().CookieNameLoginCSRF(ctx), murmur3.Sum32([]byte(session.LoginRequest.Client.LegacyClientID)))
+	if err := validateCsrfSession(r, s.r.Config(), s.r.CookieStore(ctx), clientSpecificCookieNameLoginCSRF, session.LoginRequest.CSRF); err != nil {
 		return nil, err
 	}
 
@@ -523,6 +529,7 @@ func (s *DefaultStrategy) forwardConsentRequest(ctx context.Context, w http.Resp
 	challenge := strings.Replace(uuid.New(), "-", "", -1)
 	csrf := strings.Replace(uuid.New(), "-", "", -1)
 
+	cl := sanitizeClientFromRequest(ar)
 	if err := s.r.ConsentManager().CreateConsentRequest(
 		r.Context(),
 		&OAuth2ConsentRequest{
@@ -535,7 +542,7 @@ func (s *DefaultStrategy) forwardConsentRequest(ctx context.Context, w http.Resp
 			RequestedScope:         []string(ar.GetRequestedScopes()),
 			RequestedAudience:      []string(ar.GetRequestedAudience()),
 			Subject:                as.Subject,
-			Client:                 sanitizeClientFromRequest(ar),
+			Client:                 cl,
 			RequestURL:             as.LoginRequest.RequestURL,
 			AuthenticatedAt:        as.AuthenticatedAt,
 			RequestedAt:            as.RequestedAt,
@@ -549,7 +556,8 @@ func (s *DefaultStrategy) forwardConsentRequest(ctx context.Context, w http.Resp
 		return errorsx.WithStack(err)
 	}
 
-	if err := createCsrfSession(w, r, s.r.Config(), s.r.CookieStore(ctx), s.r.Config().CookieNameConsentCSRF(ctx), csrf); err != nil {
+	clientSpecificCookieNameConsentCSRF := fmt.Sprintf("%s_%d", s.r.Config().CookieNameConsentCSRF(ctx), murmur3.Sum32([]byte(cl.LegacyClientID)))
+	if err := createCsrfSession(w, r, s.r.Config(), s.r.CookieStore(ctx), clientSpecificCookieNameConsentCSRF, csrf, s.c.ConsentRequestMaxAge(ctx)); err != nil {
 		return errorsx.WithStack(err)
 	}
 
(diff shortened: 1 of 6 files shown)


sidharthramesh:
@aarmam any reason why hashing with murmur3 is necessary in this case? Does appending the client name directly cause any security issues?

aarmam:
> @aarmam any reason why hashing with murmur3 is necessary in this case? Does appending the client name directly cause any security issues?

[Set-Cookie spec.]( https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie)

> A <cookie-name> can contain any US-ASCII characters except for: the control character, space, or a tab. It also must not contain a separator characters like the following: ( ) < > @ , ; : \ " / [ ] ? = { }.

[Admin API, POST /clients](https://www.ory.sh/docs/hydra/reference/api#operation/createOAuth2Client) endpoint allows such characters for Client ID.

aeneasr (commented):
Thank you for these changes! On a high level, I think they look good. One question I have though is whether this is the right approach? Are there many use cases where multiple clients perform simoultaneously OAuth2 flows that a user is executing? What for sure is a problem is that a user might initiate two or more OAuth2 flows by accident, and then ends up in a situation where they can't proceed due to CSRF failures.

The underlying problem is that in Ory Hydra we regenerate the CSRF token regardless of whether the principal has changed or not, as we usually do not know whether the principal has changed in the first place.

In an optimal system, CSRF tokens are regenerated when the authentication session changes (logout, different user logs in) and then stay consistent for the remainder of the cookie lifetime. This helps prevent issues where a user loads two pages with forms and ends up in a race condition where the CSRF cookie from form B overrides the one from form A, making it impossible to submit form A without reloading it (hence setting a new cookie).

In Hydra the problem is that we don't really know what the principal is until we accepted the login challenge, however, by then, it is too late for us to set the CSRF token.

So I'm wondering whether there is a "better" solution to this problem? We could also, for example, set the CSRF cookie when the flow initializes and only reset it once the flow is completed successfully? That could still cause issues when performing multiple flows at the same time, but it would alleviate the use case I mentioned above:

> What for sure is a problem is that a user might initiate two or more OAuth2 flows by accident, and then ends up in a situation where they can't proceed due to CSRF failures.

Another option I thought of is to set the CSRF cookie based on some other parameter, for example the `state` parameter as that still represents a unique identifier that could be used to track the flow from start to finish. But then again, the state parameter can be user-chosen and is not guaranteed to be collision free.

Looking forward to hear some ideas.

sidharthramesh:
In my use case there are multiple iframes - each a separate OAuth2 client executing the login flow simultaneously. So it’s different client executing the oauth flow simultaneously, always!

I understand your use case of users accidentally initiating multiple auth flows (maybe click the back button before a full redirect?) - but they are usually rare and with the same client, so this PR is unlikely to solve those use cases.

Regarding your solution of waiting for the flow to end before resetting the CSRF token might work for both situations if we make the season specific for user + client combination.

Using the state sounds like a good idea. It’s 8 digits long and the probability of collisions are sort of low? In case of collision maybe a fallback on some other method?

aeneasr:
I guess the problem with state is that it is purely user-set and we have no control over the randomness. Using it as the basis for defensive measures introduced server-side *may* open up specific attack bectors that, depending on use case, could be severe.

aarmam:
> Thank you for these changes! On a high level, I think they look good. One question I have though is whether this is the right approach? Are there many use cases where multiple clients perform simoultaneously OAuth2 flows that a user is executing? 

@aeneasr 
Our use case actually is silent refresh (prompt=none)

mig5:
> Are there many use cases where multiple clients perform simoultaneously OAuth2 flows that a user is executing

Each of my RPs uses a different oAuth2.0 client, and it only takes opening 2 such RPs in different tabs, which users seem to do all the time. These apps are *automatically* initiating the auth flow just on trying to load the front page, e.g the user is not clicking on a 'Login' button or anything like that which would 'skew' the timing in order to see the second tab 'auto log in' thanks to auth having completed via the first.

Example: https://github.com/vouch/vouch-proxy (auth_request in Nginx to force authentication of a site via the OP)

aeneasr:
I see, thank you for the use case examples. That makes sense. If possible, I would like to use this PR to come up with a solution that works even better for simoultaneous OAuth2 flows from either distinct OAuth2 Clients, or even the same OAuth2 clients. I highly appreciate your input here, if you have ideas!

sidharthramesh:
This PR is extremely useful for a lot of our use cases in its current situation. We were using a very old fork of Hydra with our own local changes, but just tried this PR and everything works nicely. Would be really great if this feature gets merged!

aeneasr:
You're right, also changing the cookie mechanism *should* not lead to many issues later on if we decide to change the CSRF approach, so this is certainly already an improvement worthy to be merged!

consent/strategy_oauth_test.go
```diff
@@ -271,6 +273,76 @@ func TestStrategyLoginConsentNext(t *testing.T) {
...
+
+		_, oauthRes := makeOAuth2Request(t, reg, hc, c, url.Values{"redirect_uri": {c.RedirectURIs[0]}, "scope": {"openid"}})
+		assert.EqualValues(t, http.StatusFound, oauthRes.StatusCode)
+		loginChallengeRedirect, err := oauthRes.Location()
+		require.NoError(t, err)
+		defer oauthRes.Body.Close()
+
+		setCookieHeader := oauthRes.Header.Get("set-cookie")
+		assert.NotNil(t, setCookieHeader)
+		assert.Regexp(t, fmt.Sprintf("oauth2_authentication_csrf_%d_insecure=.*", murmur3.Sum32([]byte(c.OutfacingID))), setCookieHeader)
```
aeneasr, line 330: Does this apply only to the `_insecure` cookie? If not, can you please add the secure cookie check also? :)

consent/strategy_oauth_test.go
```diff
@@ -271,6 +273,76 @@ func TestStrategyLoginConsentNext(t *testing.T) {
...
+		defer loginChallengeRes.Body.Close()
+		loginVerifierRedirect, err := loginChallengeRes.Location()
+
+		loginVerifierRes, err := hc.Get(loginVerifierRedirect.String())
+		require.NoError(t, err)
+		defer loginVerifierRes.Body.Close()
+
+		setCookieHeader = loginVerifierRes.Header.Values("set-cookie")[1]
+		assert.NotNil(t, setCookieHeader)
+		assert.Regexp(t, fmt.Sprintf("oauth2_consent_csrf_%d_insecure=.*", murmur3.Sum32([]byte(c.OutfacingID))), setCookieHeader)
```
aeneasr, line 343: Does this apply only to the `_insecure` cookie? If not, can you please add the secure cookie check also? :)

aeneasr (commented):
Just two minor things in tests

sidharthramesh:
Hey @aarmam let me know if you need any help! Saw that you marked this as a draft.

aarmam:
> Hey @aarmam let me know if you need any help! Saw that you marked this as a draft.

Hey @sidharthramesh I will update all my pull requests next week and mark them ready for review again! Sorry to keep you waiting - will take this PR first! :)

consent/strategy_oauth_test.go
aarmam: I think tests in [TestValidateCsrfSession](https://github.com/ory/hydra/pull/3059/files#diff-adc7dbe135263d29b9a563ef62fbcfadfac40cf27b6fbeb75452084cc5c8b9eb) cover setting secure cookie? This PR did not change ([1](https://github.com/ory/hydra/pull/3059/files#diff-265e72885dd37a9d8c6f81e2c928b2bfad035cecd98051ef2f04ea88a56ed540R287), [2](https://github.com/ory/hydra/pull/3059/files#diff-265e72885dd37a9d8c6f81e2c928b2bfad035cecd98051ef2f04ea88a56ed540R364), [3](https://github.com/ory/hydra/pull/3059/files#diff-265e72885dd37a9d8c6f81e2c928b2bfad035cecd98051ef2f04ea88a56ed540R579), [4](https://github.com/ory/hydra/pull/3059/files#diff-265e72885dd37a9d8c6f81e2c928b2bfad035cecd98051ef2f04ea88a56ed540R615)) how cookies are set and tests in consent/strategy_oauth_test.go only validate, that cookie with expected prefix are set.

consent/strategy_oauth_test.go
aarmam: I think tests in [TestValidateCsrfSession](https://github.com/ory/hydra/pull/3059/files#diff-adc7dbe135263d29b9a563ef62fbcfadfac40cf27b6fbeb75452084cc5c8b9eb) cover setting secure cookie? This PR did not change ([1](https://github.com/ory/hydra/pull/3059/files#diff-265e72885dd37a9d8c6f81e2c928b2bfad035cecd98051ef2f04ea88a56ed540R287), [2](https://github.com/ory/hydra/pull/3059/files#diff-265e72885dd37a9d8c6f81e2c928b2bfad035cecd98051ef2f04ea88a56ed540R364), [3](https://github.com/ory/hydra/pull/3059/files#diff-265e72885dd37a9d8c6f81e2c928b2bfad035cecd98051ef2f04ea88a56ed540R579), [4](https://github.com/ory/hydra/pull/3059/files#diff-265e72885dd37a9d8c6f81e2c928b2bfad035cecd98051ef2f04ea88a56ed540R615)) how cookies are set and tests in consent/strategy_oauth_test.go only validate, that cookie with expected prefix are set.

sidharthramesh:
Thank you @aarmam!
I looked through your changes and also built the image and tried it out in our staging environment. I can confirm that we can have multiple oauth clients simultaneously logging on using iframes successfully!

@aeneasr looking forward to your review.

aeneasr:
Hey, this looks pretty good I think! We're currently freezing any pushes to master as we're working on v2.x and have a tremendous amount of merge conflicts which we want to avoid. We're planning to not introduce new features to hydra v1.x branch for the time being, unless it's critical

aeneasr:
You can help by making this PR against branch v2.x but I understand if that is too much work right now

sidharthramesh:
wow 😄

consent/strategy_default.go
```diff
@@ -584,7 +592,8 @@ func (s *DefaultStrategy) verifyConsent(ctx context.Context, w http.ResponseWrit
 		return nil, errorsx.WithStack(fosite.ErrServerError.WithHint("The authenticatedAt value was not set."))
 	}
 
-	if err := validateCsrfSession(r, s.r.Config(), s.r.CookieStore(ctx), s.r.Config().CookieNameConsentCSRF(ctx), session.ConsentRequest.CSRF); err != nil {
+	clientSpecificCookieNameConsentCSRF := fmt.Sprintf("%s_%d", s.r.Config().CookieNameConsentCSRF(ctx), murmur3.Sum32([]byte(session.ConsentRequest.Client.LegacyClientID)))
```
aeneasr, line 595:
```suggestion
	clientSpecificCookieNameConsentCSRF := fmt.Sprintf("%s_%d", s.r.Config().CookieNameConsentCSRF(ctx), murmur3.Sum32([]byte(session.ConsentRequest.Client.ID)))
```

:)

aeneasr (changes requested):

aeneasr (approved):

PR #3066, opened 2022-04-13
feat: add hsm key set prefix to support multiple hydra instances on t…

aarmam:
This pull request adds configuration option `hsm.key_set_prefix` to support multiple Ory Hydra instances to store keys on the same HSM partition. For example if `hsm.key_set_prefix=app1.` then key set `hydra.openid.id-token` would be generated/requested/deleted on HSM with `CKA_LABEL=app1.hydra.openid.id-token`

This will not affect Hydra API in any way. `GET /keys/hydra.openid.id-token` will return key set from HSM with label `app1.hydra.openid.id-token`

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/driver/registry_sql.go
+++ b/driver/registry_sql.go
@@ -86,7 +86,7 @@ func (m *RegistrySQL) Init(ctx context.Context) error {
 		}
 
 		if m.C.HsmEnabled() {
-			hardwareKeyManager := hsm.NewKeyManager(m.HsmContext(), m.C.HsmKeyPrefix())
+			hardwareKeyManager := hsm.NewKeyManager(m.HsmContext(), m.C)
 			m.defaultKeyManager = jwk.NewManagerStrategy(hardwareKeyManager, m.persister)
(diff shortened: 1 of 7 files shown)


driver/registry_sql.go
```diff
@@ -86,7 +86,7 @@ func (m *RegistrySQL) Init(ctx context.Context) error {
 		}
 
 		if m.C.HsmEnabled() {
-			hardwareKeyManager := hsm.NewKeyManager(m.HsmContext())
+			hardwareKeyManager := hsm.NewKeyManager(m.HsmContext(), m.C.HsmKeyPrefix())
```
aeneasr, line 89: Could you instead please pass the function itself, or alternatively `m.C`? That way, hot reloading this config value would work.

aeneasr (changes requested):

aeneasr (approved):

PR #3092, opened 2022-04-26
Added client metadata flag, fixes #3091

### LINKED ISSUE #3091: CLI `clients create` command is missing flag to set Metadata
### Preflight checklist

- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.

### Describe the bug

When creating clients from the CLI, it's not currently possible to set the `OAuth2Client.Metadata` property. 

### Reproducing the bug

1. Run `hydra clients create --help` and check the flags.

### Relevant log output

_No response_

### Relevant configuration

_No response_

### Version

Latest

_No response_

Binary

### Additional Context

I earlier submitted https://github.com/ory/x/pull/497 which will allow setting `OAuth2Client.Metadata` to a `map[string]string` value, which works well for my use case. However looking at the API reference it seems that metadata is a `JSONRawMessage` or `interface{}`, so should the CLI permit setting `OAuth2Client.Metadata` to be arbitrary JSON as a string input, or does a `map[string]string` suffice for most CLI use cases?

If `map[string]string` is sufficient I'll submit a PR to add that. Thanks.

rikkuness:
Adds a `--metadata` flag to both `hydra clients create` and `hydra clients update`. Also ensures that valid JSON is being passed.

## Related issue(s)

Fixes #3091

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.


aeneasr (changes requested):
Thank you! Can you please add this field to the tests at: https://github.com/ory/hydra/blob/master/cmd/root_test.go? :)

rikkuness:
Sorry about that, I missed that file the first time around 😄 

Tests added, I tried to add a test to test for invalid JSON with the `expectErr` in the tests but as the CLI is using `cmdx.Must()` which calls an `os.Exit()`, the `rootCmd.Execute()` never actually returns an error so you can't test for it.

cmd/cli/handler_client.go
```diff
@@ -101,6 +101,13 @@ func (h *ClientHandler) CreateClient(cmd *cobra.Command, args []string) {
 	ek, encryptSecret, err := newEncryptionKey(cmd, nil)
 	cmdx.Must(err, "Failed to load encryption key: %s", err)
 
+	var metadata models.JSONRawMessage
+	metadataStr := flagx.MustGetString(cmd, "metadata")
+	if metadataStr != "" {
+		err = json.Unmarshal([]byte(flagx.MustGetString(cmd, "metadata")), &metadata)
+		cmdx.Must(err, "Failed to parse metadata: %s", err)
```
aeneasr, line 108: Can you please use https://github.com/tidwall/gjson/blob/master/gjson.go#L2549 instead and improve the error message a bit? :)

aeneasr (changes requested):

cmd/cli/handler_client.go
rikkuness: Sure :) Although the JSON string, will it still need to also be parsed out to type `JSONRawMessage` to pass into the `OAuth2Client` struct?

aeneasr: Yes, but `JSONRawMessage` is just `[]byte()`, so:

```
var JSONRawMessage = []byte(`{"foo":"bar"}`)
```

rikkuness: So if I set `Metadata` as above as just a raw string from the CLI input, then it gets added to the client as: 
```json
"metadata": "{\"foo\":\"bar\"}",
```
Which is not what I would expect to see. However if I do something like:
```go
var metadata interface{}
json.Unmarshal([]byte(flagx.MustGetString(cmd, "metadata")), &metadata)
```

Then in the resulting client output I get:
```json
"metadata": {
	"foo": "bar"
},
```

Which is much more like what I was expecting

aeneasr:
```
var metadata json.RawMessage = []byte(flagx.MustGetString(cmd, "metadata"))
```

is equivalent and to much faster than

```
var metadata interface{}
json.Unmarshal([]byte(flagx.MustGetString(cmd, "metadata")), &metadata)
```

:)

rikkuness: Ohhhh, my bad, yeah I was missing the fact `JSONRawMessage` implemented an Unmarshaler! 

I'll update that now, thanks!

aeneasr: awesome :)

rikkuness: Apologies, I'd made some local changes and then completely forgot about this PR 😬 I've just pushed now, does that look more like what you'd expect? It seemed long winded but `gjson.Valid` returned a bool so I wasn't sure the preferred way to error out, at the moment it's just printing there was an error and carrying on anyway

rikkuness:
Hey, is this superseded by the recent commits improve `create client`? Looks like those commits closed the original issue

aeneasr:
Yes!

rikkuness:
Awesome, thanks!

Issue #2462, opened 2021-04-12
Public endpoints different than issuer


ppawliczek:
**Describe the bug**

In the configuration there are parameters urls->self->public (URLS_SELF_PUBLIC) and urls->self->issuer (URLS_SELF_ISSUER). The former one is described as "This is the base location of the public endpoints of your ORY Hydra installation.". However, when I tried to set URLS_SELF_ISSUER to different value, the authorization and token endpoints reported in the metadata file (.../.well-known/openid-configuration) changes together with the "issuer" field (the parameter URLS_SELF_PUBLIC is ignored).

**Expected behavior**

I would expect that location of the endpoints is determined by URLS_SELF_PUBLIC, while the value of the issuer is determined by URLS_SELF_ISSUER.

aeneasr:
Ah this looks like a regression from a recent patch we merged. You are correct, issuer should only be used for the issuer values. Would you be open to a PR?

ppawliczek:
#2441 should fix it

PR #3094, opened 2022-04-27
fix: use public url instead of issuer url

CiiDyR:
FIX: This patch changes URL generation to use `urls.self.public` instead of `urls.self.issuer`.

## Related issue(s)
This PR tries to fix URL generation. Currently, some URLs (jwks_uri, revocation_endpoint and end_session_endpoint) are generated relative to the issuer URL, as opposed to the public URL.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/driver/config/provider.go
+++ b/driver/config/provider.go
@@ -388,7 +388,7 @@ func (p *Provider) OAuth2AuthURL() *url.URL {
 }
 
 func (p *Provider) JWKSURL() *url.URL {
-	return p.p.RequestURIF(KeyJWKSURL, urlx.AppendPaths(p.IssuerURL(), "/.well-known/jwks.json"))
+	return p.p.RequestURIF(KeyJWKSURL, urlx.AppendPaths(p.PublicURL(), "/.well-known/jwks.json"))
 }
 
 func (p *Provider) TokenRefreshHookURL() *url.URL {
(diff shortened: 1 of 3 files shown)


aeneasr (changes requested):
Thank you for this PR! Given the breaking change of this feature but the small effect it has, I don't think we will be able to merge this upstream. The Issuer URL is the public-facing URL intended for all OpenID Connect URLs (e.g. JWKS), and it falls back to the public URL if the issuer URL is not set. This PR would remove that and always use the public URL, which might not overlap with the issuer URL (for whatever reason) and cause issues.

CiiDyR:
No worries! I just ran into it, as our issuer url doesn't actually point to Hydra, but rather our implementation as we intend to run hydra under api.[domain], and it just "looked" better ;-) And a related issue had a comment of yours that made me think right now isn't actually intended behavior (https://github.com/ory/hydra/issues/2462#issuecomment-819343065). No harm done for us if isn't merged :-)

aeneasr:
Absolutely! Some people have also been asking to use a dedicated URL for revokation and some of the other endpoints, through a config parameter, which we could also add similar to the userinfo one!

For the config, this is located here: https://github.com/ory/hydra/blob/master/spec/config.json

aeneasr:
To clean up our back log a bit, I will close this PR here. Please open a new PR if you have some time to work on the suggested changes - thank you so much! :)

PR #3110, opened 2022-05-10
fix: retry transient crdb transaction failures

grantzvolsky:
## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.


persistence/sql/transaction/transaction.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/gobuffalo/pop/v6"
+	"github.com/jmoiron/sqlx"
+	"github.com/ory/x/errorsx"
+)
+
+type transactionContextType string
+
+const TransactionContextKey transactionContextType = "transactionConnection"
+
+func Transaction(ctx context.Context, conn *pop.Connection, f func(ctx context.Context, c *pop.Connection) error) error {
```
aeneasr, line 16: Could we use `popx` instead? It seems that the functions have indentical functionalities

aeneasr:
Looks like there's a failing test in HSM: https://github.com/ory/hydra/runs/6376507740?check_suite_focus=true#step:6:4043

persistence/sql/transaction/transaction.go
grantzvolsky: Done. The implementation is a bit hacky where we need to check whether popx/GetConnection returned the fallback; lmk if I should add something along the lines of `GetConnectionOrNil` to popx to make the check below cleaner:
```
	fallback := &pop.Connection{TX: &pop.Tx{}}
	tx := popx.GetConnection(ctx, fallback)
	if tx.TX == fallback.TX || tx.TX == nil {
```

aeneasr (approved):

Issue #2989, opened 2022-02-14
CORS not available on /health/ready


rawnsley:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/hydra/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/hydra/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

When CORS is enabled for the public endpoints, the HTTP response headers are set correctly for `/.well-known/openid-configuration` but not for `/health/ready`.

This is the logical endpoint to use for confirming service health and we would like to `fetch` it from our generic HTML status page.

### Reproducing the bug

Returns expected result...

`curl 'https://HYDRA.SERVER/.well-known/openid-configuration' -H 'User-Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:82.0) Gecko/20100101 Firefox/82.0' -H 'Accept: application/json' -H 'Accept-Language: en-US,en;q=0.5' --compressed -H 'Origin: http://localhost:9595' -H 'Connection: keep-alive' -H 'Referer: http://localhost:9595/' -H 'Cache-Control: max-age=0' --verbose
`

Returns unexpected result...

`curl 'https://HYDRA.SERVER/health/ready' -H 'User-Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:82.0) Gecko/20100101 Firefox/82.0' -H 'Accept: application/json' -H 'Accept-Language: en-US,en;q=0.5' --compressed -H 'Origin: http://localhost:9595' -H 'Connection: keep-alive' -H 'Referer: http://localhost:9595/' -H 'Cache-Control: max-age=0' --verbose`

### Relevant configuration

```yml
server:
  cors:
    enabled: true
```

### Version

1.11.0

### On which operating system are you observing this issue?

Linux

### In which environment are you deploying?

Docker

aeneasr:
```
server:
  cors:
    enabled: true
```

[...]

PR #3114, opened 2022-05-18
FIX: Add CORS healthcheck handlers for alive/ready

barelyhuman:
defines and implements the
healthcheck struct to create a
definitive wrapper around the healthx handlers

also adds a cors middleware wrapper
for http.Handler types withouth the
iface defaults (might need changing)

## Related issue(s)
#2989  

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

Needed another PR due to a partially pushed (amended git commit) on the older branch that'd need a force push, hence this.

### DIFF
--- a/driver/registry_base.go
+++ b/driver/registry_base.go
@@ -7,6 +7,8 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/rs/cors"
+
 	"github.com/hashicorp/go-retryablehttp"
 
 	"github.com/ory/hydra/fositex"
@@ -96,6 +98,7 @@ type RegistryBase struct {
 	ats          jwk.JWTSigner
 	hmacs        *foauth2.HMACSHAStrategy
 	fc           *fositex.Config
+	publicCORS   *cors.Cors
 }
 
 func (m *RegistryBase) GetJWKSFetcherStrategy(ctx context.Context) fosite.JWKSFetcherStrategy {
@@ -136,11 +139,26 @@ func (m *RegistryBase) OAuth2AwareMiddleware(ctx context.Context) func(h http.Ha
 	return m.oa2mw
 }
 
+func (m *RegistryBase) addPublicCORSOnHandler(ctx context.Context) func(http.Handler) http.Handler {
+	corsConfig, corsEnabled := m.Config().CORS(ctx, config.PublicInterface)
+	if !corsEnabled {
+		return func(h http.Handler) http.Handler {
+			return h
+		}
+	}
+	if m.publicCORS == nil {
+		m.publicCORS = cors.New(corsConfig)
+	}
+	return func(h http.Handler) http.Handler {
+		return m.publicCORS.Handler(h)
+	}
+}
+
 func (m *RegistryBase) RegisterRoutes(ctx context.Context, admin *x.RouterAdmin, public *x.RouterPublic) {
 	m.HealthHandler().SetHealthRoutes(admin.Router, true)
 	m.HealthHandler().SetVersionRoutes(admin.Router)
 
-	m.HealthHandler().SetHealthRoutes(public.Router, false)
+	m.HealthHandler().SetHealthRoutes(public.Router, false, healthx.WithMiddleware(m.addPublicCORSOnHandler(ctx)))
 
 	admin.Handler("GET", prometheus.MetricsPrometheusPath, promhttp.Handler())
 
--- a/health/handler_test.go
+++ b/health/handler_test.go
@@ -1,0 +1,87 @@
+package health
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/ory/x/contextx"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/ory/hydra/driver/config"
+	"github.com/ory/hydra/internal"
+	"github.com/ory/hydra/x"
+	"github.com/ory/x/healthx"
+)
+
+func TestPublicHealthHandler(t *testing.T) {
+	ctx := context.Background()
+
+	doCORSRequest := func(t *testing.T, endpoint string) *http.Response {
+		req, err := http.NewRequest(http.MethodGet, endpoint, nil)
+		require.NoError(t, err)
+		req.Host = "example.com"
+		req.Header.Add("Origin", "https://example.com")
+		resp, err := http.DefaultClient.Do(req)
+		require.NoError(t, err)
+		return resp
+	}
+
+	expectCORSHeaders := func(t *testing.T, resp *http.Response) {
+		assert.Equal(t, http.StatusOK, resp.StatusCode)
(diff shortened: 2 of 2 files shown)


driver/registry_base.go
```diff
@@ -102,11 +104,26 @@ func (m *RegistryBase) OAuth2AwareMiddleware() func(h http.Handler) http.Handler
...
 
+func (m *RegistryBase) AddCORSonHandler(h http.Handler) http.Handler {
+	c := cors.New(cors.Options{
+		AllowedMethods:   []string{"GET", "POST", "PUT", "PATCH", "DELETE"},
+		AllowedHeaders:   []string{"Authorization", "Content-Type"},
+		ExposedHeaders:   []string{"Content-Type"},
+		AllowCredentials: true,
+	})
+	return c.Handler(h)
+}
```
Benehiko, line 115: I think we should create a new cors instance only once inside the registry base and base it on the config we have available as a `provider` in the `config` pacakge. See https://github.com/ory/hydra/blob/43b391d95f17cfd7414786cec0c602c15e29e956/driver/config/serve.go#L62-L69

Here is an example of how we use it https://github.com/ory/hydra/blob/63402dee7604141118fead91491abe6763150f1c/x/oauth2cors/cors.go#L41-L52

driver/registry_base.go
```diff
@@ -102,11 +104,26 @@ func (m *RegistryBase) OAuth2AwareMiddleware() func(h http.Handler) http.Handler
...
 	m.HealthHandler().SetHealthRoutes(admin.Router, true)
 	m.HealthHandler().SetVersionRoutes(admin.Router)
 
-	m.HealthHandler().SetHealthRoutes(public.Router, false)
+	// Register a healthcheck setup
+	hcHandlers := health.HealthCheck{}
+	hcHandlers.RegisterHealthHandlers(m.HealthHandler(), false)
+
+	public.Handler("GET", healthx.AliveCheckPath, m.AddCORSonHandler(hcHandlers.Alive))
+	public.Handler("GET", healthx.ReadyCheckPath, m.AddCORSonHandler(hcHandlers.Ready))
```
Benehiko, line 126: I could be wrong, but i don't think we need to recreate the healthx routes inside hydra since we have them defined in ory/x and it is being reused across all projects.

I think the best approach here is to wrap the health handler with the cors middleware.
We retrieve the health handler inside the `registry_base` so we should be able to wrap them here.

https://github.com/ory/hydra/blob/3115dde229a6be936ad4d844d778d6ee82279643/driver/registry_base.go#L217-L241

health/healthcheck.go
```diff
@@ -0,0 +1,50 @@
+package health
```
Benehiko, line 1: if we wrap our CORS middleware then the entire file here will be redundant

Benehiko (changes requested):

driver/registry_base.go
zepatrik: I guess @barelyhuman did this because

> I've intentionally not added it for the version route from the `healthx` library.

But I would also say we just wrap the whole health handler including the version.

barelyhuman: The thing is that the `cors` library needs a proper `http.Handler` reference and the `healthx` one's are built on top of httprouter's handler signature, also why I had to create the `HealthCheck` type, the type only wraps the healthx calls. 

I could get rid of the wrapper but then we'd need to either enable it on the all the routes of the router or add a new router that handles it and the new router would need to be on the same iface which complicates understanding what's going on

health/healthcheck.go
barelyhuman: That was one of the ideas , to just update the `x/healthx` package and wrap them there, though they wouldn't satisfy the cors package requirement for the handler type, we'd have to add the same structs there as well. 

If you wish for it to be changed in that repo, I'll do so.

driver/registry_base.go
Benehiko: i see, okay maybe let's rather then have a parameter on the healthx package that accepts the `router` and a `handler` middleware. Then in ory/x we do the wrapping of these paths if the middleware is defined 
https://github.com/ory/x/blob/master/healthx/handler.go#L84-L87

We can then just create the middleware inside hydra and then pass it along to ory/x.
I think this would be the best since we then don't duplicate code and don't need to refactor a lot

barelyhuman: https://github.com/ory/hydra/issues/2989#issuecomment-1128606791

back to the original set of thoughts then, 
cool, I'll raise a PR the healthx with the modification and also change this one accordingly

zepatrik: Sorry, we did not expect it to be that complicated here. You can first do the PR in `ory/x` and then adjust this one as needed once we merged it.

barelyhuman: It's not a biggie, will probably take this up post work today 👍🏼

Benehiko: Just out of interest I looked into the `healthx` package a bit and I can see it is poorly written. I imagined it would have an http.Handler return type so you could just wrap it in your middleware.

For a quick and dirty solution one could probably do something like this 

```go
opt, _ := m.C.CORS(config.PublicInterface)
c := cors.New(opt)

public.Handler("GET", healthx.AliveCheckPath, c.Handler(http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
   m.HealthHandler().Alive(writer, request, nil)
})))

public.Handler("GET", healthx.ReadyCheckPath, c.Handler(http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
   m.HealthHandler().Ready(false)(writer, request, nil)
})))
```

but yeah i think it's best to fix the healthx package :laughing:

barelyhuman: > Just out of interest I looked into the `healthx` package a bit and I can see it is poorly written. I imagined it would have an `http.Handler` return type so you could just wrap it in your middleware.
> 
> For a quick and dirty solution one could probably do something like this
> 
> ```go
> opt, _ := m.C.CORS(config.PublicInterface)
> c := cors.New(opt)
> 
> public.Handler("GET", healthx.AliveCheckPath, c.Handler(http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
>    m.HealthHandler().Alive(writer, request, nil)
> })))
> 
> public.Handler("GET", healthx.ReadyCheckPath, c.Handler(http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
>    m.HealthHandler().Ready(false)(writer, request, nil)
> })))
> ```
> 
> but yeah i think it's best to fix the healthx package 😆

I did do this, then I moved it to match with the New + Composition pattern being used everywhere and ended up refactoring to use a new struct altogether 

Though, as long as the repos using healthx are version locked , i think the changes made there should be fine. 
Wouldn't wanna break everything

Benehiko: yes they are version locked :)

zepatrik (approved):
I figured it would be easier to make this against the v2.x branch as that already uses the newest ory/x.
Also, I refactored the test a bit.

Benehiko (approved):
nice :)

PR #3115, opened 2022-05-18
Janitor optimization use requested at index

arnolf:
Hi,

on large tables, janitor queries could take minutes.

Table like refresh_token can grow before there is some expired tokens.

By adding missing indexes on requested_at columns, and by ordering on this field we can avoid full table scan, and/or in memory sorting.

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/internal/testhelpers/janitor_test_helper.go
+++ b/internal/testhelpers/janitor_test_helper.go
@@ -507,7 +507,7 @@ func JanitorTests(conf *config.Provider, consentManager consent.Manager, clientM
 
 				// cleanup
 				t.Run("step=cleanup", func(t *testing.T) {
-					require.NoError(t, fositeManager.FlushInactiveLoginConsentRequests(ctx, time.Now().Round(time.Second), 2, 1))
+					require.NoError(t, fositeManager.FlushInactiveLoginConsentRequests(ctx, time.Now().Round(time.Second), 4, 1))
 				})
 
 				// validate
--- a/persistence/sql/migrations/20220516123200000000_oauth2.down.sql
+++ b/persistence/sql/migrations/20220516123200000000_oauth2.down.sql
@@ -1,4 +1,3 @@
-DROP INDEX hydra_oauth2_access_expiry;
-DROP INDEX hydra_oauth2_refresh_expiry;
-DROP INDEX hydra_oauth2_authentication_request_expiry;
-DROP INDEX hydra_oauth2_consent_request_expiry;
+DROP INDEX hydra_oauth2_refresh_requested_at_idx;
+DROP INDEX hydra_oauth2_authentication_request_requested_at_idx;
+DROP INDEX hydra_oauth2_consent_request_requested_at_idx;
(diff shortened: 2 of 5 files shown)


aeneasr (changes requested):
Thank you very much! We are in the process of refactoring the storage layer. Could you please:

1. Confirm this issue exists on the v2.x branch
2. If it still exists, make your changes against that branch

Thank you very much!

arnolf:
Hello,

I had a look on v2.x branch. It seems the code changed a lot.
The authentication and request tables has been merged in an unique flow tables, so I can't give you production feedback on this table as the v2 has not yet been released :(

However, the main issue still arise on refresh_token table even if the query changed (https://github.com/ory/hydra/blob/v2.x/persistence/sql/persister_oauth2.go#L430) :
- it is still missing an hydra_oauth2_refresh_requested_at_idx index on (nid, requested_at)
- and ORDER BY still using signature instead of requested_at

By the way, I think it would be better if the existing index on access_token table would be on (nid, requested_at), rather than on (requested_at, nid). With composite index, it's always better to put more selective index at first, that is, equality first, and then comparison.

I've made tests we a small amount of data, and forcing usage of indexes on PostgreSQL:

Here the actual query plan :
Limit  (cost=12.27..12.28 rows=1 width=52) (actual time=0.021..0.021 rows=0 loops=1)
  ->  Sort  (cost=12.27..12.28 rows=1 width=52) (actual time=0.020..0.021 rows=0 loops=1)
        Sort Key: requested_at
        Sort Method: quicksort  Memory: 25kB
        ->  Index Scan using hydra_oauth2_refresh_client_id_idx on hydra_oauth2_refresh  (cost=0.14..12.26 rows=1 width=52) (actual time=0.014..0.015 rows=0 loops=1)
              Index Cond: (nid = '4511a21c-67e1-4a80-8999-444c1a418ce6'::uuid)
              Filter: (requested_at < '2021-12-12 00:00:00'::timestamp without time zone)
              Rows Removed by Filter: 1
Planning time: 0.149 ms
Execution time: 0.051 ms

With index on (nid, requested_at)
Limit  (cost=8.16..8.17 rows=1 width=44) (actual time=0.015..0.015 rows=0 loops=1)
  ->  Sort  (cost=8.16..8.17 rows=1 width=44) (actual time=0.014..0.015 rows=0 loops=1)
        Sort Key: signature
        Sort Method: quicksort  Memory: 25kB
        ->  Index Scan using hydra_oauth2_refresh_requested_at_idx on hydra_oauth2_refresh  (cost=0.14..8.15 rows=1 width=44) (actual time=0.009..0.009 rows=0 loops=1)
              Index Cond: ((nid = '4511a21c-67e1-4a80-8999-444c1a418ce6'::uuid) AND (requested_at < '2021-12-12 00:00:00'::timestamp without time zone))
Planning time: 0.101 ms
Execution time: 0.034 ms

With index on (nid, requested_at) and ORDER BY requested_at, we have no sorting, no full scan, and we have full usage of index. :
Limit  (cost=0.14..8.15 rows=1 width=52) (actual time=0.011..0.011 rows=0 loops=1)
  ->  Index Scan using hydra_oauth2_refresh_requested_at_idx on hydra_oauth2_refresh  (cost=0.14..8.15 rows=1 width=52) (actual time=0.010..0.010 rows=0 loops=1)
        Index Cond: ((nid = '4511a21c-67e1-4a80-8999-444c1a418ce6'::uuid) AND (requested_at < '2021-12-12 00:00:00'::timestamp without time zone))
Planning time: 0.124 ms
Execution time: 0.036 ms

Since the code change a lot, I think it would be better to create a new pull request, or an issue (I don't know yet if I would have time to create a new pull request).

aeneasr:
Superseded by #3516

Issue #3082, opened 2022-04-21
Pass current session (access token extensions) to refresh token hook


alekbarszczewski:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/hydra/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/hydra/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

I am adding custom session extensions to access token when accepting consent request:
```csharp
var request = new HydraAcceptConsentRequest(
    grantScope: scope.ToList(),
    session: new HydraConsentRequestSession(accessToken: new Dictionary<string, object>()
    {
        { "sid", sessionId }
    })
);

var result = await _adminApi.AcceptConsentRequestAsync(consentChallenge, request);
```

And JWT token looks like this:
```json
{
  "sub": "xxx",
  ...
  "ext": { // <--- extensions
    "sid": "14628d10-8c63-4c7f-8a9c-3e37318ce7d6"
  },
  ...
}
```

Normally when token is refreshed "ext" field remain the same after each refresh which is good, "sid" remains the same which is what I want. However when using refresh token hook (a webhook which is called before token refresh to refresh claims described here https://www.ory.sh/docs/hydra/guides/claims-at-refresh#webhook-configuration) there is no option to keep "ext" field the same.

1. When refresh token webhook returns 200 OK with empty json then "ext" field is removed from new JWT
2. When refresh token webhook returns 200 OK with no body then hydra throws error (body parsing error)
3. When refresh token webhook return 204 No Content then hydra throws error (it's expecting 200 and json content)

I could return extensions from refresh token webhook, but the problem is that previous extensions are not passed to the webhook payload which looks like this:

```
{
  "subject": "foo",
  "client_id": "bar",
  "granted_scopes": ["openid", "offline"],
  "granted_audience": []
}
```

[...]

PR #3146, opened 2022-06-11
feat: add ability to allow token refresh from hook without overriding the session claims

zachabney:
The refresh token hook currently requires a 200 response code with an updated session payload. There is no way to permit the token refresh without overriding the session data. It's useful, however, to be able to use the token refresh hook in order to accept/reject token refreshes without the need to update the session data.

This adds the ability for the token refresh hook to return a `204 No Content` response code to indicate that the token refresh is permitted, but not to update the session data.

The documentation has also been updated with PR https://github.com/ory/docs/pull/863

## Related issue(s)

#3082 

## Checklist

- [X] I have referenced an issue containing the design document if my change introduces a new feature.
- [X] I am following the [contributing code guidelines](../blob/master/CONTRIBUTING.md#contributing-code).
- [X] I confirm that this pull request does not address a security vulnerability. If this pull request addresses a security.
      vulnerability, I confirm that I got green light (please contact [security@ory.sh](mailto:security@ory.sh)) from the
      maintainers to push the changes.
- [X] I have added tests that prove my fix is effective or that my feature works.
- [X] I have added or changed [the documentation](https://github.com/ory/docs).

## Further Comments

In the future, the token refresh hook request payload could be expanded to include the existing session data and allow for the `sid` to be set on the refreshed id token as described in #3082, but this is a quick workaround that'll continue to be useful even if that is implemented.

### DIFF
--- a/oauth2/oauth2_auth_code_test.go
+++ b/oauth2/oauth2_auth_code_test.go
@@ -968,6 +968,24 @@ func TestAuthCodeWithMockStrategy(t *testing.T) {
 						require.True(t, gjson.GetBytes(idTokenBody, "hooked").Bool())
 					})
 
+					t.Run("should not override session data if token refresh hook returns no content", func(t *testing.T) {
+						hs := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+							w.WriteHeader(http.StatusNoContent)
+						}))
+						defer hs.Close()
+
+						conf.MustSet(config.KeyRefreshTokenHookURL, hs.URL)
+						defer conf.MustSet(config.KeyRefreshTokenHookURL, nil)
+
+						res, err := testRefresh(t, &refreshedToken, ts.URL, false)
+						require.NoError(t, err)
+						assert.Equal(t, http.StatusOK, res.StatusCode)
+
+						body, err := ioutil.ReadAll(res.Body)
+						require.NoError(t, err)
+						require.NoError(t, json.Unmarshal(body, &refreshedToken))
+					})
+
 					t.Run("should fail token refresh with `server_error` if hook fails", func(t *testing.T) {
 						hs := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 							w.WriteHeader(http.StatusInternalServerError)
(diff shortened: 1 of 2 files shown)


oauth2/oauth2_auth_code_test.go
```diff
@@ -968,6 +968,24 @@ func TestAuthCodeWithMockStrategy(t *testing.T) {
...
+						conf.MustSet(config.KeyRefreshTokenHookURL, hs.URL)
+						defer conf.MustSet(config.KeyRefreshTokenHookURL, nil)
+
+						res, err := testRefresh(t, &refreshedToken, ts.URL, false)
+						require.NoError(t, err)
+						assert.Equal(t, http.StatusOK, res.StatusCode)
+
+						body, err := ioutil.ReadAll(res.Body)
+						require.NoError(t, err)
+						require.NoError(t, json.Unmarshal(body, &refreshedToken))
```
aeneasr, line 986: Could you please add one assertion that checks whether the contents actually remained the same? You can probably copy paste from here: https://github.com/ory/hydra/pull/3146/files#diff-6d883efffdabd9715dc9872121018df30a5843c81e25dc6c4af2c3edc13fb21cR957-R968

aeneasr (changes requested):

oauth2/oauth2_auth_code_test.go
zachabney: Certainly! Good idea. That's been added.

oauth2/oauth2_auth_code_test.go
```diff
@@ -969,21 +969,40 @@ func TestAuthCodeWithMockStrategy(t *testing.T) {
 					})
 
 					t.Run("should not override session data if token refresh hook returns no content", func(t *testing.T) {
+						if strat.d != "jwt" {
+							t.Skip()
```
aeneasr, line 973: This test would work for both JWT and opaque if you use:

```
accessTokenClaims := testhelpers.IntrospectToken(t, oauthConfig, &refreshedToken, ts)
```

https://github.com/ory/hydra/pull/3146/commits/82a919edb23de6729ceab4fc48d4471199f25bff#diff-6d883efffdabd9715dc9872121018df30a5843c81e25dc6c4af2c3edc13fb21cR957

The problem is that the JWT strategy works a bit different so we need to ensure this works as intended for both strategies :)

aeneasr (changes requested):

oauth2/oauth2_auth_code_test.go
zachabney: Ah, that's what I was missing. Thanks! Should be updated now.

aeneasr (approved):

PR #3259, opened 2022-09-16
feat: add /.well-known/jwks.json caching for HSM key manager

aarmam:
This PR adds caching for well known keys when using HSM key manager. Operations per sec on HSM are limited and caching well known keys would make sense. 
Also if I understand https://github.com/ThalesIgnite/crypto11/blob/master/crypto11.go correctly, then crypto.Signer is thread-safe and hsm.GetKey/hsm.GetKeySet could be cached also, but I will create another PR for this.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/driver/registry_sql.go
+++ b/driver/registry_sql.go
@@ -138,7 +138,7 @@ func (m *RegistrySQL) Init(
 		}
 
 		if m.Config().HSMEnabled() {
-			hardwareKeyManager := hsm.NewKeyManager(m.HSMContext(), m.Config())
+			hardwareKeyManager := hsm.NewKeyManager(m.HSMContext(), m.Config(), m.Logger())
 			m.defaultKeyManager = jwk.NewManagerStrategy(hardwareKeyManager, m.persister)
 		} else {
 			m.defaultKeyManager = m.persister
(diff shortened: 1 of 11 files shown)


aeneasr (changes requested):
Thank you for the contribution! I looked a bit through the code. Getting caching to work is very hard in security components, because any mistake (cache key collision, cache invalidation, ...) can lead to catastrophic outcomes. The code currently has several problems:

1. Cache invalidation across multiple nodes is unsolved (what happenes to the cache on container A when you rotate the key on container B?)
2. Cache key collision is possible because the cache key is the set id which is not guaranteed to be unique for two distinct keys
3. No locking mechanism for concurrent access on the cache map
4. No limit on cache size and thus memory consumption

Before introducing caching in Ory Hydra, I would recommend to write a design document outlining the challenges and approaches of how to solve caching for this component at scale.

We are generally very tough on merging features that introduce caches, because caches have been the no 1 offender when it comes to security CVEs at Ory. It is quite likely that a cache feature such as this one will never be merged to master.

aarmam:
> 1. Cache invalidation across multiple nodes is unsolved (what happenes to the cache on container A when you rotate the key on container B?)

Right this is a problem in current implementation.

> 2. Cache key collision is possible because the cache key is the set id which is not guaranteed to be unique for two distinct keys

Yes keySetCache contains all keys with set id. When any of these keys are deleted by DeleteKey or DeleteKeySet, then the cache key is evicted

> 3. No locking mechanism for concurrent access on the cache map

Locking is implemented using RWMutex on methods.

> 4. No limit on cache size and thus memory consumption

As this feature was intended for GetWellKnownKeys only I assumed there is reasonable amount of keys.

> Before introducing caching in Ory Hydra, I would recommend to write a design document outlining the challenges and approaches of how to solve caching for this component at scale.
> We are generally very tough on merging features that introduce caches, because caches have been the no 1 offender when it comes to security CVEs at Ory. It is quite likely that a cache feature such as this one will never be merged to master.

Thanks even caching was intended for GetWellKnownKeys only im not sure if we will pursue implementing it then. We initially thought caching well known keys would be useful for deployments using cloud HSM where each operation costs.

Issue #2850, opened 2021-11-10
Watching tls cert/key file changes to reload the server automatically


StarAurryon:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/hydra/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/hydra/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

When using hydra within kubernetes and with certmanager the TLS certificates are only valid for 90 days. After 60 days cert manager renew the cert and update the files in the Pod.

I have not found anything in the doc.

### Describe your ideal solution

Hydra (but also Kratos) could watch the TLS key and cert files if defined in the config file for change and restart the http server if needed.

### Workarounds or alternatives

Restarting the service with an external supervisor.

### Version

All

### Additional Context

If there is no solution, as always I can provide a PR ;)

aeneasr:
PRs welcomed :) Technically, I think this probably needs a HTTP server restart which might be difficult (but probably achievable) to implement given the way how we bootstrap.

It will need a lot of tests to ensure the restart happens cleanly (e.g. no left-over config instances, no left-over FS watches, etc).

zepatrik:
One could also not re-initialize everything, but instead keep the registry and just start a new `http.Server` that uses the same handlers? That results still in a short down-time I assume, but if you instead keep the same `net.Socket` and only create a new server it could even happen with zero downtime?

StarAurryon:
> That results still in a short down-time
I don't know how go handle the closing of the http server. I hope that it finalizes current requests properly. After that clients might have their tls session dropped as the key/cert will be changed. I don't know how this is handled with HTTP2 as multiple requests can use the same TCP session, if i am not wrong.
> If you instead keep the same net.Socket
I believe that's the way to go as you don't have to check again if the port is already bound and if Kubernetes check it at a wrong time the service could be considered as down. Its better to keep the port open and delay the request acceptance.

[...]

PR #3265, opened 2022-09-22
feat: TLS certificate hot-reloading

alnr:
Automatic TLS certificate hot-reloading when the cert changes on disk.

#2850


cmd/server/helper_cert.go
```diff
@@ -57,66 +56,70 @@ func AttachCertificate(priv *jose.JSONWebKey, cert *x509.Certificate) {
...
 		d.Logger().WithError(err).Fatal("Unable to fetch or generate HTTPS TLS key pair")
+		return nil // in case Fatal is hooked
 	}
 
 	if len(priv.Certificates) == 0 {
 		cert, err := tlsx.CreateSelfSignedCertificate(priv.Key)
 		if err != nil {
-			d.Logger().WithError(err).Fatalf(`Could not generate a self signed TLS certificate`)
+			d.Logger().WithError(err).Fatal(`Could not generate a self signed TLS certificate`)
+			return nil // in case Fatal is hooked
```
aeneasr, line 88: I think we should instead return the error

cmd/server/helper_cert.go
```diff
@@ -57,66 +56,70 @@ func AttachCertificate(priv *jose.JSONWebKey, cert *x509.Certificate) {
...
 
 		AttachCertificate(priv, cert)
 		if err := d.SoftwareKeyManager().DeleteKey(ctx, TlsKeyName, priv.KeyID); err != nil {
 			d.Logger().WithError(err).Fatal(`Could not update (delete) the self signed TLS certificate`)
+			return nil // in case Fatal is hooked
 		}
 
 		if err := d.SoftwareKeyManager().AddKey(ctx, TlsKeyName, priv); err != nil {
 			d.Logger().WithError(err).Fatalf(`Could not update (add) the self signed TLS certificate: %s %x %d`, cert.SignatureAlgorithm, cert.Signature, len(cert.Signature))
+			return nil // in case Fatalf is hooked
```
aeneasr, line 99: I think we should instead return the error

cmd/server/helper_cert.go
```diff
@@ -57,66 +56,70 @@ func AttachCertificate(priv *jose.JSONWebKey, cert *x509.Certificate) {
...
+		return nil // in case Fatal is hooked
 	}
 
 	pemCert := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: priv.Certificates[0].Raw})
 	pemKey := pem.EncodeToMemory(block)
 	ct, err := tls.X509KeyPair(pemCert, pemKey)
 	if err != nil {
-		d.Logger().WithError(err).Fatalf("Could not decode certificate")
+		d.Logger().WithError(err).Fatal("Could not decode certificate")
+		return nil // in case Fatal is hooked
```
aeneasr, line 119: I think we should instead return the error

cmd/server/helper_cert.go
```diff
@@ -57,66 +56,70 @@ func AttachCertificate(priv *jose.JSONWebKey, cert *x509.Certificate) {
...
 	block, err := jwk.PEMBlockForKey(priv.Key)
 	if err != nil {
-		d.Logger().WithError(err).Fatalf("Could not encode key to PEM")
+		d.Logger().WithError(err).Fatal("Could not encode key to PEM")
+		return nil // in case Fatal is hooked
 	}
 
 	if len(priv.Certificates) == 0 {
 		d.Logger().Fatal("TLS certificate chain can not be empty")
+		return nil // in case Fatal is hooked
```
aeneasr, line 111: I think we should instead return the error

cmd/server/helper_cert.go
```diff
@@ -57,66 +56,70 @@ func AttachCertificate(priv *jose.JSONWebKey, cert *x509.Certificate) {
...
 			d.Logger().WithError(err).Fatalf(`Could not update (add) the self signed TLS certificate: %s %x %d`, cert.SignatureAlgorithm, cert.Signature, len(cert.Signature))
+			return nil // in case Fatalf is hooked
 		}
 	}
 
 	block, err := jwk.PEMBlockForKey(priv.Key)
 	if err != nil {
-		d.Logger().WithError(err).Fatalf("Could not encode key to PEM")
+		d.Logger().WithError(err).Fatal("Could not encode key to PEM")
+		return nil // in case Fatal is hooked
```
aeneasr, line 106: I think we should instead return the error

aeneasr (changes requested):
Please always return the error if it is available. In a possible scenario, someone could accidentally overwrite the hook to not fatal, which would then end up in incorrectly returning a nil error, most likely causing a panic somewhere down the line. Thanks!

alnr:
> Please always return the error if it is available. In a possible scenario, someone could accidentally overwrite the hook to not fatal, which would then end up in incorrectly returning a nil error, most likely causing a panic somewhere down the line. Thanks!

I did not change the interface semantics of `GetOrCreateTLSCertificate` it this PR: it is designed to terminate the process through the logger's ExitFunc if any error is encountered. That may or may not be the best design, but I decided to keep the semantics in this PR.

Without the `return nil` lines added in this PR, the logic was super broken if `ExitFunc` was hooked as the execution would simply continue nonsensically.

We have two options:
- replace `return nil` by `panic()` to guarantee the execution doesn't continue after `logger.Fatalf()` (easy fix)
- change the interface of `GetOrCreateTLSCertificate` to return an error in addition to the certificate info, and leave the logging to the caller (more involved fix).

aeneasr:
Ah, my bad, I overlooked that we return the certificate here and not the error. Ok, then this behavior is acceptable :)

PR #3333, opened 2022-11-02
fix: :wrench: checking for columns existence.

architeacher:
[This](https://github.com/ory/hydra/blob/v1.11.10/persistence/sql/migrations/20210928155900000000_support_amr_claim.mysql.up.sql) migration runs smoothly on a small size database, but once the size of the data in the table gets bigger, the altering of the table would work fine.
On the other hand the 2nd step which updates the field `amr` data
```sql
UPDATE hydra_oauth2_consent_request SET amr='';
```
would fail with the following error:
```bash
ERROR 1370 (42000): alter routine command denied to user 'hydra'@'%' for routine 'hydra.partialUpdates'
```
So if you try the re-run the migration again, it would complain that the column `amr` already exists, and in order to fix that I wrote this stored procedure to check of column existence.

## Checklist

      introduces a new feature.
- [ x] I am following the
      the changes.
      works.


aeneasr (changes requested):
Thank you for the PR! Unfortunately, we can't change SQL migrations retroactively which is why we can't merge this PR :/

architeacher:
@aeneasr Thanks for the time, you put into reviewing this PR, I can see your concern which is valid.
This migration does not add or remove anything to the schema other than the changes that were already existing, it was just rewritten to consider failure for large tables when being migrated, such that if you re-run the migration script again it would not complain from a column that was already created.

aeneasr:
Thank you, I've decided to not merge this PR. The primary reason is that I don't understand the procedure and that it seems to be related to a very specific edge case which could happen on MySQL. What bothers me in general is why MySQL seems to not have rolled back the transaction and schema change. But this could be due to a limitation in MySQL itself. It's hard to say.

PR #3338, opened 2022-11-04
chore: update Ory CLI with breaking changes to the format task

kevgo:
https://github.com/ory/cli/pull/257 changed the name of the command to add copyright headers to files.


kevgo:
When Ory CLI v0.1.48 ships, the format task will turn green and you can ship this PR.

Issue #2911, opened 2022-01-03
No longer allow users to set the client ID


aeneasr:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/hydra/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/hydra/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

It should no longer be possible to set the Client ID.

### Describe your ideal solution

The Client ID should be generated by the server.

### Workarounds or alternatives

None

### Version

master

eric-poitras:
What is the rationale behind this ?

aeneasr:
It’s according to spec

tsoestini:
Is there some context for understanding this?
A link to what is changing?
A link to the spec?
Who is the user?
I don't allow my users to set the client id, but it's important that I be able to set it. I can't imagine an OAuth spec would rule that out.

mig5:
As far as I can see, the oAuth2.0 standard doesn't make any statement about whether the client id can be manually defined or not.

https://www.rfc-editor.org/rfc/rfc6749#section-2.2

>    The authorization server issues the registered client a client
   identifier -- a unique string representing the registration
   information provided by the client.  The client identifier is not a
   secret; it is exposed to the resource owner and MUST NOT be used
   alone for client authentication.  The client identifier is unique to
   the authorization server.
>
>    The client identifier string size is left undefined by this
   specification.  The client should avoid making assumptions about the
   identifier size.  The authorization server SHOULD document the size
   of any identifier it issues.
   
On the other hand, OAuth.com's document https://www.oauth.com/oauth2-servers/client-registration/client-id-secret/ says
   
   > The client_id is a public identifier for apps. Even though it’s public, it’s best that it isn’t guessable by third parties, so many implementations use something like a 32-character hex string. If the client ID is guessable, it makes it slightly easier to craft phishing attacks against arbitrary applications. It must also be unique across all clients that the authorization server handles.
   
[...]

PR #3354, opened 2022-11-10
feat: add config flag to allow specifying clientId on client creation

tinolazreg:
Add a configuration flag to override the new change in Hydra v2.0, which made it no longer possible to choose the OAuth 2.0 Client ID. 
With this change, we will still enforce that the Client ID has to be a UUID, but it gives more flexibility to the users of Hydra.
A lot of the arguments for allowing such a flag is already detailed and discussed in the related issue.

## Related issue(s)
https://github.com/ory/hydra/issues/2911

## Checklist
      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/client/handler.go
+++ b/client/handler.go
@@ -7,11 +7,12 @@ import (
 	"context"
 	"crypto/subtle"
 	"encoding/json"
-	"github.com/ory/hydra/driver/config"
 	"io"
 	"net/http"
 	"strings"
 	"time"
+
+	"github.com/ory/hydra/driver/config"
 
 	"github.com/ory/x/pagination/tokenpagination"
 
--- a/client/handler.go
+++ b/client/handler.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"crypto/subtle"
 	"encoding/json"
+	"github.com/ory/hydra/driver/config"
 	"io"
 	"net/http"
 	"strings"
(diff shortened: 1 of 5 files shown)


aeneasr (changes requested):
Unfortunately, as explained in [this comment](https://github.com/ory/hydra/issues/2911#issuecomment-1298440000), it's not possible to merge this PR. With this PR, we just push the inevitable (not being able to use your own IDs; ory changing to a new uuid format) down the line. I do understand that it is annoying that this is no longer possible, and Ory wants to be as helpful as we can in softening the impact of this change with pointed guides that hopefully help users adapt.

tinolazreg:
> Unfortunately, as explained in [this comment](https://github.com/ory/hydra/issues/2911#issuecomment-1298440000), it's not possible to merge this PR. With this PR, we just push the inevitable (not being able to use your own IDs; ory changing to a new uuid format) down the line. I do understand that it is annoying that this is no longer possible, and Ory wants to be as helpful as we can in softening the impact of this change with pointed guides that hopefully help users adapt.

Worth a try :) I thought maybe you could allow it, since it will be an opt-in flag that we would use at our own risk. But I get your point. I won't try to pursue this any further.

aeneasr:
Thank you for your understanding! Closing as discussed

PR #3424, opened 2023-01-25
chore: append /v2 to module path

alnr:
This should allow others who depend on this module to import it.

Fixed a couple other small nits.


aeneasr (approved):
LGTM

Issue #3244, opened 2022-09-05
Add token web hook for all grants types 


fehrnah:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/hydra/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/hydra/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

Currently there exists a web hooks for refresh_token grants, but customizing other grant types is either not possible or not easy.

For example, customizing access_code tokens can be done in the consent endpoint, but customizing client_credentials is not possible.

### Describe your ideal solution

Add access_token, jwt_profile and client_credentials web hooks working in a similar way to the existing refresh_token web hook.

### Workarounds or alternatives

#1748 Would allow limited customization of all claims, not allowing more dynamic claims.
#1383 Is about setting static claims to client_credential tokens

### Version

2.0

synclpz:
I have some demand for on-the-fly scope/audience check due to API consumer might have changed it's subscriptions on the API portal which has different client (consumer) lifecycle than Hydra has. Also inserting some API-GW related consumer ID should make call control easier on GW side. This is still not possible using client_credentials flow with Hydra.

aeneasr:
Yeah, I think this is a pretty good idea

fehrnah:
Great, I'm already working on it, I still have some tests to run nut I should be able to create a PR soon.

PR #3427, opened 2023-01-29
feat: add token request hooks for all grant types

sgal:
Added a generic token hook that is called for all grant types and includes `payload` with a single allowed value - `assertion` to cover the `jwt-bearer` grant type customization.

Existing `refresh hook` is left unchanged and will be deprecated in favor of the new hook.

## Related issue(s)
#3244
https://github.com/ory/fosite/issues/729

This change completes and extends the work done in https://github.com/ory/hydra/pull/3254

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

Docs update PR - https://github.com/ory/docs/pull/1241


oauth2/hook.go
```diff
@@ -66,103 +70,159 @@ type RefreshTokenHookResponse struct {
...
-		if err := json.NewDecoder(resp.Body).Decode(&respBody); err != nil {
-			return errorsx.WithStack(
-				fosite.ErrServerError.
-					WithWrap(err).
-					WithDescription("The refresh token hook target responded with an error.").
-					WithDebugf("Response from refresh token hook could not be decoded: %s", err),
-			)
-		}
+	if requester.GetGrantTypes().ExactOne("urn:ietf:params:oauth:grant-type:jwt-bearer") || requester.GetGrantTypes().ExactOne("client_credentials") {
+		payload = requester.GetRequestForm()
```
sgal, line 142: Added request payload to hook request only for jwt_bearer and client_credentials grant types. Other grant types don't have anything in their payload that could be useful in the hook.

oauth2/oauth2_auth_code_test.go
```diff
@@ -235,7 +235,7 @@ func TestAuthCodeWithDefaultStrategy(t *testing.T) {
 		assert.True(t, i.Get("active").Bool(), "%s", i)
 		assert.EqualValues(t, conf.ClientID, i.Get("client_id").String(), "%s", i)
 		assert.EqualValues(t, expectedSubject, i.Get("sub").String(), "%s", i)
-		assert.EqualValues(t, `{"foo":"bar"}`, i.Get("ext").Raw, "%s", i)
+		assert.EqualValues(t, `bar`, i.Get("ext.foo").String(), "%s", i)
```
sgal, line 238: Changed this to allow for being able to verify ext fields added in the hook tests.

oauth2/hook.go
```diff
@@ -66,103 +70,159 @@ type RefreshTokenHookResponse struct {
...
-		var respBody RefreshTokenHookResponse
-		if err := json.NewDecoder(resp.Body).Decode(&respBody); err != nil {
-			return errorsx.WithStack(
-				fosite.ErrServerError.
-					WithWrap(err).
-					WithDescription("The refresh token hook target responded with an error.").
-					WithDebugf("Response from refresh token hook could not be decoded: %s", err),
-			)
-		}
+	if requester.GetGrantTypes().ExactOne("urn:ietf:params:oauth:grant-type:jwt-bearer") || requester.GetGrantTypes().ExactOne("client_credentials") {
```
hperl, line 141: Can't we just check the `hookType` here instead of the requester grants?

hperl (changes requested):
This looks very good! I just had some minor questions & suggestions.

sgal:
@hperl I fixed the comments, please have a look!

oauth2/hook.go
hperl: Makes sense!

hperl (dismissed):
LGTM! :tada:

oauth2/hook.go
```diff
@@ -66,103 +70,159 @@ type RefreshTokenHookResponse struct {
...
+	}
+
+	reqBody := TokenHookRequest{
+		Session:         session,
+		Requester:       requesterInfo,
+		Subject:         session.GetSubject(),
+		ClientID:        requester.GetClient().GetID(),
+		GrantedScopes:   requester.GetGrantedScopes(),
+		GrantedAudience: requester.GetGrantedAudience(),
+	}
```
aeneasr, line 160: I know it was like this before, but isn't the information kinda duplicated?

oauth2/hook.go
```diff
@@ -66,103 +70,159 @@ type RefreshTokenHookResponse struct {
...
-		if err := json.NewDecoder(resp.Body).Decode(&respBody); err != nil {
-			return errorsx.WithStack(
-				fosite.ErrServerError.
-					WithWrap(err).
-					WithDescription("The refresh token hook target responded with an error.").
-					WithDebugf("Response from refresh token hook could not be decoded: %s", err),
-			)
-		}
+	if grantType == "urn:ietf:params:oauth:grant-type:jwt-bearer" || grantType == "client_credentials" {
+		payload = requester.GetRequestForm()
```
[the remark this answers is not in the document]
sgal: Are secrets sent in the form data? I believe those are sent in the headers, which are not part of this value.

On the why part: only `jwt-bearer` and `client-credentials` can have a meaningful payload that would not be available elsewhere.

- `jwt-bearer` - `assertion` JWT that could be used in the webhook to read custom claims and make business-specific decisions
- `client_credentials` - not really sure how would one use the payload here, I'm open to remove it as well.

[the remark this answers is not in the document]
sgal: If we need to strip sensitive values, do you have the list anywhere?

sgal:
@aeneasr I fixed the following parts:
- Remove `client_secret` from the `payload` that is sent to the webhook
- Only send the `payload` in the `jwt-bearer` grant type
- Remove the data duplication in the webhook request

Please have a look.

sgal:
Docs are also updated - https://github.com/ory/docs/pull/1241

oauth2/.snapshots/TestAuthCodeWithMockStrategy-strategy=opaque-case=0-description=should_pass_request_if_strategy_passes-should_call_refresh_token_hook_if_configured.json
```diff
@@ -32,16 +32,10 @@
     "allowed_top_level_claims": []
   },
   "requester": {
-    "client_id": "app-client",
```
aeneasr, line 35: This would be a breaking change. I'm not sure how to handle this? Two thoughts come to mind:

1. It probably makes sense to remove the inner duplication (why would someone use the nested keys)
2. The `requester` key is probably more accurate and groups (similar to session) the key correctly

What do you think?

spec/config.json
```diff
@@ -975,6 +975,24 @@
           "description": "Sets the refresh token hook endpoint. If set it will be called during token refresh to receive updated token claims.",
           "format": "uri",
           "examples": ["https://my-example.app/token-refresh-hook"]
+        },
+        "authorization_code_hook": {
```
aeneasr, line 979: Do we need all of these config elements? Maybe it makes more sense to have one endpoint and pass the grant type and let the webhook deal with that?

sgal: Yeah, that makes much more sense. Let me take a stab at it.

aeneasr: To keep backwards compatibility we could also just add a new key "hook" and deprecate the refresh token hook at some point. That way we don't break BC?

sgal:
@aeneasr Changed to having a generic `token hook` config, along with the existing one. Cleaned up the request payload for the new hook to remove duplication.

aeneasr (approved):
Very nice

sgal:
@fehrnah FYI, it's finally in! Thanks for laying the foundation for this feature 👍

fehrnah:
> @fehrnah FYI, it's finally in! Thanks for laying the foundation for this feature 👍 

Congratulations!

aeneasr:
Thank you both :)

PR #3431, opened 2023-02-05
fix: enable datadog tracer

sawadashota:
Load OpenTelemetry configuration to set up datadog tracer.

## Related issue(s)

none

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.


.docker/Dockerfile-build
```diff
@@ -6,7 +6,7 @@ WORKDIR /go/src/github.com/ory/hydra
 RUN mkdir -p ./internal/httpclient
 
 COPY go.mod go.sum ./
-COPY internal/httpclient/go.* ./internal/httpclient
+COPY internal/httpclient/go.* ./internal/httpclient/
```
sawadashota, line 9: Without a trailing slash, it will fail.

```
Step 6/21 : COPY internal/httpclient/go.* ./internal/httpclient
When using COPY with more than one source file, the destination must be a directory and end with a /
make: *** [docker] Error 1
```

aeneasr:
Seems like tests are failing, can you PTAL? :)

sawadashota:
@aeneasr I'm sorry, but I've fixed it!

spec/config.json
```diff
@@ -1022,154 +1022,7 @@
...
-              "examples": [
-                {
-                  "server_url": "http://localhost:9411/api/v2/spans"
-                }
-              ]
-            }
-          }
-        }
-      }
+      "$ref": "ory://tracing-config"
```
aeneasr, line 1025: @alnr is this correct?

alnr: This is fine for now. I will follow up with a PR to render the config schema into a self-contained file like we do in keto and kratos.

@sawadashota I think you may be missing a `go mod tidy` in this PR :)

alnr (approved):
Thanks @sawadashota

PR #3446, opened 2023-02-22
hperl/per-client-token-types

hperl:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


aeneasr:
For the tests, please check that `jwt->jwt`, `jwt->opaque`, `opaque->jwt`, `opaque->opaque` and flows:

* client credentials: issue token, validate token, invalidate token
* auth code: issue token, validate token, refresh token, validate token, invalidate token, call userinfo endpoint
* jwt-bearer flow: does it issue the token in the correct format?

driver/config/provider.go
```diff
@@ -186,18 +186,10 @@ func (p *DefaultProvider) IsDevelopmentMode(ctx context.Context) bool {
 }
 
 func (p *DefaultProvider) WellKnownKeys(ctx context.Context, include ...string) []string {
-	if p.AccessTokenStrategy(ctx) == AccessTokenJWTStrategy {
-		include = append(include, x.OAuth2JWTKeyName)
-	}
-
-	include = append(include, x.OpenIDConnectKeyName)
+	include = append(include, x.OAuth2JWTKeyName, x.OpenIDConnectKeyName)
```
hperl, line 189: The `.well-known` endpoint does not depend on a specific client, so we don't know whether or not we need to also list the JWT key names here.

@aeneasr I see no harm in just always including the JWT key names. Correct?

jwk/handler_test.go
```diff
@@ -50,7 +50,7 @@ func TestHandlerWellKnown(t *testing.T) {
 		err = json.NewDecoder(res.Body).Decode(&known)
 		require.NoError(t, err, "problem in decoding response")
 
-		require.Len(t, known.Keys, 1)
+		require.GreaterOrEqual(t, len(known.Keys), 1)
```
hperl, line 53: There are now both the OIDC and JWT keys present, which is why we need to check `>=` here.

oauth2/fosite_store_helpers.go
```diff
@@ -491,14 +492,17 @@ func testHelperFlushTokensWithLimitAndBatchSize(x InternalRegistry, limit int, b
...
+			if _, err := m.GetAccessTokenSession(ctx, requests[i].ID, ds); err == nil {
+				foundCount++
 			} else {
-				require.Error(t, err)
+				require.ErrorIs(t, err, fosite.ErrNotFound)
+				notFoundCount++
 			}
 		}
+		assert.Equal(t, limit, notFoundCount, "should have deleted %d tokens", limit)
+		assert.Equal(t, totalCount-limit, foundCount, "should have found %d tokens", totalCount-limit)
```
hperl, line 505: Because we delete the access keys up to `limit` based on the signature, and we now always hash the signature, this will give us a random ordering of deleted vs non-deleted tokens.

cypress/integration/oauth2/authorize_code.js
```diff
@@ -3,91 +3,98 @@
 
 import { prng } from "../../helpers"
 
+const accessTokenStrategies = ["opaque", "jwt"]
```
hperl, line 6: All OIDC and JWT suites now execute with an `opaque` and a `jwt` client. Since we start the suite twice, once with and without `--jwt`, we get the whole matrix tested between global and client access token strategy.

persistence/sql/persister_oauth2.go
```diff
@@ -161,7 +161,7 @@ func (r *OAuth2RequestSQL) toRequest(ctx context.Context, session fosite.Session
 
 // hashSignature prevents errors where the signature is longer than 128 characters (and thus doesn't fit into the pk).
 func (p *Persister) hashSignature(ctx context.Context, signature string, table tableName) string {
-	if table == sqlTableAccess && p.config.IsUsingJWTAsAccessTokens(ctx) {
+	if table == sqlTableAccess {
```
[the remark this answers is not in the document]
aeneasr: Because that would break all existing tokens, and because that's not the reason why we hash:

>  // hashSignature prevents errors where the signature is longer than 128 characters (and thus doesn't fit into the pk).

This is only the case for JWT signatures which are typically > 128 chars.

I understand though why this is here, but we need to figure out either how to get the strategy in here, or how to solve this backwards compatible.

cypress/integration/oauth2/authorize_code.js
aeneasr: 🧠

jwk/handler_test.go
aeneasr: Thank you, very helpful

driver/config/provider.go
aeneasr: Yeah, it's fine!

aeneasr (changes requested):
Please make this not available in OIDC dynamic client registration.

persistence/sql/persister_oauth2.go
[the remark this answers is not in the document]
hperl: fixed

aeneasr (dismissed):
Great job, looks good!

aeneasr:
merge conflicts 🙄

hperl:
@aeneasr all resolved now :)

Issue #3448, opened 2023-02-23
Add support for login session revocation by session ID


sgal:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/hydra/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/hydra/blob/master/CONTRIBUTING.md).
- [X] This issue affects my [Ory Network](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

We are building an OIDC login solution with Hydra and want to have a Single Sign Out as the only sign-out strategy. This means that if the user is signing out of any of our apps, they should be signed out from all of the apps we have.

While this is generally supported via RP Initiated Logout in Hydra, the UX of it is quite bad (especially in the native apps). The reason for the bad UX is the non-customizable consent prompt on iOS (and some Android phones as well).
The dialog looks like this
![consent](https://user-images.githubusercontent.com/2971261/220867074-123d7243-7b5a-4759-8982-2a698131dd30.png)

We want to avoid the prompt altogether to make the sign-out experience quick and predictable for our users.

### Describe your ideal solution

The ideal solution for us would be a headless/admin API to revoke the login session by the session ID (`sid` claim in the ID token).

Similar to the existing API that revokes all login sessions for a given subject - https://www.ory.sh/docs/reference/api#tag/oAuth2/operation/revokeOAuth2LoginSessions

We also can use an `id_token_hint` to get backchannel logout notifications and let the registered clients know that logout was performed. If this part is difficult, it can be skipped, since we can implement a workaround on our end for the notifications.

### Workarounds or alternatives

Since there is no easy way to achieve a headless log-out now, we would have to connect directly to the DB that Hydra uses and invalidate the sessions ourselves. This is extremely hacky and would like to result in data corruption and failed login sessions.

### Version

2.x

PR #3450, opened 2023-02-26
feat: add ability to revoke login sessions by SessionID

sgal:
Added support to `revokeOAuth2LoginSessions` API to revoke a single session by a SessionID (`sid` claim in the id_token) and execute an OpenID Connect Back-channel logout.

## Related issue(s)
https://github.com/ory/hydra/issues/3448

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/consent/handler.go
+++ b/consent/handler.go
@@ -230,7 +230,6 @@ type revokeOAuth2LoginSessions struct {
 	// The subject to revoke authentication sessions for.
 	//
 	// in: query
-	// required: true
 	Subject string `json:"subject"`
 
 	// OAuth 2.0 Subject
--- a/consent/handler.go
+++ b/consent/handler.go
@@ -149,6 +149,7 @@ type listOAuth2ConsentSessions struct {
 	// in: query
 	// required: true
 	Subject string `json:"subject"`
+
 	// The login session id to list the consent sessions for.
 	//
 	// in: query
@@ -231,15 +232,27 @@ type revokeOAuth2LoginSessions struct {
 	// in: query
 	// required: true
 	Subject string `json:"subject"`
+
+	// OAuth 2.0 Subject
+	//
+	// The subject to revoke authentication sessions for.
+	//
+	// in: query
+	SessionID string `json:"sid"`
 }
 
 // swagger:route DELETE /admin/oauth2/auth/sessions/login oAuth2 revokeOAuth2LoginSessions
 //
-// # Revokes All OAuth 2.0 Login Sessions of a Subject
+// # Revokes OAuth 2.0 Login Sessions by either a Subject or a SessionID
 //
-// This endpoint invalidates a subject's authentication session. After revoking the authentication session, the subject
-// has to re-authenticate at the Ory OAuth2 Provider. This endpoint does not invalidate any tokens and
-// does not work with OpenID Connect Front- or Back-channel logout.
+// This endpoint invalidates authentication sessions. After revoking the authentication session(s), the subject
+// has to re-authenticate at the Ory OAuth2 Provider. This endpoint does not invalidate any tokens.
+//
+// If you send the subject in a query param, all authentication sessions that belong to that subject are revoked.
+// No OpennID Connect Front- or Back-channel logout is performed in this case.
+//
+// Alternatively, you can send a SessionID via `sid` query param, in which case, only the session that is connected
+// to that SessionID is revoked. OpenID Connect Back-channel logout is performed in this case.
 //
 //	Consumes:
 //	- application/json
@@ -253,6 +266,17 @@ type revokeOAuth2LoginSessions struct {
 //	  204: emptyResponse
 //	  default: errorOAuth2
 func (h *Handler) revokeOAuth2LoginSessions(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	sid := r.URL.Query().Get("sid")
+	if sid != "" {
+		if err := h.r.ConsentStrategy().HandleHeadlessLogout(r.Context(), w, r, sid); err != nil {
+			h.r.Writer().WriteError(w, r, err)
+			return
+		}
+
+		w.WriteHeader(http.StatusNoContent)
+		return
+	}
+
 	subject := r.URL.Query().Get("subject")
 	if subject == "" {
 		h.r.Writer().WriteError(w, r, errorsx.WithStack(fosite.ErrInvalidRequest.WithHint(`Query parameter 'subject' is not defined but should have been.`)))
--- a/consent/handler.go
+++ b/consent/handler.go
@@ -266,6 +266,13 @@ type revokeOAuth2LoginSessions struct {
 //	  default: errorOAuth2
 func (h *Handler) revokeOAuth2LoginSessions(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
 	sid := r.URL.Query().Get("sid")
+	subject := r.URL.Query().Get("subject")
+
+	if sid == "" && subject == "" {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(fosite.ErrInvalidRequest.WithHint(`Either 'subject' or 'sid' query parameters need to be defined.`)))
+		return
+	}
+
 	if sid != "" {
 		if err := h.r.ConsentStrategy().HandleHeadlessLogout(r.Context(), w, r, sid); err != nil {
 			h.r.Writer().WriteError(w, r, err)
@@ -273,12 +280,6 @@ func (h *Handler) revokeOAuth2LoginSessions(w http.ResponseWriter, r *http.Reque
 		}
 
 		w.WriteHeader(http.StatusNoContent)
-		return
-	}
-
-	subject := r.URL.Query().Get("subject")
-	if subject == "" {
-		h.r.Writer().WriteError(w, r, errorsx.WithStack(fosite.ErrInvalidRequest.WithHint(`Query parameter 'subject' is not defined but should have been.`)))
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


consent/handler.go
```diff
@@ -231,15 +232,27 @@ type revokeOAuth2LoginSessions struct {
 	// in: query
 	// required: true
```
sgal, line 233: Not sure how to make the API accept either `subject` or `sid` without introducing a breaking change to this schema.

consent/handler.go
```diff
@@ -231,15 +232,27 @@ type revokeOAuth2LoginSessions struct {
...
+// # Revokes OAuth 2.0 Login Sessions by either a Subject or a SessionID
+//
+// This endpoint invalidates authentication sessions. After revoking the authentication session(s), the subject
+// has to re-authenticate at the Ory OAuth2 Provider. This endpoint does not invalidate any tokens.
 //
-// This endpoint invalidates a subject's authentication session. After revoking the authentication session, the subject
-// has to re-authenticate at the Ory OAuth2 Provider. This endpoint does not invalidate any tokens and
-// does not work with OpenID Connect Front- or Back-channel logout.
+// If you send the subject in a query param, all authentication sessions that belong to that subject are revoked.
+// No OpennID Connect Front- or Back-channel logout is performed in this case.
```
sgal, line 252: Looks like this can be fixed as well, maybe outside of this change.

consent/strategy.go
```diff
@@ -15,5 +15,6 @@ var _ Strategy = new(DefaultStrategy)
 type Strategy interface {
 	HandleOAuth2AuthorizationRequest(ctx context.Context, w http.ResponseWriter, r *http.Request, req fosite.AuthorizeRequester) (*AcceptOAuth2ConsentRequest, error)
 	HandleOpenIDConnectLogout(ctx context.Context, w http.ResponseWriter, r *http.Request) (*LogoutResult, error)
+	HandleHeadlessLogout(ctx context.Context, w http.ResponseWriter, r *http.Request, sid string) error
```
sgal, line 18: Best name I could come up with, feedback/alternative naming is very much welcome.

consent/strategy_default.go
```diff
@@ -983,6 +983,37 @@ func (s *DefaultStrategy) HandleOpenIDConnectLogout(ctx context.Context, w http.
...
 
+func (s *DefaultStrategy) HandleHeadlessLogout(ctx context.Context, w http.ResponseWriter, r *http.Request, sid string) error {
+	loginSession, lsErr := s.r.ConsentManager().GetRememberedLoginSession(ctx, sid)
+
+	// This is ok (session probably already revoked), do nothing!
+	if lsErr != nil {
+		return nil
+	}
+
+	if err := s.executeBackChannelLogout(r.Context(), r, loginSession.Subject, sid); err != nil {
```
sgal, line 994: Everything from here until the end of the function is the same as in `completeLogout` function. Is this the correct direction to take here?

consent/handler.go
aeneasr: Hm good question, I think in that case it's ok to break the schema here and we'll release it as a minor patch

consent/strategy_default.go
```diff
@@ -983,6 +983,37 @@ func (s *DefaultStrategy) HandleOpenIDConnectLogout(ctx context.Context, w http.
...
+	if err := s.r.ConsentManager().DeleteLoginSession(r.Context(), sid); errors.Is(err, sqlcon.ErrNoRows) {
+		// This is ok (session probably already revoked), do nothing!
+	} else if err != nil {
+		return err
+	}
+
+	s.r.AuditLogger().
+		WithRequest(r).
+		WithField("subject", loginSession.Subject).
+		Info("User logout completed!")
```
aeneasr, line 1012: Clarify that this is the headless flow

consent/strategy_default.go
aeneasr: That looks good to me! Since this is called via an API, we won't have front-channel log-out (because no browser is involved). I think this is fine, but should probably be clarified in the API / function doc. You may opt to create a new function with the flow as to avoid duplication, but it's not required since it's only 2 functional statements

aeneasr (commented):
Looks quite good already! We'll a few tests for this but then this can ship

consent/strategy_logout_test.go
```diff
@@ -99,6 +99,23 @@ func TestLogoutFlows(t *testing.T) {
 		return string(ioutilx.MustReadAll(resp.Body)), resp
 	}
 
+	makeHeadlessLogoutRequest := func(t *testing.T, hc *http.Client, values url.Values) (body string, resp *http.Response) {
+		var err error
+		req, err := http.NewRequest(http.MethodDelete, adminTS.URL+"/admin/oauth2/auth/sessions/login?"+values.Encode(), nil)
```
sgal, line 104: Not sure why, but adding `/admin` prefix here solved the tcp connect issue.

consent/strategy_logout_test.go
```diff
@@ -490,4 +507,20 @@ func TestLogoutFlows(t *testing.T) {
...
 		wg.Wait()
 	})
+
+	t.Run("case=should execute backchannel logout in headless flow with sid", func(t *testing.T) {
+		sid := make(chan string)
+		acceptLoginAsAndWatchSid(t, subject, sid)
+
+		backChannelWG := newWg(1)
+		c := createClientWithBackchannelLogout(t, backChannelWG, func(t *testing.T, logoutToken gjson.Result) {
+			assert.EqualValues(t, <-sid, logoutToken.Get("sid").String(), logoutToken.Raw)
```
sgal, line 517: Here is where my golang channels knowledge is clearly not sufficient.

This test hangs because channel var `sid` is being read twice(L517 and L522) while being written once.

Not sure how to make it work with using the value 2 times here (once for calling the headless logout and once for verifying the `sid` claim in the logout token)

consent/strategy_logout_test.go
aeneasr: 👍

consent/strategy_logout_test.go
aeneasr: What is it that you're trying to achieve?

consent/strategy.go
aeneasr: Sounds good to me :)

consent/strategy_logout_test.go
sgal: I need `sid` value in 2 places - to pass in the headless logout request as a query param, and in the L517 where we verify the value inside of the `logout_token`.

consent/strategy_logout_test.go
```diff
@@ -164,7 +181,9 @@ func TestLogoutFlows(t *testing.T) {
 				require.NoError(t, err)
 				if sid != nil {
 					go func() {
-						sid <- *res.LoginSessionId
+						for i := 0; i < numSidConsumers; i++ {
```
sgal, line 184: Not sure this is the best way to solve the issue of reading from `sid` channel 2 times in the headless logout tests, but this fixes the issue. What do you think @aeneasr?

consent/strategy_logout_test.go
```diff
@@ -238,7 +257,7 @@ func TestLogoutFlows(t *testing.T) {
 
 	t.Run("case=should execute backchannel logout if issued without rp-involvement", func(t *testing.T) {
 		sid := make(chan string)
-		acceptLoginAsAndWatchSid(t, subject, sid)
+		acceptLoginAsAndWatchSid(t, subject, sid, 1)
```
sgal, line 260: Honestly, not a big fan of this approach, because it is confusing. AFAIK, golang does not have default argument values to skip this change.

consent/strategy_logout_test.go
```diff
@@ -490,4 +509,24 @@ func TestLogoutFlows(t *testing.T) {
 
 		wg.Wait()
 	})
+
+	t.Run("case=should execute backchannel logout in headless flow with sid", func(t *testing.T) {
+		sid := make(chan string, 2)
```
sgal, line 514: Using a buffered channel here to make sure that we have both values in it and do not block the goroutine that tries to read it.

consent/strategy_logout_test.go
sgal: Fixed with a wrapper function to encapsulate this logic for most tests

consent/strategy_default.go
```diff
@@ -983,6 +991,27 @@ func (s *DefaultStrategy) HandleOpenIDConnectLogout(ctx context.Context, w http.
 	return s.completeLogout(ctx, w, r)
 }
 
+func (s *DefaultStrategy) HandleHeadlessLogout(ctx context.Context, w http.ResponseWriter, r *http.Request, sid string) error {
+	loginSession, lsErr := s.r.ConsentManager().GetRememberedLoginSession(ctx, sid)
+
+	// This is ok (session probably already revoked), do nothing!
+	if lsErr != nil {
```
hperl, line 998: Should we test if `errors.Is(err, x.ErrNotFound)`?

hperl: Also, should we not still perform the backchannel logout in this case?

hperl (changes requested):
Thanks, this looks very good! I have some small comments and then it's ready to merge :)

consent/strategy_default.go
sgal: 👍 for the specific error check.

For backchannel logout, we need a subject, which is stored in the Login Session. If we cannot find the session, then there is no subject as well, and we risk breaking the OIDC compliance by skipping the subject claim.

sgal:
@hperl Fixed the comments, please have a look.

consent/handler.go
hperl: As it is implemented now there is no breaking change, because we remove required, and add a new parameter. So old clients will continue to work.

hperl (approved):
Thanks for the changes, LGTM! :tada:

consent/strategy_logout_test.go
aeneasr: There probably is some go magic possible but I think this is fine :)

aeneasr (approved):

kabirbaidhya:
Hey @aeneasr, when are you planning to release this change?

PR #3451, opened 2023-02-27
feat: allow skipping the consent flow

hperl:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


consent/handler.go
```diff
@@ -588,6 +588,10 @@ func (h *Handler) getOAuth2ConsentRequest(w http.ResponseWriter, r *http.Request
 		request.RequestedAudience = []string{}
 	}
 
+	if request.Client.SkipConsent {
+		request.Skip = true
```
aeneasr, line 592: This typically means that there is an existing, previous consent for that user & app. In this case though it says: just skip this!

I'm not 100% sure but it's possible that `Skip==true` has other implications (e.g. remember can not also be true?).

client/handler_test.go
```diff
@@ -328,6 +328,15 @@ func TestHandler(t *testing.T) {
...
 					statusCode: http.StatusBadRequest,
 				},
+				{
+					d: "setting skip_consent fails",
+					payload: &client.Client{
+						RedirectURIs: []string{"http://localhost:3000/cb"},
+						SkipConsent:  true,
+					},
+					path:       client.DynClientsHandlerPath,
+					statusCode: http.StatusBadRequest,
```
aeneasr, line 338: We test that we can't set it, but how do we test that we can set it?

aeneasr (commented):
Nice! Some nits

client/handler_test.go
hperl: In the E2E tests setting is tested.

consent/handler.go
kmherrmann: i've been thinking about this too - should we actually record an auto-granted consent rather than skipping it entirely (assuming i'm reading this correctly)?
otoh, it's a per-client setting, so remember/not remember has no real effect as a new consent would also just be granted. Only if the users turn off skipConsent later, this matters, and in that case i'd argue it's actually better to not consider the auto-granted consents in the future.

hperl: What the consent app does depends on it's implementation, but both hydra-login-consent-node (https://github.com/ory/hydra-login-consent-node/blob/master/src/routes/consent.ts#L39) as well as the consent we use in the Ory Network return early by accepting the consent with all requested scopes on behalf of the user.

I agree that the semantics are different between "skip this for this user because consent was already given" and "skip this for all users because consent is not necessary", but the result and the code path is identical.

hperl: Regarding analytics, we can surely emit events when a consent was skipped.

client/handler_test.go
hperl: Here: https://github.com/ory/hydra/pull/3451/files#diff-62406830adc9cbc1fec4e4768b0debd0379361f10a6786ce4b933390bcf5523dR96

consent/handler.go
aeneasr: Ok I see. In that case, it probably makes sense to maybe just add this logic to the consent app where it checks for the client field? Then, I think the semantics would be cleaner and the code is minimal :)

client/handler_test.go
aeneasr: Maybe one explicit test? :)

consent/handler.go
hperl: We can do that! I added the check in https://github.com/ory/hydra-login-consent-node/pull/137 (and will add in Cloud).

I still want to check that the consent is skipped in Hydra E2E, so we need to merge the logic in the consent Node app first.

hperl: Done, and tested with hydra-consent-node

client/handler_test.go
hperl: done.

kmherrmann (approved):
Amazing, thanks!
My review is obviously superficial wrt. to implementation details

client/client.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/gobuffalo/pop/v6"
 	"github.com/gofrs/uuid"
 
-	jose "gopkg.in/square/go-jose.v2" // Naming the dependency jose is important for go-swagger to work, see https://github.com/go-swagger/go-swagger/issues/1587
+	"gopkg.in/square/go-jose.v2" // Naming the dependency jose is important for go-swagger to work, see https://github.com/go-swagger/go-swagger/issues/1587
```
alnr, line 15: This change seems to conflict with the comment

internal/httpclient/model_o_auth2_client.go
```diff
@@ -1368,6 +1370,38 @@ func (o *OAuth2Client) SetSectorIdentifierUri(v string) {
 	o.SectorIdentifierUri = &v
 }
 
+// GetSkipConsent returns the SkipConsent field value if set, zero value otherwise.
+func (o *OAuth2Client) GetSkipConsent() bool {
+	if o == nil || o.SkipConsent == nil {
+		var ret bool
+		return ret
```
alnr, line 1377:
```suggestion
		return false
```

alnr (changes requested):
tiny nits

client/client.go
hperl: Yea IDK, my IDE (Goland) does this automatically. I think this is not needed anyways (because the package is named `jose` already), but we can remove the comment. OK?

aeneasr (approved):

bulolo:
how to use skip_consent?

kmherrmann:
Please see here: https://www.ory.sh/docs/oauth2-oidc/custom-login-consent/flow#skipping-consent-for-trusted-clients

PR #3502, opened 2023-04-26
fix: proper SameSite=None in dev mode

hperl:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/driver/config/provider.go
+++ b/driver/config/provider.go
@@ -277,7 +277,8 @@ func (p *DefaultProvider) CookieSameSiteMode(ctx context.Context) http.SameSite 
 	case "strict":
 		return http.SameSiteStrictMode
 	case "none":
-		if p.IsDevelopmentMode(ctx) {
+		if p.IssuerURL(ctx).Scheme != "https" {
+			// SameSite=None can only be set for HTTPS issuers.
 			return http.SameSiteLaxMode
 		}
 		return http.SameSiteNoneMode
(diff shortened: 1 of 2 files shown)


zepatrik (commented):
This could be a possible explanation: https://github.com/ory/hydra/commit/921f8c23d507ed84ed36792dfa343f235da2e501
Do we automatically set the cookies to not secure when in dev mode? In that case we need to change that as well, so they are always secure. Then `none` should work.

zepatrik (approved):

zepatrik (commented):
As discussed, we cannot assume an instance in dev mode is not using TLS. It is also not possible to tell in the service itself because the TLS connection might be terminated in a side-car or on the ingress level.

driver/config/provider.go
```diff
@@ -277,9 +277,6 @@ func (p *DefaultProvider) CookieSameSiteMode(ctx context.Context) http.SameSite
 	case "strict":
 		return http.SameSiteStrictMode
 	case "none":
-		if p.IsDevelopmentMode(ctx) {
```
aeneasr, line 280: Please revert this - but you can add that:

1. If TLS is disabled and dev mode is on, (you can check this via the issuer or public URL for example) then set this to lax
2. If TLS is enabled then return "None"

hperl: This won't work for Ory network, as TLS is always terminated before reaching Hydra there.

aeneasr: We can identify whether TLS is enabled by looking at the public or issuer URL. If that is prefixed https, we're serving TLS.

hperl: pure genius :)

zepatrik (approved):

aeneasr (approved):

PR #3506, opened 2023-05-02
chore: bump deps

alnr:
This should speed up INSERTs a good bit by bumping the pop dependency. Currently, we're creating a prepared statement for each insert, which is discarded immediately afterwards.

https://github.com/gobuffalo/pop/pull/807


kmherrmann:
Great!
Any indication on how much we'll gain from this?

alnr:
> Great! Any indication on how much we'll gain from this?

Not right now. Once we have benchmarks we can compare to the status quo.

PR #3539, opened 2023-06-06
feat: migrate image to distroless

Demonsthere:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

## Further Comments
- ~~Downgrade alpine image to 3.17 due to [failing package upgrade](https://github.com/ory/hydra/actions/runs/5188782643/jobs/9352884594#step:8:110)~~
- Migrate to debian builder and distroless:base image to avoid CVE issues in base containers. We cannot migrate to static as hydra is build with dynamic linking.


.docker/Dockerfile-build
```diff
@@ -17,28 +18,14 @@ COPY . .
 
 RUN go build -tags sqlite,json1 -o /usr/bin/hydra
 
-FROM alpine:3.18
+#########################
 
-RUN addgroup -S ory; \
-    adduser -S ory -G ory -D  -h /home/ory -s /bin/nologin; \
-    chown -R ory:ory /home/ory
+FROM gcr.io/distroless/static-debian11:nonroot as runner
```
zepatrik, line 23: Did we check whether this has a compatible glibc, i.e. works with SQLite? Note that here we don't use musl, so the binary has to be linked against glibc.

zepatrik (commented):
If it works with sqlite :+1:

.docker/Dockerfile-build
hperl: From what I remember you have to set `CGO_ENABLED 0` for Go to work in non-libc environments.

.docker/Dockerfile-build
```diff
@@ -1,44 +1,28 @@
...
 ENV GO111MODULE on
 ENV CGO_ENABLED 1
 
-RUN go mod download
-
 COPY . .
+RUN go mod download &&\
+    go build -tags sqlite,json1 -o /usr/bin/hydra
 
-RUN go build -tags sqlite,json1 -o /usr/bin/hydra
```
aeneasr, line 18: Please keep these instructions as they were - they prevent excessive downloading of go modules. Thanks!

aeneasr (dismissed):

zepatrik (approved):

zepatrik (approved):

PR #3575, opened 2023-07-17
feat: add support for OIDC VC

hperl:
## Related issue(s)

Requires https://github.com/ory/fosite/pull/758

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


oauth2/oauth2_auth_code_test.go
```diff
@@ -347,9 +334,102 @@ func TestAuthCodeWithDefaultStrategy(t *testing.T) {
...
+			verifiableCredential, err := createVerifiableCredential(t, reg, token, &hydraoauth2.CreateVerifiableCredentialRequestBody{
+				Format: "jwt_vc_json",
+				Type:   []string{"VerifiableCredential", "UserInfoCredential"},
+				Proof: &hydraoauth2.VerifiableCredentialProof{
+					ProofType: "jwt",
+					JWT:       proofJWT,
+				},
+			})
+			require.NoError(t, err)
+			assertVerifiableCredentialContainsPublicKey(t, reg, verifiableCredential, pubKeyJWK)
```
hperl, line 413: TODO: 

* Also test for various error conditions
* Test priming request

oauth2/handler.go
```diff
@@ -1154,3 +1181,165 @@ func (h *Handler) logOrAudit(err error, r *http.Request) {
...
+//	Consumes:
+//	- application/json
+//
+//	Schemes: http, https
+//
+//	Responses:
+//	  200: verifiableCredentialResponse
+//	  400: verifiableCredentialPrimingResponse
+//	  default: errorOAuth2
+func (h *Handler) createVerifiableCredential(w http.ResponseWriter, r *http.Request) {
```
hperl, line 1200: TODO: Extract some functionality, test individually:

* Proof verification
* VC issuance (from session)

persistence/sql/persister_nonce_test.go
```diff
@@ -0,0 +1,4 @@
+// Copyright © 2023 Ory Corp
+// SPDX-License-Identifier: Apache-2.0
+
+package sql
```
hperl, line 4: TODO: Tests for the nonce persister.

oauth2/handler.go
```diff
@@ -1154,3 +1207,170 @@ func (h *Handler) logOrAudit(err error, r *http.Request) {
 		x.LogAudit(r, err, h.r.Logger())
 	}
 }
+
+// swagger:route POST /credentials oAuth2 createVerifiableCredential
+//
+// # Delete OAuth 2.0 Access Tokens from specific OAuth 2.0 Client
```
hperl, line 1213: copy paste

oauth2/handler.go
```diff
@@ -1154,3 +1207,175 @@ func (h *Handler) logOrAudit(err error, r *http.Request) {
...
+type VerifiableCredentialResponse struct {
+	Format     string `json:"format"`
+	Credential string `json:"credential_draft_00"`
+}
+
+// swagger:model verifiableCredentialPrimingResponse
+type VerifiableCredentialPrimingResponse struct {
+	Format          string `json:"format"`
+	Nonce           string `json:"c_nonce"`
+	NonceExperiesIn int64  `json:"c_nonce_expires_in"`
```
alnr, line 1378: Typo

alnr, line 1378:
```suggestion
	NonceExpiresIn int64  `json:"c_nonce_expires_in"`
```

oauth2/handler.go
```diff
@@ -1154,3 +1207,175 @@ func (h *Handler) logOrAudit(err error, r *http.Request) {
...
+type createVerifiableCredentialRequest struct {
+	// in: body
+	Body CreateVerifiableCredentialRequestBody
+}
+
+// CreateVerifiableCredentialRequestBody The request body to request a verifiable credential.
+type CreateVerifiableCredentialRequestBody struct {
+	Format string   `json:"format"`
+	Types  []string `json:"types"`
+	Proof  *VerifiableCredentialProof
```
alnr, line 1360:
```suggestion
	Proof  *VerifiableCredentialProof `json:"proof"`
```

persistence/sql/persister_nonce.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+func accessTokenHash(accessToken string) []byte {
+	return sha256.New().Sum([]byte(accessToken))
+}
+
+// Set the purpose to something unique to avoid ciphertext confusion with other usages of the AEAD cipher.
+var purpose = []byte("nonce")
+
+func (p *Persister) NewNonce(ctx context.Context, accessToken string, expiresIn time.Time) (res string, err error) {
```
alnr, line 28: I feel that generating this nonce (while a great idea in principle) could be made simpler and potentially more secure by not hashing and comparing the access token. 

Here's my take:

```go
// Copyright © 2023 Ory Corp
// SPDX-License-Identifier: Apache-2.0

package sql

import (
	"context"
	"fmt"
	"strconv"
	"time"

	"github.com/ory/fosite"
	"github.com/ory/x/errorsx"
	"github.com/ory/x/otelx"
)

func (p *Persister) NewNonce(ctx context.Context, accessToken string, expiresIn time.Time) (res string, err error) {
	ctx, span := p.r.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.NewNonce")
	defer otelx.End(span, &err)

	plaintext := fmt.Sprintf("%d", expiresIn.Unix())
	aad := []byte("vc-nonce-at:" + accessToken)

	return p.r.FlowCipher().Encrypt(ctx, []byte(plaintext), aad)
}

func (p *Persister) IsNonceValid(ctx context.Context, accessToken, nonce string) (err error) {
	ctx, span := p.r.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.IsNonceValid")
	defer otelx.End(span, &err)

	aad := []byte("vc-nonce-at:" + accessToken)
	plaintext, err := p.r.FlowCipher().Decrypt(ctx, nonce, aad)
	if err != nil {
		return errorsx.WithStack(fosite.ErrInvalidRequest.WithHintf("Invalid nonce."))
	}

	exp, err := strconv.ParseInt(string(plaintext), 10, 64)
	if err != nil {
		return errorsx.WithStack(fosite.ErrInvalidRequest.WithHintf("Invalid nonce.")) // should never happen
	}

	if exp < time.Now().Unix() {
		return errorsx.WithStack(fosite.ErrInvalidRequest.WithHintf("The nonce has expired."))
	}

	return nil
}
```

The protobuf solution is admittedly more easily extendable in the future.

oauth2/oauth2_auth_code_test.go
```diff
@@ -977,6 +1150,103 @@ func TestAuthCodeWithDefaultStrategy(t *testing.T) {
...
+	expectedID := fmt.Sprintf("did:jwk:%s", base64.RawURLEncoding.EncodeToString(pubKeyRaw))
+	require.Equal(t, expectedID, token.Claims.(jwt.MapClaims)["vc"].(map[string]any)["credentialSubject"].(map[string]any)["id"])
+}
+
+func createVerifiableCredential(
+	t *testing.T,
+	reg driver.Registry,
+	token *oauth2.Token,
+	createVerifiableCredentialReq *hydraoauth2.CreateVerifiableCredentialRequestBody,
+) (vcRes *hydraoauth2.VerifiableCredentialResponse, vcErr *fosite.RFC6749Error, err error) {
```
alnr, line 1190: This signature is kinda wonky. We never assign to `vcErr` in the body of the function.

The idea is to return either a successful VerifiableCredentialResponse, or an RFC6749Error, correct?

In that case, we should remove the third`err error` return value and rely on `require.NoError` inside the function.

alnr (changes requested):
Very nicely done!

persistence/sql/persister_nonce.go
hperl: My reason on why to encrypt the access token was that it is unexpected that the nonce contains the access token in plain text. Clients might pass the nonce around or store them less securely than they would the access token.

The reason for protobuf over a hand-rolled format was so that we can extend what goes into the nonce without too much headache.

oauth2/oauth2_auth_code_test.go
hperl: Thanks, that'd work!

internal/httpclient/api/openapi.yaml
```diff
@@ -1365,6 +1365,39 @@ paths:
       summary: Get Trusted OAuth2 JWT Bearer Grant Type Issuer
       tags:
       - oAuth2
+  /credentials:
+    post:
+      description: This endpoint deletes OAuth2 access tokens issued to an OAuth 2.0
```
aeneasr, line 1370: This description does not look correct

internal/httpclient/api/openapi.yaml
```diff
@@ -1365,6 +1365,39 @@ paths:
...
+              schema:
+                $ref: '#/components/schemas/verifiableCredentialPrimingResponse'
+          description: verifiableCredentialPrimingResponse
+        default:
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/errorOAuth2'
+          description: errorOAuth2
+      summary: Delete OAuth 2.0 Access Tokens from specific OAuth 2.0 Client
```
aeneasr, line 1398: This description does not look correct

internal/httpclient/api/openapi.yaml
```diff
@@ -1365,6 +1365,39 @@ paths:
...
+      tags:
+      - oAuth2
```
aeneasr, line 1400: Shouldn't this be the oidc tag?

oauth2/handler.go
```diff
@@ -404,6 +412,38 @@ type oidcConfiguration struct {
...
+	// Contains the URL of the Verifiable Credentials Endpoint.
+	CredentialsEndpointDraft00 string `json:"credentials_endpoint_draft_00"`
+
+	// OpenID Connect Verifiable Credentials Supported
+	//
+	// JSON array containing a list of the Verifiable Credentials supported by this authorization server.
+	CredentialsSupportedDraft00 []CredentialSupportedDraft00 `json:"credentials_supported_draft_00"`
+}
+
+type CredentialSupportedDraft00 struct {
```
aeneasr, line 427: Please add the correct swagger annotations (including `swagger:model`)

oauth2/handler.go
```diff
@@ -1154,3 +1207,175 @@ func (h *Handler) logOrAudit(err error, r *http.Request) {
 		x.LogAudit(r, err, h.r.Logger())
 	}
 }
+
+// swagger:route POST /credentials oAuth2 createVerifiableCredential
```
aeneasr, line 1211: Consider moving this to the oidc tag (i believe we have such a tag if i am not mistaken)

oauth2/handler.go
```diff
@@ -1154,3 +1207,175 @@ func (h *Handler) logOrAudit(err error, r *http.Request) {
...
+	session := NewSessionWithCustomClaims("", h.c.AllowedTopLevelClaims(ctx))
+	accessToken := fosite.AccessTokenFromRequest(r)
+	tokenType, _, err := h.r.OAuth2Provider().IntrospectToken(ctx, accessToken, fosite.AccessToken, session)
+
+	if err != nil {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(err))
+		return
+	}
+	if tokenType != fosite.AccessToken {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(fosite.ErrInvalidRequest.WithHintf("the token is not an access token")))
```
aeneasr, line 1242: Hints are regular sentences and have proper punctuation:

`The provided token is not an access token.`

oauth2/handler.go
```diff
@@ -1154,3 +1207,175 @@ func (h *Handler) logOrAudit(err error, r *http.Request) {
...
+		return
+	}
+
+	var request CreateVerifiableCredentialRequestBody
+	if err := json.NewDecoder(r.Body).Decode(&request); err != nil {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(err))
```
aeneasr, line 1248: Will return a 500 error if this can not be decoded

oauth2/handler.go
```diff
@@ -1154,3 +1207,175 @@ func (h *Handler) logOrAudit(err error, r *http.Request) {
...
+//	  400: verifiableCredentialPrimingResponse
+//	  default: errorOAuth2
+func (h *Handler) createVerifiableCredential(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	session := NewSessionWithCustomClaims("", h.c.AllowedTopLevelClaims(ctx))
+	accessToken := fosite.AccessTokenFromRequest(r)
+	tokenType, _, err := h.r.OAuth2Provider().IntrospectToken(ctx, accessToken, fosite.AccessToken, session)
+
+	if err != nil {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(err))
```
aeneasr, line 1238: I don't think you need to WIthStack this, as it comes from a component already that wraps the error with a stack trace. Overriding this here means that we lose the stack trace.

oauth2/handler.go
```diff
@@ -1154,3 +1207,175 @@ func (h *Handler) logOrAudit(err error, r *http.Request) {
...
+	}
+
+	var request CreateVerifiableCredentialRequestBody
+	if err := json.NewDecoder(r.Body).Decode(&request); err != nil {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(err))
+		return
+	}
+
+	if request.Format != "jwt_vc_json" {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(fosite.ErrInvalidRequest.WithHintf("the format is not supported")))
```
aeneasr, line 1253: Make this a proper sentence

oauth2/handler.go
```diff
@@ -1154,3 +1207,175 @@ func (h *Handler) logOrAudit(err error, r *http.Request) {
...
+		return
+	}
+	if request.Proof == nil {
+		// Handle priming request
+		nonceLifespan := h.r.Config().GetVerifiableCredentialsNonceLifespan(ctx)
+		nonceExpiresIn := time.Now().Add(nonceLifespan).UTC()
+		nonce, err := h.r.OAuth2Storage().NewNonce(ctx, accessToken, nonceExpiresIn)
+		if err != nil {
+			h.r.Writer().WriteError(w, r, errorsx.WithStack(err))
```
aeneasr, line 1262: No need for WithStack (I think?)

oauth2/handler.go
```diff
@@ -1154,3 +1207,175 @@ func (h *Handler) logOrAudit(err error, r *http.Request) {
...
+			return
+		}
+		w.WriteHeader(http.StatusBadRequest)
+		h.r.Writer().Write(w, r, &VerifiableCredentialPrimingResponse{
```
aeneasr, line 1266: I think that `Writes` sets the status code. I believe that herodot has a method to override the status code - use that instead

oauth2/handler.go
```diff
@@ -1154,3 +1207,175 @@ func (h *Handler) logOrAudit(err error, r *http.Request) {
...
+				Description: "Could not issue a verifiable credential because the proof is missing in the request.",
+			},
+			Format:          "jwt_vc",
+			Nonce:           nonce,
+			NonceExperiesIn: int64(nonceLifespan.Seconds()),
+		})
+		return
+	}
+	if request.Proof.ProofType != "jwt" {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(fosite.ErrInvalidRequest.WithHintf("the proof type is not supported")))
```
aeneasr, line 1278: Sentence

oauth2/handler.go
```diff
@@ -1154,3 +1207,175 @@ func (h *Handler) logOrAudit(err error, r *http.Request) {
...
+		return
+	}
+
+	header, _, ok := strings.Cut(request.Proof.JWT, ".")
+	if !ok {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(fosite.ErrInvalidRequest.WithHintf("invalid JWT in proof")))
```
aeneasr, line 1284: Sentence

oauth2/handler.go
```diff
@@ -1154,3 +1207,175 @@ func (h *Handler) logOrAudit(err error, r *http.Request) {
...
+		return
+	}
+
+	rawHeader, err := base64.RawURLEncoding.DecodeString(header)
+	if err != nil {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(fosite.ErrInvalidRequest.WithHintf("invalid JWT header in proof")))
```
aeneasr, line 1290: Sentence - also consider using the JWT library to parse the header.

oauth2/handler.go
```diff
@@ -1154,3 +1207,175 @@ func (h *Handler) logOrAudit(err error, r *http.Request) {
...
+		return
+	}
+	jwk := gjson.GetBytes(rawHeader, "jwk").String()
+	proofJWK, err := josex.LoadJSONWebKey([]byte(jwk), true)
+	if err != nil {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(fosite.ErrInvalidRequest.WithHintf("invalid JWK in JWT header in proof")))
+		return
+	}
+
+	token, err := jwt.Parse(request.Proof.JWT, func(token *jwt.Token) (any, error) {
```
aeneasr, line 1300: I'm pretty sure that you can provide a custom key discovery function to the JWT library, which would be way more elegant and secure (and has lots of tests)

oauth2/handler.go
```diff
@@ -1154,3 +1207,175 @@ func (h *Handler) logOrAudit(err error, r *http.Request) {
...
+		return proofJWK, nil
+	})
+	if err != nil {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(fosite.ErrInvalidRequest.WithHintf("bad signature for JWT in proof")))
```
aeneasr, line 1304: Sentence

oauth2/handler.go
```diff
@@ -1154,3 +1207,175 @@ func (h *Handler) logOrAudit(err error, r *http.Request) {
...
+
+	rawHeader, err := base64.RawURLEncoding.DecodeString(header)
+	if err != nil {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(fosite.ErrInvalidRequest.WithHintf("invalid JWT header in proof")))
+		return
+	}
+	jwk := gjson.GetBytes(rawHeader, "jwk").String()
+	proofJWK, err := josex.LoadJSONWebKey([]byte(jwk), true)
+	if err != nil {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(fosite.ErrInvalidRequest.WithHintf("invalid JWK in JWT header in proof")))
```
aeneasr, line 1296: Sentence

oauth2/handler.go
```diff
@@ -1154,3 +1207,175 @@ func (h *Handler) logOrAudit(err error, r *http.Request) {
...
+		return proofJWK, nil
+	})
+	if err != nil {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(fosite.ErrInvalidRequest.WithHintf("bad signature for JWT in proof")))
+		return
+	}
+
+	nonce, ok := token.Claims["nonce"].(string)
+	if !ok {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(fosite.ErrInvalidRequest.WithHintf("nonce claim is missing")))
```
aeneasr, line 1310: Sentence

oauth2/handler.go
```diff
@@ -1154,3 +1207,175 @@ func (h *Handler) logOrAudit(err error, r *http.Request) {
...
+		return
+	}
+
+	if err = h.r.OAuth2Storage().IsNonceValid(ctx, accessToken, nonce); err != nil {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(fosite.ErrInvalidRequest.WithHintf("nonce is invalid")))
```
aeneasr, line 1315: Sentence

Also, why is it invalid? Give the user some insights

oauth2/handler.go
```diff
@@ -1154,3 +1207,175 @@ func (h *Handler) logOrAudit(err error, r *http.Request) {
...
+	if err != nil {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(err))
+		return
+	}
+
+	session.Claims.Add("vc", map[string]any{
+		"@context": []string{"https://www.w3.org/2018/credentials/v1"},
+		"type":     []string{"VerifiableCredential", "UserInfoCredential"},
+		"credentialSubject": map[string]any{
+			"id": fmt.Sprintf("did:jwk:%s", base64.RawURLEncoding.EncodeToString(proofJWKJSON)),
```
aeneasr, line 1332: Is this the correct encoding?

oauth2/handler.go
```diff
@@ -1154,3 +1207,175 @@ func (h *Handler) logOrAudit(err error, r *http.Request) {
...
+// swagger:parameters createVerifiableCredential
+//
+//lint:ignore U1000 Used to generate Swagger and OpenAPI definitions
+type createVerifiableCredentialRequest struct {
+	// in: body
+	Body CreateVerifiableCredentialRequestBody
+}
+
+// CreateVerifiableCredentialRequestBody The request body to request a verifiable credential.
+type CreateVerifiableCredentialRequestBody struct {
```
aeneasr, line 1357: Missing swagger annotation

oauth2/oauth2_auth_code_test.go
```diff
@@ -74,35 +83,14 @@ type clientCreator interface {
 // - [x] If `id_token_hint` is handled properly
 //   - [x] What happens if `id_token_hint` does not match the value from the handled authentication request ("accept login")
 func TestAuthCodeWithDefaultStrategy(t *testing.T) {
-	ctx := context.TODO()
+	ctx := context.Background()
```
aeneasr, line 86: Why not context.TODO?

persistence/sql/persister_nonce.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/ory/x/errorsx"
+	"github.com/ory/x/otelx"
+)
+
+func accessTokenHash(accessToken string) []byte {
+	return sha256.New().Sum([]byte(accessToken))
+}
+
+// Set the purpose to something unique to avoid ciphertext confusion with other usages of the AEAD cipher.
+var purpose = []byte("nonce")
```
aeneasr, line 26: Make this explicit for vc? Maybe `oidc_vc_nonce_draft00`?

persistence/sql/persister_nonce_test.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/ory/fosite"
+	"github.com/ory/hydra/v2/internal"
+	"github.com/ory/x/contextx"
+	"github.com/ory/x/randx"
+)
+
+func TestPersister_Nonce(t *testing.T) {
```
aeneasr, line 20: Add a test for when the purpose is not the correct one?

aeneasr (commented):
Thank you! This looks pretty good. I think there's a bit of clean-up we need to do.

Also, should we add a proper end-to-end test in playwright/cypress for this?

persistence/sql/persister_nonce.go
alnr: The nonce we return does not contain the AAD, though. Only the base64-encoded ciphertext is returned. The AAD is only constructed on-the-fly when needed, and never exposed.

alnr:
Please re-run `make sdk`

persistence/sql/persister_nonce.go
hperl: OK got it, and I agree that this is nicer, because then the nonce is a lot smaller as well.

oauth2/handler.go
hperl: I do pass the discovery function here. The key is discovered from the JWT header.

oauth2/handler.go
hperl: https://github.com/quartzjer/did-jwk/blob/main/spec.md

alnr:
Just a tiny lint issue

hperl:
> LGTM. Needs [ory/fosite#758](https://github.com/ory/fosite/pull/758) to be merged first, right?

There is a circular dependency for pushing Fosite and Hydra. Fosite uses Hydra for the OIDC conformancy tests (which fail on the Fosite PR), and will only pass with master Hydra when this PR here is merged.

So I suggest merging this first, then the Fosite PR, then remove the `replace` directive in Hydra.

alnr (approved):

oauth2/oauth2_auth_code_test.go
aeneasr: > Background returns a non-nil, empty Context. It is never canceled, has no values, and has no deadline. It is typically used by the main function, initialization, and tests, and as the top-level Context for incoming requests.

Makes sense!

oauth2/handler.go
```diff
@@ -404,6 +414,43 @@ type oidcConfiguration struct {
...
+	//
+	// JSON array containing a list of the Verifiable Credentials supported by this authorization server.
+	CredentialsSupportedDraft00 []CredentialSupportedDraft00 `json:"credentials_supported_draft_00"`
+}
+
+// Verifiable Credentials Metadata (Draft 00)
+//
+// Includes information about the supported verifiable credentials.
+//
+// swagger:model CredentialSupportedDraft00
```
aeneasr, line 433: Swagger definitions always start lowercased, unless they're enums :)

aeneasr (commented):
Very nice, just one minor swagger comment ...

aeneasr (approved):

PR #3587, opened 2023-07-31
feat: add option to specify additional SQL migrations

alnr:


driver/factory.go
```diff
@@ -125,6 +146,19 @@ func New(ctx context.Context, sl *servicelocatorx.Options, opts []OptionsModifie
 		CallRegistry(ctx, r)
 	}
 
-	c.Source(ctx).SetTracer(ctx, r.Tracer(ctx))
```
alnr, line 128: https://github.com/ory/x/pull/695

go.mod
```diff
@@ -170,6 +170,7 @@ require (
 	github.com/knadh/koanf/v2 v2.0.1 // indirect
 	github.com/kr/pretty v0.3.1 // indirect
 	github.com/kr/text v0.2.0 // indirect
+	github.com/laher/mergefs v0.1.1
```
zepatrik, line 173: I quickly checked the code, and this lib  does not recursively merge directories. We have a merge helper that does this, I guess it would be a better fit?
https://github.com/ory/x/blob/d50f3ce3b4dabde485cbe4bbf86ffb955493e7ce/fsx/merge.go

cmd/migrate_status.go
```diff
@@ -20,6 +21,7 @@ func NewMigrateStatusCmd(slOpts []servicelocatorx.Option, dOpts []driver.Options
 		RunE:  cli.NewHandler(slOpts, dOpts, cOpts).Migration.MigrateStatus,
 	}
 
+	cmdx.RegisterFormatFlags(cmd.PersistentFlags())
```
zepatrik, line 24: Why are these needed? Not that it would break anything, but if they are not actually supported, it is probably confusing for users when the flags show up in the help messages.

zepatrik (changes requested):

cmd/migrate_status.go
alnr: These are in fact required and the program will abort if they are not registered.

alnr: This was previously a bug which I worked around in cloud. This fixes the core issue.

go.mod
alnr: Neither our own implementation (`fsx.Merge`) nor `mergefs.Merge` have perfect recursive merging. See this failing test https://github.com/ory/x/compare/master...merge-test
That's unfortunate but feels like "fixing" either would be a substantial amount of work for limited benefit.

We use laher/mergefs in Kratos already. It has a much smaller implementation than `fsx.Merge`. WDYT?

alnr: fixed!

zepatrik (approved):

aeneasr (approved):

PR #3593, opened 2023-08-03
fix: query by hashed signature only on access code table

alnr:


persistence/sql/persister_oauth2.go
```diff
@@ -279,17 +261,9 @@ func (p *Persister) deleteSessionBySignature(ctx context.Context, signature stri
 	ctx, span := p.r.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.deleteSessionBySignature")
 	defer otelx.End(span, &err)
 
-	signature = p.hashSignature(ctx, signature, table)
```
hperl, line 282: We still need this.

persistence/sql/persister_oauth2.go
```diff
@@ -166,14 +166,6 @@ func SignatureHash(signature string) string {
 	return fmt.Sprintf("%x", sha512.Sum384([]byte(signature)))
 }
 
-// hashSignature prevents errors where the signature is longer than 128 characters (and thus doesn't fit into the pk).
-func (p *Persister) hashSignature(_ context.Context, signature string, table tableName) string {
```
hperl, line 170: We still need this function, because we only hash access token signatures.

persistence/sql/persister_oauth2.go
```diff
@@ -279,17 +261,9 @@ func (p *Persister) deleteSessionBySignature(ctx context.Context, signature stri
...
-	// We look for the signature as well as the hash of the signature here.
-	// This is because we now always store the hash of the signature in the database,
-	// regardless of the type of the signature. In previous versions, we only stored
-	// the hash of the signature for JWT tokens.
-	//
-	// This code will be removed in a future version.
 	err = sqlcon.HandleError(
 		p.QueryWithNetwork(ctx).
-			Where("signature IN (?, ?)", signature, SignatureHash(signature)).
+			Where("signature = ?", SignatureHash(signature)).
```
hperl, line 266:
```suggestion
			Where("signature = ?", signature).
```

persistence/sql/persister_oauth2.go
```diff
@@ -359,8 +333,8 @@ func (p *Persister) InvalidateAuthorizeCodeSession(ctx context.Context, signatur
 	return sqlcon.HandleError(
 		p.Connection(ctx).
 			RawQuery(
-				fmt.Sprintf("UPDATE %s SET active=false WHERE signature=? AND nid = ?", OAuth2RequestSQL{Table: sqlTableCode}.TableName()),
-				signature,
+				fmt.Sprintf("UPDATE %s SET active = false WHERE signature = ? AND nid = ?", OAuth2RequestSQL{Table: sqlTableCode}.TableName()),
+				SignatureHash(signature),
```
hperl, line 337: No need to hash because authorization code signatures are not hashed.

persistence/sql/persister_nid_test.go
```diff
@@ -1448,12 +1450,12 @@ func (s *PersisterTestSuite) TestInvalidateAuthorizeCodeSession() {
 
 			require.NoError(t, r.Persister().InvalidateAuthorizeCodeSession(s.t2, sig))
 			actual := persistencesql.OAuth2RequestSQL{Table: "code"}
-			require.NoError(t, r.Persister().Connection(context.Background()).Find(&actual, sig))
+			require.NoError(t, r.Persister().Connection(context.Background()).Find(&actual, persistencesql.SignatureHash(sig)))
```
hperl, line 1453: Change not needed.

persistence/sql/persister_nid_test.go
```diff
@@ -1448,12 +1450,12 @@ func (s *PersisterTestSuite) TestInvalidateAuthorizeCodeSession() {
...
 			require.Equal(t, true, actual.Active)
 
 			require.NoError(t, r.Persister().InvalidateAuthorizeCodeSession(s.t1, sig))
 			actual = persistencesql.OAuth2RequestSQL{Table: "code"}
-			require.NoError(t, r.Persister().Connection(context.Background()).Find(&actual, sig))
+			require.NoError(t, r.Persister().Connection(context.Background()).Find(&actual, persistencesql.SignatureHash(sig)))
```
hperl, line 1458: Change not needed.

persistence/sql/persister_nid_test.go
```diff
@@ -1748,10 +1750,10 @@ func (s *PersisterTestSuite) TestRevokeRefreshToken() {
 			actual := persistencesql.OAuth2RequestSQL{Table: "refresh"}
 
 			require.NoError(t, r.Persister().RevokeRefreshToken(s.t2, request.ID))
-			require.NoError(t, r.Persister().Connection(context.Background()).Find(&actual, signature))
+			require.NoError(t, r.Persister().Connection(context.Background()).Find(&actual, persistencesql.SignatureHash(signature)))
```
hperl, line 1753: Change not needed.

persistence/sql/persister_nid_test.go
```diff
@@ -1748,10 +1750,10 @@ func (s *PersisterTestSuite) TestRevokeRefreshToken() {
...
 			require.Equal(t, true, actual.Active)
 			require.NoError(t, r.Persister().RevokeRefreshToken(s.t1, request.ID))
-			require.NoError(t, r.Persister().Connection(context.Background()).Find(&actual, signature))
+			require.NoError(t, r.Persister().Connection(context.Background()).Find(&actual, persistencesql.SignatureHash(signature)))
```
hperl, line 1756: Change not needed.

persistence/sql/persister_nid_test.go
```diff
@@ -1778,10 +1780,10 @@ func (s *PersisterTestSuite) TestRevokeRefreshTokenMaybeGracePeriod() {
 			}
 
 			require.NoError(t, store.RevokeRefreshTokenMaybeGracePeriod(s.t2, request.ID, signature))
-			require.NoError(t, r.Persister().Connection(context.Background()).Find(&actual, signature))
+			require.NoError(t, r.Persister().Connection(context.Background()).Find(&actual, persistencesql.SignatureHash(signature)))
```
hperl, line 1783: Change not needed.

persistence/sql/persister_nid_test.go
```diff
@@ -1778,10 +1780,10 @@ func (s *PersisterTestSuite) TestRevokeRefreshTokenMaybeGracePeriod() {
...
 			require.Equal(t, true, actual.Active)
 			require.NoError(t, store.RevokeRefreshTokenMaybeGracePeriod(s.t1, request.ID, signature))
-			require.NoError(t, r.Persister().Connection(context.Background()).Find(&actual, signature))
+			require.NoError(t, r.Persister().Connection(context.Background()).Find(&actual, persistencesql.SignatureHash(signature)))
```
hperl, line 1786: Change not needed.

persistence/sql/persister_nid_test.go
```diff
@@ -558,11 +560,11 @@ func (s *PersisterTestSuite) DeleteAccessTokenSession() {
 			require.NoError(t, r.Persister().DeleteAccessTokenSession(s.t2, sig))
 
 			actual := persistencesql.OAuth2RequestSQL{Table: "access"}
-			require.NoError(t, r.Persister().Connection(context.Background()).Find(&actual, sig))
+			require.NoError(t, r.Persister().Connection(context.Background()).Find(&actual, persistencesql.SignatureHash(sig)))
```
hperl, line 563: Only these changes are needed in this file, the rest you can revert after you fix the code so that the hashing only happens for access tokens.

hperl (changes requested):
You found all the right places, but the hasing only applies to access tokens (and the respective table), not all the other tables (refresh, id, code, pkce, oidc).

aeneasr:
Doesn't this break BC @hperl ?

consent/handler.go
```diff
@@ -1026,12 +1026,3 @@ func (h *Handler) getOAuth2LogoutRequest(w http.ResponseWriter, r *http.Request,
 
 	h.r.Writer().Write(w, r, request)
 }
-
-func (h *Handler) flowFromCookie(r *http.Request) (*flow.Flow, error) {
```
alnr, line 1030: Unused

hperl:
> Doesn't this break BC @hperl ?

This should not not break backwards compatibility. We introduced undconditionally hashing all new access token signatures (before, we hashed only JWT access tokens). We kept the lookup for the raw signature in place so that old access tokens could still be found.

Since some time has passed, all current access tokens should have a hashed signature already, so removing the additional lookup for the raw signature can be removed safely.

hperl (approved):
LGTM!

aeneasr:
> Since some time has passed, all current access tokens should have a hashed signature already, so removing the additional lookup for the raw signature can be removed safely.

So it does have BC implications and those simply don't apply if you don't no longer have active tokens in the store. Scenarios where this isn't the case:

- Someone upgrading from hydra pre-hash to this version without running the interim version for a while. We have at least one customer I know that will run into this.
- Someone running (unreasonably high) TTLs for their access tokens 

Have we verified that there is indeed no active tokens in Ory Network any more that would be affected by this?

hperl (approved):
LGTM!

aeneasr (dismissed):

hperl (approved):

PR #3595, opened 2023-08-04
feat: wrap the final response from /oauth2/authorize in a transaction

alnr:
Primarily, this is a performance enhancement when using CockroachDB global tables.


aeneasr:
All important changes from this PR are on master already, with the exception of the transaction which is done in https://github.com/ory/hydra/pull/3763/files

PR #3601, opened 2023-08-10
fix: enable CORS with hot-reloaded origins

zepatrik:
As far as I can tell CORS cannot work right now :sweat_smile: 
See https://github.com/ory/kratos/pull/3423


cmd/server/handler.go
```diff
@@ -60,9 +59,6 @@ func EnhanceMiddleware(ctx context.Context, sl *servicelocatorx.Options, d drive
 	}
 
 	n.UseHandler(router)
-	corsx.ContextualizedMiddleware(func(ctx context.Context) (opts cors.Options, enabled bool) {
-		return d.Config().CORS(ctx, iface)
-	})
```
zepatrik, line 65: This function returns a middleware, but it was never used. So CORS is not working on master? Except for the per-client CORS?

internal/httpclient/go.sum
zepatrik, line 1: No idea what happened here, but `go mod tidy` is convinced this is correct.

cmd/server/handler.go
aeneasr: 🤔

aeneasr (approved):

cmd/server/handler.go
```diff
@@ -78,11 +74,11 @@ func RunServeAdmin(slOpts []servicelocatorx.Option, dOpts []driver.OptionsModifi
 		ctx := cmd.Context()
 		sl := servicelocatorx.NewOptions(slOpts...)
 
-		d, err := driver.New(cmd.Context(), sl, append(dOpts, driver.WithOptions(configx.WithFlags(cmd.Flags()))))
+		d, err := driver.New(cmd.Context(), sl, append(dOpts, driver.WithOptions(append(cOpts, configx.WithFlags(cmd.Flags()))...)))
```
zepatrik, line 77: Not sure why those options were not passed on?

cmd/server/handler.go
```diff
@@ -312,6 +308,9 @@ func serve(
 ) {
 	defer wg.Done()
 
+	cfg, _ := d.Config().CORS(ctx, iface)
+	handler = cors.New(cfg).Handler(handler)
```
zepatrik, line 312: Always enabled now because the function hot-reloads "enabled" as well.

aeneasr (dismissed):

aeneasr (approved):

PR #3603, opened 2023-08-10
fix: reject invalid JWKS in client configuration / dependency cleanup and bump

alnr:
1. bumped dependencies, cleaned up go.mod, and fixed some lint issues.
2. fixed two issues of missing context passing, leading to missing timeouts/cancelation and gaps in tracing
3. fixed an issue where we would accept a malformed JWKS in the OAuth client configuration (on creation or update). Once in the database, it can't be correctly deserialzed anymore. That bricks the complete `GET /admin/clients` endpoint.

Users encountering this issue are advised to fix up offending JWKS keys in their database manually. A good start point for a query is 
```sql
select id, jwks from hydra_client where jwks <> '{}';
```
From there, inspect the `jwks` column and scan for invalid keys. Those should be replaced by `{}`.


aeneasr (dismissed):
LGTM

aeneasr:
test fail

PR #3633, opened 2023-09-19
feat: add authentication options to hooks

zepatrik:
Pre-requisite for https://github.com/ory/network/issues/324


driver/config/provider.go
```diff
@@ -467,12 +467,46 @@ func (p *DefaultProvider) AccessTokenStrategy(ctx context.Context, additionalSou
 	return s
 }
 
-func (p *DefaultProvider) TokenHookURL(ctx context.Context) *url.URL {
-	return p.getProvider(ctx).RequestURIF(KeyTokenHookURL, nil)
+type HookConfig struct {
+	URL     string            `json:"url"`
+	Headers map[string]string `json:"headers"`
```
aeneasr, line 472: I would prefer if this was uniform with how we define hook configs in Kratos. WDYT?

zepatrik: I agree, and first thought I can reuse some code. But because the hooks are so vastly different, here you can't even use JSONNet, I went with this. The two fields `headers` and `url` are actually the same, right? So it is still possible to make them the same at some point.

aeneasr (approved):

PR #3639, opened 2023-09-29
feat: remove flow cookie

hperl:
This patch removes the flow cookie. All information is already tracked in the request query parameters as part of the {login|consent}_{challenge|verifier}.

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


alnr (approved):
It's beautiful.

aeneasr (approved):

consent/manager_test_helpers.go
```diff
@@ -480,17 +480,13 @@ func ManagerTests(deps Deps, m Manager, clientManager client.Manager, fositeMana
...
 					require.NoError(t, err)
 					compareAuthenticationRequest(t, c, got2.LoginRequest)
 
-					_, err = m.VerifyAndInvalidateLoginRequest(ctx, nil, loginVerifier)
-					require.Error(t, err)
-
 					loginChallenge = x.Must(f.ToLoginChallenge(ctx, deps))
 					got1, err = m.GetLoginRequest(ctx, loginChallenge)
 					require.NoError(t, err)
-					assert.True(t, got1.WasHandled)
```
aeneasr, line 493: Should this assertion not still be here? The flow was handled so it should be true right?

consent/manager_test_helpers.go
```diff
@@ -544,32 +540,18 @@ func ManagerTests(deps Deps, m Manager, clientManager client.Manager, fositeMana
...
-					consentRequest.ID = f.ConsentChallengeID.String()
+					consentRequest.ID = got2.ID
 					compareConsentRequest(t, consentRequest, got2.ConsentRequest)
 					assert.Equal(t, consentRequest.ID, got2.ID)
 					assert.Equal(t, h.GrantedAudience, got2.GrantedAudience)
 
-					// Trying to update this again should return an error because the consent request was used.
-					h.GrantedAudience = sqlxx.StringSliceJSONFormat{"new-audience", "new-audience-2"}
-					_, err = m.HandleConsentRequest(ctx, f, h)
-					require.Error(t, err)
```
aeneasr, line 557: Why is it ok to remove this assertion?

consent/strategy_default.go
```diff
@@ -658,19 +645,20 @@ func (s *DefaultStrategy) forwardConsentRequest(
...
+func (s *DefaultStrategy) verifyConsent(ctx context.Context, _ http.ResponseWriter, r *http.Request, verifier string) (_ *flow.AcceptOAuth2ConsentRequest, _ *flow.Flow, err error) {
 	ctx, span := trace.SpanFromContext(ctx).TracerProvider().Tracer("").Start(ctx, "DefaultStrategy.verifyConsent")
 	defer otelx.End(span, &err)
 
-	f, err := s.flowFromCookie(r)
+	// We decode the flow here once again because VerifyAndInvalidateConsentRequest does not return the flow
+	f, err := flowctx.Decode[flow.Flow](ctx, s.r.FlowCipher(), verifier, flowctx.AsConsentVerifier)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, errorsx.WithStack(fosite.ErrAccessDenied.WithHint("The consent verifier has already been used, has not been granted, or is invalid."))
```
aeneasr, line 655: This is not an accurate error description. `flowctx.Decode` only returns an error when the decoding fails (gob/base64).

consent/strategy_default.go
```diff
@@ -356,15 +352,13 @@ func (s *DefaultStrategy) verifyAuthentication(
...
 
-	f, err := s.flowFromCookie(r)
+	// We decode the flow from the cookie again because VerifyAndInvalidateLoginRequest does not return the flow
+	f, err := flowctx.Decode[flow.Flow](ctx, s.r.FlowCipher(), verifier, flowctx.AsLoginVerifier)
 	if err != nil {
-		return nil, errorsx.WithStack(fosite.ErrAccessDenied.WithHint("The flow cookie is missing in the request."))
-	}
-	if f.Client.GetID() != req.GetClient().GetID() {
-		return nil, errorsx.WithStack(fosite.ErrInvalidClient.WithHint("The flow cookie client id does not match the authorize request client id."))
+		return nil, errorsx.WithStack(fosite.ErrAccessDenied.WithHint("The login verifier has already been used, has not been granted, or is invalid."))
```
aeneasr, line 358: This is not an accurate error description. `flowctx.Decode` only returns an error when the decoding fails (gob/base64).

persistence/sql/persister_consent.go
```diff
@@ -349,36 +333,18 @@ func (p *Persister) HandleLoginRequest(ctx context.Context, f *flow.Flow, challe
...
-		return nil, errorsx.WithStack(fosite.ErrInvalidRequest.WithDebug("Login verifier does not match login request."))
-	}
-
-	// Update flow from login request, but keep requested at.
-	updatedFlow.NID = f.NID
-	updatedFlow.RequestedAt = f.RequestedAt
-	updatedFlow.LoginCSRF = f.LoginCSRF
-	updatedFlow.LoginVerifier = f.LoginVerifier
-	*f = *updatedFlow
-
```
aeneasr, line 381: Could you briefly explain why we needed this in the first place so I better understand the reason why it's ok to remove this?

aeneasr (changes requested):
See above

consent/manager_test_helpers.go
hperl: Previously we persisted the flow in the database, so `VerifyAndInvalidateLoginRequest` set `WasHandled` to `true` in the DB. Since we now store the flow in the login verifier, there is no place to store information about whether the flow was already handled.

This is already true even without this patch. 

If we want double-submit protection, we should build a separate `NonceService` (`new` creates a new nonce, `use` uses a nonce once) to make sure that the verifiers are only used once.

consent/manager_test_helpers.go
hperl: Because we can't detect double submit any more.

persistence/sql/persister_consent.go
hperl: Previously, we merged two flows here: The one decoded from the login verifier (`verifier` param), and the one from the cookie (`f` param). The code above ensured that these two flows are the same, so that the same flow information ends up in the cookie agan as well as in the consent challenge.

After removing the cookie flow, we do not have to do this any more.

consent/manager_test_helpers.go
aeneasr: I see - do we protect against double submit for the consent verifier? Or can I generate as many authorization codes as I want with the same consent verifier?

Can I log in as often as I want to using the same login verifier?

hperl: You can log in as often as you want, but the double-submit will be caught at the consent verification step.

I added a test case for consent verifier double submit.

aeneasr:
- Handle 500s on reuse gracefully

aeneasr (approved):

PR #3704, opened 2024-01-28
fix: add "form_post" to list of supported response modes

mitar:
Fosite supports this for some time. Not sure if Hydra on purpose does not support it or just does not list it?


aeneasr:
This was merged in another pr

PR #3770, opened 2024-05-25
feat: Refresh token expiration window

StarAurryon:
This is a rebase, squash merge and fix of the previous PR here: #2827 

Introduce optional configuration value oauth2.refresh_token_rotation.grace_period that can be set as a duration (1m,
Add a new migration: boolean used SQL column in hydra_oauth2_refresh

When grace_period has positive duration the act of using a refresh token will not invalidate the token and instead mark it as used in the database and set its expiration time to the UTCNOW + the value of the grace period. This will allow a client to continue to use a refresh token for the duration of the grace period.

## Related issue(s)
#1831

## Checklist

- [X ] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [ X] I have referenced an issue containing the design document if my change
      introduces a new feature.
- [ X] I am following the
- [X ] I have read the [security policy](../security/policy).
- [ X] I confirm that this pull request does not address a security
      the changes.
- [X ] I have added tests that prove my fix is effective or that my feature
      works.

### DIFF
--- a/driver/config/provider.go
+++ b/driver/config/provider.go
@@ -101,6 +101,7 @@ const (
 	KeyExcludeNotBeforeClaim                     = "oauth2.exclude_not_before_claim"
 	KeyAllowedTopLevelClaims                     = "oauth2.allowed_top_level_claims"
 	KeyMirrorTopLevelClaims                      = "oauth2.mirror_top_level_claims"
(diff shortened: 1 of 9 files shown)


StarAurryon:
Just a quick question here regarding grace period. Shouldn't we also ensure that when a new refresh token is issued the grand parent refresh token is revoked and also all the "brother" refresh tokens are revoked ?

My understanding is that hydra does not need to think like that a the moment as no grace period was implemented.

PR #3790, opened 2024-07-02
feat: Pass sanitized payload+client metadata to token hook

terev:
Enables passing additional useful data to token hooks. Previously only a constrained set of parameters was being passed to token hooks. This PR adds functionality to pass all non sensitive parameters. Additionally it includes any metadata the client has configured.

I added a test to ensure that all supported grant types are handled by the function that returns sensitive parameters. This should help to avoid a situation where sensitive parameters are leaked to a hook if support for a new grant type is added, and it requires sensitive payload parameters. However there is still the case of sensitive parameters required for client authentication. I included the two sensitive parameters I'm aware of `client_secret` and `client_assertion` but if support for new schemes is added its possible this may get missed. Is this a concern?

## Related issue(s)
Related work I referenced: https://github.com/ory/hydra/pull/3427

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments
I added a helper function `GrantTypesSupported` that just returns all of the grant types hydra supports. This is useful in the sensitive grant type parameter test to ensure it's exhaustive. I made this a function as a workaround to ensure data integrity because golang doesn't support constant collection types.

I updated the database container images used by the test make rule to tags/images that support multiplatforms (at least linux/amd64 and linux/arm64).


Makefile
```diff
@@ -64,9 +64,9 @@ test-resetdb: node_modules
 	docker rm --force --volumes hydra_test_database_mysql || true
```
aeneasr, line 64: Please revert the changes in these files because this would need a change in other files (CI, dockertest) as well to keep consistency, and it's unrleated to this PR. Thanks!

oauth2/token_hook.go
```diff
@@ -166,12 +196,28 @@ func TokenHook(reg interface {
...
+			return slices.Contains(sensitiveParameters, s)
+		})
 		request := Request{
 			ClientID:        requester.GetClient().GetID(),
+			ClientMetadata:  clientMeta,
 			GrantedScopes:   requester.GetGrantedScopes(),
 			GrantedAudience: requester.GetGrantedAudience(),
 			GrantTypes:      requester.GetGrantTypes(),
-			Payload:         requester.Sanitize([]string{"assertion"}).GetRequestForm(),
+			Payload:         payload,
```
aeneasr, line 220: This is a fundamentally different approach to forwarding request data:

- Before: We restrict the list of parameters to a set of known values that don't leak sensitive data
- Now: We try to guesstimate a list of parameters we better not include but can not guarantee that all requirements are met

I'm really not sure if I agree with this approach. I would rather extend the default allowed parameters to include all known parameters that are OK to forward like `prompt` etc.

The problem with the parameters is that it includes data from the POST payload which is quite likely to contain sensitive things like tokens, passwords, assertions, signatures, ...

oauth2/token_hook.go
```diff
@@ -166,12 +196,28 @@ func TokenHook(reg interface {
...
+		sensitiveParameters := append([]string{"client_secret", "client_assertion"}, sensitiveAccessRequestParametersForGrantTypes(requester.GetGrantTypes()...)...)
+
+		payload := maps.Clone(requester.GetRequestForm())
+		// Delete all sensitive parameters from the access request form before sending it to the hook.
+		maps.DeleteFunc(payload, func(s string, _ []string) bool {
+			return slices.Contains(sensitiveParameters, s)
+		})
 		request := Request{
 			ClientID:        requester.GetClient().GetID(),
+			ClientMetadata:  clientMeta,
```
aeneasr, line 216: I mean if we have the client it probably makes sense to just send the client instead.

aeneasr (changes requested):

oauth2/token_hook.go
terev: @aeneasr 

Thanks for having a look, really appreciate your team's work! My goal is to pass any **additional** and non sensitive parameters to the token hook. I was unable to find any guidelines around transmission of additional token endpoint parameters other than that they have to be stored. So I figured this meant they might be fine to utilize in an extensible manner.

For my approach, my thinking was that the set of sensitive parameters for a given token request is actually constrained based on the grant types requested. So its easy to determine this set and filter them out. Is there additional sensitive parameters I didn't include? For what its worth assertion is already specifically leaked to token hooks.

A couple other approaches i considered:
- Always filter all sensitive parameters out but I opted to generate the set based on grant type so there's a clean signal of test failure if a new grant type is added but not supported by this function. 
- Make the list of acceptable additional parameters configurable in the token_hook (with validation to exclude any oauth params).

terev:
@aeneasr I've reduced the scope of this mr to just passing the oauth client to the token hook, as it seems like the extra  parameters part of the original motivation was controversial. I would still be very interested to hear your thoughts on this if you have any.

aeneasr (approved):

PR #3819, opened 2024-08-13
fix: broken JSON round-tripping for custom claims

alnr:
Adding custom claims with numerical types (think JavaScript Number) previously did not
round-trip through Hydra correctly. For example, passing UNIX timestamps in custom claims
would end up as floating points in exponential notation in the final token. That, in turn,
confused or broke downstream consumers of the token, including Kratos.

Ref https://github.com/go-jose/go-jose/issues/144

### DIFF
--- a/consent/strategy_oauth_test.go
+++ b/consent/strategy_oauth_test.go
@@ -289,6 +289,7 @@ func TestStrategyLoginConsentNext(t *testing.T) {
 
 		subject := "aeneas-rekkas"
 		c := createDefaultClient(t)
+		now := 1723546027 // Unix timestamps must round-trip through Hydra without converting to floats or similar
 		testhelpers.NewLoginConsentUI(t, reg.Config(),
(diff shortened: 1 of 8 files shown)


aeneasr (approved):

aeneasr:
There was this PR: https://github.com/ory/hydra/pull/3722

Is that the same issue, and if yes, #3722 seems to use the number decoding in more places.

aeneasr:
Also, are there any implications for backwards compatibility with prior consent sessions?

alnr:
Yes I did see https://github.com/ory/hydra/pull/3722. As the author discovered, using `json.Number` does not fix this issue. The underlying problem is described here: https://github.com/go-jose/go-jose/issues/144

I don't think this will cause any regressions. This change mostly or exclusively impacts how _numbers_ are serialized into the final token, it doesn't really change any underlying data. While we're handling the custom claims in hydra, we serialize+deserialize them to/from JSON a couple of times. Each time, any "type" information about whether the number is a float or integer is lost since this distinction doesn't exist in JSON.

PR #3826, opened 2024-08-29
chore: compile a static binary in Docker

alnr:
Move from distroless/static-debian12 to distroless/base-nossl-debian12 to get fewer CVE alerts

Uses the technique described in https://www.arp242.net/static-go.html

### DIFF
--- a/.docker/Dockerfile-build
+++ b/.docker/Dockerfile-build
@@ -9,17 +9,16 @@ RUN apt-get update && apt-get upgrade -y &&\
 COPY go.mod go.sum ./
 COPY internal/httpclient/go.* ./internal/httpclient/
 
-ENV GO111MODULE on
-ENV CGO_ENABLED 1
+ENV CGO_ENABLED=1
 
 RUN go mod download
 
 COPY . .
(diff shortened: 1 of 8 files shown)


alnr:
> Does SQLite with JSON still work correctly if this patch is applied?

I believe so. The `json1` tag may in fact no longer be necessary. See https://www.sqlite.org/compile.html (`SQLITE_ENABLE_JSON1`).

Wanted to get a CI run through first of all, and see if I can find any regressions.

alnr:
https://github.com/mattn/go-sqlite3/issues/1044

[This commit](https://github.com/mattn/go-sqlite3/commit/aa1e904220efab26cfe86242d52d0403e64b95c4) deleted all build tags associated with JSON. The first released version with that change is 1.14.13. We should be good.

aeneasr:
Oh, that's great!

aeneasr:
The best way to test for regression is to probably execute the quickstart using the updated image

alnr:
Image works! Tested a couple of things, including QuickStart

PR #3845, opened 2024-09-20
feat: built-in login/consent UI for `hydra perform authorization-code`

alnr:
This allows you to easily test an Authorization Code flow against hydra without having to have a consent UI otherwise available.

If you want to use the built-in UI, configure the login+consent UIs to point to the same server as the redirect URI:
```sh
ory patch oauth2-config --project $PROJECT_ID --replace '/urls/login="http://127.0.0.1:4446/login"' --replace '/urls/consent="http://127.0.0.1:4446/consent"'
```

Then run
```sh
ory perform authorization-code --project $PROJECT_ID --client-id xyz --client-secret abc
```
or
```sh
hydra perform authorization-code --client-id xyz --client-secret abc --auth-url http://... --token-url http://...
```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -94,7 +94,7 @@ jobs:
           GOGC: 100
         with:
           args: --timeout 10m0s
-          version: v1.55.2
+          version: v1.61.0
           skip-pkg-cache: true
       - name: Run go-acc (tests)
         run: |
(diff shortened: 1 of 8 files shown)


aeneasr:
Does this require changes to the docs? Such as the quickstart guides?

alnr:
> Does this require changes to the docs? Such as the quickstart guides?

This is for very advanced users only. If someone finds this feature in the source code, they will figure out how to configure Hydra to use it, IMO.

alnr:
All the quickstart tutorials etc continue to functions as before.

Issue #3829, opened 2024-09-02
Incorrect Handling of Scopes with Special Character  "|"  in  scp  Claim


Ajayn84:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/hydra/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/hydra/blob/master/CONTRIBUTING.md).
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://www.ory.sh/l/sign-up-newsletter).

### Describe the bug

When using a scope that includes the special character "|" (e.g., A|B), we expect the scp claim in the access token to contain the scope as a single string, i.e.,["A|B"]. However, the actual result is that the scope is split into separate elements, resulting in
["A", "B"]

### Reproducing the bug

1. Create a client having  scope with the special character "|", such as "A|B".
2. Perform oauth2 login with hydra 
3. Request an access token with this scope.
4. Inspect the scp claim in the generated access token.

### Version

2.2

### On which operating system are you observing this issue?

None

### In which environment are you deploying?

Docker

### Additional Context

Looking at the scopes stored in hydra_client table, the scopes are stored space seperated, but in hydra_oauth2_access table these are stored using "|", might be resulting into this issue

Ajayn84:
After a bit of digging, have found the issue to be during persisting the scopes in some tables we are using "|" (pipe) as the delimiter. below are the code references https://github.com/ory/hydra/blob/f83193f90814691d10d88493eb6357cf37565115/persistence/sql/persister_oauth2.go#L113 and https://github.com/ory/hydra/blob/f83193f90814691d10d88493eb6357cf37565115/persistence/sql/persister_oauth2.go#L161

If we escape and unescape using some logic like below , then it might resolve this issue
![image](https://github.com/user-attachments/assets/09dbfd57-3028-4b59-9833-8b6b64d72a37)
![image](https://github.com/user-attachments/assets/07c942b2-8d77-46bb-8b41-620c42cc7b3c)
![image](https://github.com/user-attachments/assets/deb14f79-85f9-4bdf-a47b-1fca12ccd964)

Ajayn84:
[...]

PR #3849, opened 2024-09-25
Update persister_oauth2.go to handle special character | coming  in t…

Ajayn84:
…he scopes as part of consent request

Url encoded and decoded while fetching values from the table,  as "|" is a seperator used to store scopes

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/persistence/sql/persister_oauth2.go
+++ b/persistence/sql/persister_oauth2.go
@@ -110,8 +110,8 @@ func (p *Persister) sqlSchemaFromRequest(ctx context.Context, signature string, 
 		RequestedAt:       r.GetRequestedAt(),
 		InternalExpiresAt: sqlxx.NullTime(expiresAt),
 		Client:            r.GetClient().GetID(),
-		Scopes:            strings.Join(r.GetRequestedScopes(), "|"),
-		GrantedScope:      strings.Join(r.GetGrantedScopes(), "|"),
+		Scopes:            strings.Join(escapeDelimiter(r.GetRequestedScopes()), "|"),
+		GrantedScope:      strings.Join(escapeDelimiter(r.GetGrantedScopes()), "|"),
 		GrantedAudience:   strings.Join(r.GetGrantedAudience(), "|"),
 		RequestedAudience: strings.Join(r.GetRequestedAudience(), "|"),
 		Form:              r.GetRequestForm().Encode(),
@@ -158,8 +158,8 @@ func (r *OAuth2RequestSQL) toRequest(ctx context.Context, session fosite.Session
 		RequestedAt: r.RequestedAt,
 		// ExpiresAt does not need to be populated as we get the expiry time from the session.
 		Client:            c,
-		RequestedScope:    stringsx.Splitx(r.Scopes, "|"),
-		GrantedScope:      stringsx.Splitx(r.GrantedScope, "|"),
+		RequestedScope:    unescapeDelimiter(r.Scopes),
+		GrantedScope:      unescapeDelimiter(r.GrantedScope),
 		RequestedAudience: stringsx.Splitx(r.RequestedAudience, "|"),
 		GrantedAudience:   stringsx.Splitx(r.GrantedAudience, "|"),
 		Form:              val,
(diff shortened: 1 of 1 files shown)


Ajayn84:
@hperl @aeneasr @alnr request your reviews

Ajayn84:
Links to https://github.com/ory/hydra/issues/3829

aeneasr (changes requested):
This is a backwards incompatible change that will break all existing rows. I don't think there's appetite to fix this at the moment and the solution would need to be backwards compatible and well-tested. It's probably easier to just adjust the scope charset

Ajayn84:
> t's probably easier to just adjust the scope charset

@aeneasr  can you please elaborate, i did try to encode , but since this logic is more controlled in the Hydra, it just took that param as "%7C" and responded with scopes having "%7C" instead of "|"

We can replace the | char ,as its a requirement from Hl7 Smart App launch https://www.hl7.org/fhir/smart-app-launch/scopes-and-launch-context.html#finer-grained-resource-constraints-using-search-parameters

Were you suggesting some other way to fix this issue?

persistence/sql/persister_oauth2.go
```diff
@@ -118,8 +118,8 @@ func (p *Persister) sqlSchemaFromRequest(ctx context.Context, signature string,
 		RequestedAt:       r.GetRequestedAt(),
 		InternalExpiresAt: sqlxx.NullTime(expiresAt),
 		Client:            r.GetClient().GetID(),
-		Scopes:            strings.Join(r.GetRequestedScopes(), "|"),
-		GrantedScope:      strings.Join(r.GetGrantedScopes(), "|"),
+		Scopes:            strings.Join(escapeDelimiter(r.GetRequestedScopes()), "|"),
```
Ajayn84, line 121: we are escaping delimter "|"  using queryescape and then joining with "|" delimiter, so that scopes with "|" are replaced with and then joined using "|"

persistence/sql/persister_oauth2.go
```diff
@@ -179,13 +179,23 @@ func (r *OAuth2RequestSQL) toRequest(ctx context.Context, session fosite.Session
...
+	}
+
 	return &fosite.Request{
 		ID:          r.Request,
 		RequestedAt: r.RequestedAt,
 		// ExpiresAt does not need to be populated as we get the expiry time from the session.
 		Client:            c,
-		RequestedScope:    stringsx.Splitx(r.Scopes, "|"),
-		GrantedScope:      stringsx.Splitx(r.GrantedScope, "|"),
+		RequestedScope:    scopes,
```
Ajayn84, line 197: we are splitting the data from db with "|" and then unescaping to get "|" back in the scopes if any

sagarshah1983:
@Ajayn84 Have you found a workaround with this? We are also facing the same issue for same SMART on FHIR implementation.

@aeneasr - Could you please share the other suggestion that you may have on how to go about fixing this in hydra?

aeneasr:
Hello, we already have methods to properly delimit string arrays using JSON encoding if I'm not mistaken in our sqlxx package. However, changing the way these things are parsed is a breaking change and requires a careful migration path as it can break all existing records. Given the risk associated, we just ask you to use a different delimiter in scopes.

Furthermore, `|` is not a valid url character and would need to be decoded in URLs, which makes it not that great of an option for use in the scope field, further reducing the likelihood of getting this merged.

Ajayn84:
We cant replace the "|" pipe char with other char, as its a regulatory ask in healthcare. Hl7 Smart App launch link where it indicates the same https://www.hl7.org/fhir/smart-app-launch/scopes-and-launch-context.html#finer-grained-resource-constraints-using-search-parameters

Ajayn84:
> @Ajayn84 Have you found a workaround with this? We are also facing the same issue for same SMART on FHIR implementation.
> 
> @aeneasr - Could you please share the other suggestion that you may have on how to go about fixing this in hydra?

No @sagarshah1983 , we are also trying to implement Smart on FHIR and specs indicate use of "|" in scopes, without a workaround

Ajayn84:
> Thank you for pointing me to the spec - in that case the use case is legitimate in my view. Still, it's not trivial to implement this and if we fix it it should be a long-term solution (i.e. using the appropriate data types). I will synchronize with @vinckr to figure out what to best do here as I'm actually PTO :)

Thanks @aeneasr.

vinckr:
@Ajayn84 
can you write me an email at vincent@ory.sh to discuss the options further please?

Ajayn84:
> @Ajayn84 can you write me an email at [vincent@ory.sh](mailto:vincent@ory.sh) to discuss the options further please?

Sure @vinckr

PR #3853, opened 2024-10-01
fix: change comment on revokeOAuth2LoginSessions

sakai-303:
## Related issue(s)
This is not an issue, but the related discussions can be found below.
https://github.com/orgs/ory/discussions/118

## Checklist

- [x] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [x] I have referenced an issue containing the design document if my change
      introduces a new feature.
- [x] I am following the
      [contributing code guidelines](../blob/master/CONTRIBUTING.md#contributing-code).
- [x] I have read the [security policy](../security/policy).
- [x] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security vulnerability, I
      confirm that I got the approval (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [x] I have added tests that prove my fix is effective or that my feature
      works.
- [x] I have added or changed [the documentation](https://github.com/ory/docs).

PR #3858, opened 2024-10-10
revert: change comment on revokeOAuth2LoginSessions (#3853)

alnr:
See https://github.com/orgs/ory/discussions/118


aeneasr:
My bad, I confused it with consent.

PR #3860, opened 2024-10-14
feat: graceful refresh token rotation

aeneasr:
## Related issue(s)

Closes https://github.com/ory/hydra/pull/3770

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/driver/config/provider.go
+++ b/driver/config/provider.go
@@ -672,9 +672,9 @@ func (p *DefaultProvider) cookieSuffix(ctx context.Context, key string) string {
 }
 
 func (p *DefaultProvider) RefreshTokenRotationGracePeriod(ctx context.Context) time.Duration {
-	var duration = p.getProvider(ctx).DurationF(KeyRefreshTokenRotationGracePeriod, 0)
-	if duration > time.Hour {
+	gracePeriod := p.getProvider(ctx).DurationF(KeyRefreshTokenRotationGracePeriod, 0)
+	if gracePeriod > time.Hour {
 		return time.Hour
 	}
-	return p.getProvider(ctx).DurationF(KeyRefreshTokenRotationGracePeriod, 0)
+	return gracePeriod
 }
--- a/driver/config/provider.go
+++ b/driver/config/provider.go
@@ -102,6 +102,7 @@ const (
 	KeyExcludeNotBeforeClaim                     = "oauth2.exclude_not_before_claim"
 	KeyAllowedTopLevelClaims                     = "oauth2.allowed_top_level_claims"
 	KeyMirrorTopLevelClaims                      = "oauth2.mirror_top_level_claims"
+	KeyRefreshTokenRotationGracePeriod           = "oauth2.grant.refresh_token.rotation_grace_period" // #nosec G101
 	KeyOAuth2GrantJWTIDOptional                  = "oauth2.grant.jwt.jti_optional"
 	KeyOAuth2GrantJWTIssuedDateOptional          = "oauth2.grant.jwt.iat_optional"
 	KeyOAuth2GrantJWTMaxDuration                 = "oauth2.grant.jwt.max_ttl"
@@ -668,4 +669,12 @@ func (p *DefaultProvider) cookieSuffix(ctx context.Context, key string) string {
 	}
 
 	return p.getProvider(ctx).String(key) + suffix
+}
+
+func (p *DefaultProvider) RefreshTokenRotationGracePeriod(ctx context.Context) time.Duration {
+	var duration = p.getProvider(ctx).DurationF(KeyRefreshTokenRotationGracePeriod, 0)
+	if duration > time.Hour {
+		return time.Hour
+	}
+	return p.getProvider(ctx).DurationF(KeyRefreshTokenRotationGracePeriod, 0)
 }
--- a/internal/config/config.yaml
+++ b/internal/config/config.yaml
@@ -402,6 +402,19 @@ oauth2:
   session:
     # store encrypted data in database, default true
     encrypt_at_rest: true
+    ## refresh_token_rotation
+  # By default Refresh Tokens are rotated and invalidated with each use. See https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#section-4.13.2 for more details
+  refresh_token_rotation:
+    #
+    ## grace_period
+    # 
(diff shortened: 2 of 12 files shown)


driver/config/provider.go
```diff
@@ -669,3 +670,11 @@ func (p *DefaultProvider) cookieSuffix(ctx context.Context, key string) string {
 
 	return p.getProvider(ctx).String(key) + suffix
 }
+
+func (p *DefaultProvider) RefreshTokenRotationGracePeriod(ctx context.Context) time.Duration {
+	var duration = p.getProvider(ctx).DurationF(KeyRefreshTokenRotationGracePeriod, 0)
+	if duration > time.Hour {
+		return time.Hour
+	}
+	return p.getProvider(ctx).DurationF(KeyRefreshTokenRotationGracePeriod, 0)
```
zepatrik, line 679:
```suggestion
	gracePeriod := p.getProvider(ctx).DurationF(KeyRefreshTokenRotationGracePeriod, 0)
	if gracePeriod > time.Hour {
		return time.Hour
	}
	return gracePeriod
```

internal/config/config.yaml
```diff
@@ -402,6 +402,19 @@ oauth2:
...
     # store encrypted data in database, default true
     encrypt_at_rest: true
+    ## refresh_token_rotation
+  # By default Refresh Tokens are rotated and invalidated with each use. See https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#section-4.13.2 for more details
+  refresh_token_rotation:
+    #
+    ## grace_period
+    #
+    # Set the grace period for a refresh token to allow it to be used for the duration of this configuration after its
+    #	first use. New refresh tokens will continue to be issued.
```
zepatrik, line 412:
```suggestion
    # Set the grace period for refresh tokens to be reused. Such reused tokens will result in multiple refresh tokens being issued.
```

oauth2/fosite_store_helpers.go
```diff
@@ -553,6 +554,68 @@ func testHelperRevokeAccessToken(x InternalRegistry) func(t *testing.T) {
...
+
+	return func(t *testing.T) {
+		t.Run("Revokes refresh token when grace period not configured", func(t *testing.T) {
+			// SETUP
+			m := x.OAuth2Storage()
+			ctx := context.Background()
+
+			refreshTokenSession := fmt.Sprintf("refresh_token_%d", time.Now().Unix())
+			err := m.CreateRefreshTokenSession(ctx, refreshTokenSession, &defaultRequest)
+			assert.NoError(t, err, "precondition failed: could not create refresh token session")
```
zepatrik, line 567:
```suggestion
			require.NoError(t, err, "precondition failed: could not create refresh token session")
```
same below...

persistence/sql/persister_oauth2.go
```diff
@@ -429,7 +455,34 @@ func (p *Persister) CreateRefreshTokenSession(ctx context.Context, signature str
...
+	}
+
+	fositeRequest, err := r.toRequest(ctx, session, p)
+	if err != nil {
+		return nil, err
+	}
+
+	if r.Active {
+		return fositeRequest, nil
+	}
```
zepatrik, line 474: I guess that we should avoid unnecessary decryption of the data if possible?

```suggestion
	if r.Active {
		return r.toRequest(ctx, session, p)
	}
```

persistence/sql/persister_oauth2.go
```diff
@@ -483,10 +536,25 @@ func (p *Persister) RevokeRefreshToken(ctx context.Context, id string) (err erro
...
+	/* #nosec G201 table is static */
+	return sqlcon.HandleError(
+		p.Connection(ctx).
+			RawQuery(
+				fmt.Sprintf("UPDATE %s SET active=false, first_used_at = CURRENT_TIMESTAMP WHERE request_id=? AND nid = ? AND active=true", OAuth2RequestSQL{Table: sqlTableRefresh}.TableName()),
+				id,
+				p.NetworkID(ctx),
+			).
+			Exec(),
+	)
```
zepatrik, line 557: Do we need the branching here? I would prefer to just set the `first_used_at` column every time, even if the grace period is not defined. It will be handled correctly in `GetRefreshTokenSession` regardless.
The way it is right now, enabling the grace period would only affect new refresh requests.
```suggestion
	/* #nosec G201 table is static */
	return sqlcon.HandleError(
		p.Connection(ctx).
			RawQuery(
				fmt.Sprintf("UPDATE %s SET active=false, first_used_at = CURRENT_TIMESTAMP WHERE request_id=? AND nid = ? AND active=true", OAuth2RequestSQL{Table: sqlTableRefresh}.TableName()),
				id,
				p.NetworkID(ctx),
			).
			Exec(),
	)
```

zepatrik (changes requested):

persistence/sql/persister_oauth2.go
zepatrik: I see now, we have to return the request even with error conditions...

aeneasr:
> From the original PR:
> 
> > Shouldn't we also ensure that when a new refresh token is issued the grand parent refresh token is revoked and also all the "brother" refresh tokens are revoked?

I think we test this here: https://github.com/ory/hydra/pull/3860/files#diff-6d883efffdabd9715dc9872121018df30a5843c81e25dc6c4af2c3edc13fb21cR442

persistence/sql/persister_oauth2.go
aeneasr: You mean because we test if the config is enabled anyways, right?

zepatrik: We're using the correct config when we attempt to use the token, so I don't think we should do it here as well.

zepatrik: I don't see why we should not always set the `first_used_at` column.

1. We're updating the row regardless.
1. It can help with debugging to have it set.
2. Enabling graceful refresh later also enables it for recently refreshed tokens, not just for the ones where the setting was already enabled.

aeneasr:
@zepatrik can you please approve if you're fine with the state as it is?

aeneasr:
I have added a test case that revokes the consent session and checks that all tokens are correctly invalidated. Tests pass :)

zepatrik (approved):

Issue #3863, opened 2024-10-25
GetOrGenerateKeys locking leads to significant service degredation with high request saturation


awill1988:
### Preflight checklist

- [ ] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/hydra/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/hydra/blob/master/CONTRIBUTING.md).
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://www.ory.sh/l/sign-up-newsletter).

### Describe the bug

When operating Hydra in production, it has been observed through tracing that the `GetOrGenerateKeys` is called routinely for the Authorization Endpoint, Token Endpoint, and WellKnown endpoints. That function leverages two RWMutexes to store the keysets in memory.

As identified in https://github.com/ory/hydra/discussions/3662, there can be problems when the tail latency of the database query to retrieve JWKs is above a certain threshold under highly concurrent conditions.

Additionally, when considering the locking logic for `GetOrGenerateKeys`, there may be an opportunity to leverage `RLock()` when reading from the database and then acquiring the write lock only when generating to enable more throughput for this endpoint.

### Reproducing the bug

To recreate the issue locally:

## Strategy

1. **Database Latency** - Run `hydra` with an artificially set latency for the `persister_jwk`'s `GetKeySet` that mimics real-world conditions (>= 25ms).
2. **Request Saturation** - Using any load testing tool like [bombardier](https://github.com/codesenberg/bombardier) to saturate the server with hundreds of concurrent client connections against `/.well-known/openid-configuration` endpoint.

## Steps

0. Install bombardier: `go install github.com/codesenberg/bombardier@latest`
1. Run cockroachdb instance: `docker compose -f quickstart.yml -f quickstart-cockroach.yml up -d cockroachd`
2. Run migrations: `docker compose -f quickstart.yml -f quickstart-cockroach.yml run hydra-migrate`
3. Add artificial latency to [`persister_jwk.go`](https://github.com/awill1988/hydra/blob/9cc5f28825bb80398a9830d4fcb220147f74504c/persistence/sql/persister_jwk.go#L170): `time.Sleep(25 * time.Millisecond)`
[...]

PR #3865, opened 2024-10-28
refactor: jwk aquire lock when generating

awill1988:
Update the locking behavior of `jwk.GetOrGenerateKeys` to acquire a read lock to fetch keys  and only acquire the write lock when generating and persisting new keys.

## Related issue(s)

#3863

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

I can attempt to write up more unit tests if that is required, but since this is not a significant change I am wondering if the existing tests can cover it?

I am also not sure what would need to be written for documentation but I am happy to add it where it is preferred.

On the related issue there are some reproducible steps that can be used to demonstrate the benefit of this refactor, but I can specify steps or update here if needed.

### DIFF
--- a/jwk/helper.go
+++ b/jwk/helper.go
@@ -12,7 +12,6 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"encoding/pem"
-	"sync"
 
 	hydra "github.com/ory/hydra-client-go/v2"
 
@@ -26,34 +25,16 @@ import (
 	"github.com/pkg/errors"
 )
 
-var mapLock sync.RWMutex
-var locks = map[string]*sync.RWMutex{}
-
-func getLock(set string) *sync.RWMutex {
-	mapLock.Lock()
-	defer mapLock.Unlock()
-	if _, ok := locks[set]; !ok {
-		locks[set] = new(sync.RWMutex)
-	}
-	return locks[set]
-}
-
 func EnsureAsymmetricKeypairExists(ctx context.Context, r InternalRegistry, alg, set string) error {
 	_, err := GetOrGenerateKeys(ctx, r, r.KeyManager(), set, set, alg)
 	return err
 }
 
 func GetOrGenerateKeys(ctx context.Context, r InternalRegistry, m Manager, set, kid, alg string) (private *jose.JSONWebKey, err error) {
-	getLock(set).RLock()
 	keys, err := m.GetKeySet(ctx, set)
-	getLock(set).RUnlock()
 
 	if errors.Is(err, x.ErrNotFound) || keys != nil && len(keys.Keys) == 0 {
 		r.Logger().Warnf("JSON Web Key Set \"%s\" does not exist yet, generating new key pair...", set)
-
-		getLock(set).Lock()
-		defer getLock(set).Unlock()
-
 		keys, err = m.GenerateAndPersistKeySet(ctx, set, kid, alg, "sig")
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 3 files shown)


jwk/helper.go
```diff
@@ -44,12 +44,16 @@ func EnsureAsymmetricKeypairExists(ctx context.Context, r InternalRegistry, alg,
 }
 
 func GetOrGenerateKeys(ctx context.Context, r InternalRegistry, m Manager, set, kid, alg string) (private *jose.JSONWebKey, err error) {
-	getLock(set).Lock()
-	defer getLock(set).Unlock()
-
+	getLock(set).RLock()
```
aeneasr, line 47: I'm wondering why the lock is there in the first place. We're using a SQL backend for this, which is atomic.

awill1988: that's a good point..

awill1988: @terev if it were to keep the `map[string]*RWMutex` or maybe just `map[string]*Mutex` and lock on key generation, do you think that would work for the key set initialization condition?

terev: Yeah I think a plain mutex per keyset is fine. But maybe keep the rwlock for accessing the map. But actually use the read lock in the right case there. But to be clear I'm not part of ory so it aesnar disagrees that's cool too.

awill1988: > But to be clear I'm not part of ory so if aeneasr disagrees that's cool too.
thanks anyways @terev 😄!

Okay, @aeneasr, I think I'm following what you're saying about not even needing the read lock.

I'll read up on the `test/conformance` to grok what's going on there 🤞🏻

Also found a couple snapshots that looked like they needed a quick patch so I grabbed those just in case.

jwk/helper.go
```diff
@@ -44,12 +44,14 @@ func EnsureAsymmetricKeypairExists(ctx context.Context, r InternalRegistry, alg,
...
-	getLock(set).Lock()
-	defer getLock(set).Unlock()
-
 	keys, err := m.GetKeySet(ctx, set)
+
 	if errors.Is(err, x.ErrNotFound) || keys != nil && len(keys.Keys) == 0 {
 		r.Logger().Warnf("JSON Web Key Set \"%s\" does not exist yet, generating new key pair...", set)
+
+		l := getLock(set)
+		defer l.Lock()
```
terev, line 53: In case you missed it, I think you meant to lock the returned lock then defer unlock. Same thing below on line 69.

At the end of this `if` block it may make sense to directly return `FindPrivateKey(keys)`. If we don't it opens up the potential for deadlock if `FindPrivateKey` fails and we try to generate another keyset. Also I don't really think it would help to generate another keyset if the one we just generated didn't have a private key.

alnr:
This whole code is unsafe in the presence of multiple hydra instances. We would need to change the database code to do an `INSERT INTO ... ON UPDATE DO NOTHING RETURNING *` to properly synchronize.

aeneasr:
> This whole code is unsafe in the presence of multiple hydra instances. We would need to change the database code to do an `INSERT INTO ... ON UPDATE DO NOTHING RETURNING *` to properly synchronize.

While this is true, it doesn't really matter, because the whole thing will eventually settle. Even if we created 20 keys, it will be like if we rotated through 20 keys and just use the newest key. So the impact is really minimal in my view

alnr:
That is:
1. `SELECT * FROM HYDRA_JWK`
2. a) if found, return
2. b) if not found, generate a new one in memory (could be done under a mutex to throttle CPU usage)
3. `INSERT INTO hydra_jwk ON CONFLICT DO NOTHING`
4. `SELECT * FROM hydra jwk`
5. Continue with the key returned from the last select.

Postgres and other DBs which support `RETURNING` could do 3+4 in one step as an optimization (optional, since this is a rare event).

aeneasr:
I think that analysis is correct. So all we need really is a lock for when we generate the key and not for reading, which also solves the issue here

aeneasr:
https://github.com/ory/hydra/pull/3866/files

jwk/helper.go
awill1988: whoops!

awill1988:
closing in favor of #3866

PR #3866, opened 2024-10-29
fix: only lock for key generation

### LINKED ISSUE #3863: GetOrGenerateKeys locking leads to significant service degredation with high request saturation

aeneasr:
Closes #3865
Closes #3863

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/jwk/helper.go
+++ b/jwk/helper.go
@@ -72,7 +72,7 @@ func GetOrGenerateKeys(ctx context.Context, r InternalRegistry, m Manager, set, 
 		return nil, err
 	}
 
-	privKey, err := FindPrivateKey(keys)
+	privKey, err = FindPrivateKey(keys)
 	if err != nil {
 		return nil, err
 	}
--- a/jwk/helper.go
+++ b/jwk/helper.go
@@ -45,10 +45,6 @@ func EnsureAsymmetricKeypairExists(ctx context.Context, r InternalRegistry, alg,
 
 func GetOrGenerateKeys(ctx context.Context, r InternalRegistry, m Manager, set, kid, alg string) (private *jose.JSONWebKey, err error) {
 	keys, err := m.GetKeySet(ctx, set)
-	if err == nil && keys != nil && len(keys.Keys) > 0 {
-		return FindPrivateKey(keys)
-	}
-
 	if errors.Is(err, x.ErrNotFound) || keys != nil && len(keys.Keys) == 0 {
 		r.Logger().Warnf("JSON Web Key Set \"%s\" does not exist yet, generating new key pair...", set)
 		getLock(set).Lock()
--- a/jwk/helper.go
+++ b/jwk/helper.go
@@ -44,13 +44,17 @@ func EnsureAsymmetricKeypairExists(ctx context.Context, r InternalRegistry, alg,
 }
 
 func GetOrGenerateKeys(ctx context.Context, r InternalRegistry, m Manager, set, kid, alg string) (private *jose.JSONWebKey, err error) {
-	getLock(set).Lock()
-	defer getLock(set).Unlock()
-
 	keys, err := m.GetKeySet(ctx, set)
+	if err == nil && keys != nil && len(keys.Keys) > 0 {
+		return FindPrivateKey(keys)
+	}
+
 	if errors.Is(err, x.ErrNotFound) || keys != nil && len(keys.Keys) == 0 {
 		r.Logger().Warnf("JSON Web Key Set \"%s\" does not exist yet, generating new key pair...", set)
+		getLock(set).Lock()
 		keys, err = m.GenerateAndPersistKeySet(ctx, set, kid, alg, "sig")
+		getLock(set).Unlock()
+
 		if err != nil {
 			return nil, err
 		}
@@ -64,7 +68,9 @@ func GetOrGenerateKeys(ctx context.Context, r InternalRegistry, m Manager, set, 
 	} else {
 		r.Logger().WithField("jwks", set).Warnf("JSON Web Key not found in JSON Web Key Set %s, generating new key pair...", set)
 
+		getLock(set).Lock()
 		keys, err = m.GenerateAndPersistKeySet(ctx, set, kid, alg, "sig")
+		getLock(set).Unlock()
 		if err != nil {
 			return nil, err
 		}
--- a/jwk/helper.go
+++ b/jwk/helper.go
@@ -61,22 +61,22 @@ func GetOrGenerateKeys(ctx context.Context, r InternalRegistry, m Manager, set, 
 	privKey, privKeyErr := FindPrivateKey(keys)
 	if privKeyErr == nil {
 		return privKey, nil
-	} else {
-		r.Logger().WithField("jwks", set).Warnf("JSON Web Key not found in JSON Web Key Set %s, generating new key pair...", set)
-
-		getLock(set).Lock()
-		keys, err = m.GenerateAndPersistKeySet(ctx, set, kid, alg, "sig")
-		getLock(set).Unlock()
-		if err != nil {
-			return nil, err
-		}
-
-		privKey, err := FindPrivateKey(keys)
-		if err != nil {
-			return nil, err
-		}
-		return privKey, nil
 	}
+
+	r.Logger().WithField("jwks", set).Warnf("JSON Web Key not found in JSON Web Key Set %s, generating new key pair...", set)
+
+	getLock(set).Lock()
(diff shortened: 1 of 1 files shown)


awill1988:
I doubt this adds anything valuable but just in case, here's a couple ideas:

1. Use `sync.Map` to simplify the `getLock(set string) *RWMutex`
2. Wrap `Persister.GetKeySet` with a transaction

https://github.com/awill1988/hydra/commit/b645b23a08d3bedf7a2278a591409a7002c6064c#diff-cf3d2851cb06a9dfce1d62248d47c33494589220b8c2cca64c576a5d45b42594L167

terev:
> I doubt this adds anything valuable but just in case, here's a couple ideas:
> 
>     1. Use `sync.Map` to simplify the `getLock(set string) *RWMutex`
> 
>     2. Wrap `Persister.GetKeySet` with a transaction
> 
> 
> [awill1988@b645b23#diff-cf3d2851cb06a9dfce1d62248d47c33494589220b8c2cca64c576a5d45b42594L167](https://github.com/awill1988/hydra/commit/b645b23a08d3bedf7a2278a591409a7002c6064c#diff-cf3d2851cb06a9dfce1d62248d47c33494589220b8c2cca64c576a5d45b42594L167)

1. FWIW I agree this seems like a good idea. This use case fits one of their prescribed uses well.
2. Unless I'm missing something, I don't think the transaction would achieve anything there. Since it's a single read there's nothing to rollback.

awill1988:
@terev for 2. the thought was by putting behind a db trx I imagined it could block if there's an insert happening on the table but that's making some broad assumptions about the level of trx isolation per db implementation and likely not the right solution.

I think @alnr had mentioned something along the lines of selecting for insert.. maybe putting the whole function in a db transaction would work?

aeneasr:
> FWIW I agree this seems like a good idea. This use case fits one of their prescribed uses well.

Atomic maps are also just maps with a sync mutex. Unfortunately `sync.Map` is not generic and I don't want to deal with type assertions. So this is fine IMO.

terev:
@awill1988 I'm not 100% sure but I think trying to accomplish this in a single transaction will result in deadlocks. At least for the default isolation level. 

The sync.Map is more sophisticated than just a map protected by a mutex. But since reading from the map isn't on the hot path it seems reasonable to just use a mutex. I noticed we're not using the read lock when accessing the map so maybe we just use a plain mutex?

aeneasr:
> I think @alnr had mentioned something along the lines of selecting for insert.. maybe putting the whole function in a db transaction would work?

He did, but doing INSERT ON CONFLICT on every JWK read is (I would assume) significantly more expensive than just reading from the transaction, as we have to acquire (potentially) a write lock. I really don't think this is a good solution as this functino is being quite a lot (which is why we're getting lock contention in the first place).

alnr:
I'm not suggesting an insert on every read. Just when the select returns no rows.

aeneasr:
Hm, not sure why conformity tests are failing.

PR #3875, opened 2024-11-05
feat: store client in challenge/verified and reduce DB load

aeneasr:
Partially reverts ory/hydra#3857


alnr:
IMO we should do a partial revert. Keep the smaller JSON keys and `omitempty`s.

aeneasr:
Agreed! I did this in my commits :)

PR #3870, opened 2024-11-01
fix: cpu contention on jwk reads + suppress duplicate jwk generation

terev:
Using the reproduction steps included in #3863 I'm able to confirm this fixes the read contention:
```zsh
bombardier -d 30s -t 30s -a -c 270 -l http://localhost:4444/.well-known/openid-configuration
Bombarding http://localhost:4444/.well-known/openid-configuration for 30s using 270 connection(s)
[==============================================================================================================================================================] 30s
Done!
Statistics        Avg      Stdev        Max
  Reqs/sec      3768.36    2589.36   12964.59
  Latency       71.76ms    29.92ms   579.56ms
  Latency Distribution
     50%    67.30ms
     75%    82.45ms
     90%    98.98ms
     95%   113.15ms
     99%   184.51ms
  HTTP codes:
    1xx - 0, 2xx - 112946, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:     7.70MB/s
```

I also noticed the .well-known/jwks.json route generates extra keys (if hammered before initialized) so I changed that handler to utilize the same duplicate suppression code.

## Related issue(s)
Closes #3863 

## Checklist

- [x] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [ ] I have referenced an issue containing the design document if my change
      introduces a new feature.
- [x] I am following the
      [contributing code guidelines](../blob/master/CONTRIBUTING.md#contributing-code).
- [x] I have read the [security policy](../security/policy).
- [x] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security vulnerability, I
      confirm that I got the approval (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [x] I have added tests that prove my fix is effective or that my feature
      works.
- [ ] I have added or changed [the documentation](https://github.com/ory/docs).

PR #3876, opened 2024-11-05
fix: jwk generation incorrectly being skipped

aeneasr:
Unfortunately, ory/hydra#3870 introduced a regression in the JWK generation logic.

Reverts ory/hydra#3870

### DIFF
--- a/cmd/server/helper_cert.go
+++ b/cmd/server/helper_cert.go
@@ -58,7 +58,7 @@ func GetOrCreateTLSCertificate(ctx context.Context, d driver.Registry, iface con
 	}
 
 	// no certificates configured: self-sign a new cert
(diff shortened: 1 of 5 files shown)


terev:
Oh that's too bad. Under what condition was jwk generation being skipped?

aeneasr:
It's basically like a cache for calling the same function twice, which is a side effect in a multi-threaded app. It just behaves not like expected under certain circumstances and returns the wrong keys.

terev:
Yeah but only when there's another inflight call for the same result. Kid was intentionally left out of the key because all callers didn't rely on the kid anyways, it is just a random uuid. 

Does this have to do with multi tenancy? I can imagine the network id would be important there. Maybe that should be included in the flight key?

terev:
@aeneasr Do we need a new issue to fix this? Seems fairly important.

aeneasr:
Just reopen the other one. We're not observing these issues on our prod system, if the fix is easy we're happy to do it, but we're not heavily incentivized to fix it due to lack of commercial demand and already having spent a good portion of time on it.

Issue #3879, opened 2024-11-06
Email claim not present on ID token after issuing with refresh and webhook enabled


3schwartz:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/hydra/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/hydra/blob/master/CONTRIBUTING.md).
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://www.ory.sh/l/sign-up-newsletter).

### Issue Summary

The ID token lacks the email claim when issued using a refresh token, despite having a configured webhook as described in the [Ory Hydra documentation](https://www.ory.sh/docs/hydra/guides/claims-at-refresh#responding-to-the-webhook).

This issue was initially reported in [issue #3852](https://github.com/ory/hydra/issues/3852), which was subsequently closed. However, further investigation has allowed us to isolate the problem with more precision.

### Reproducing the bug

Generate a new Ory environment.

Create a OAuth2 client with scopes `openid`, `offline_access` and `email`.

## Validate works without webhook

Using ex. Postman go through a OIDC flow and validate what ID token has email claim.

Also validate, that ID token after issuing with refresh token, has email claim.
```
curl --location 'https://PROJECT-SLUG.projects.oryapis.com/oauth2/token' \
--header 'Content-Type: application/x-www-form-urlencoded' \
--data-urlencode 'grant_type=refresh_token' \
--data-urlencode 'refresh_token=REFRESH_TOKEN' \
--data-urlencode 'client_id=CLIENT_ID' \
--data-urlencode 'scope=openid email offline_access'
```

## Enable webhook and see email claim disappear

Generate a minimal client which can be used as webhook. Example
```
package main

import (
	"encoding/json"
	"fmt"
	"log"
	"net/http"
)

func claimsHandler(w http.ResponseWriter, r *http.Request) {
	if r.Method != http.MethodPost {
		http.Error(w, "Only POST methid is allowed", http.StatusMethodNotAllowed)
		return
	}

	var data map[string]interface{}
	err := json.NewDecoder(r.Body).Decode(&data)
	if err != nil {
		http.Error(w, "Invalid JSON input", http.StatusBadRequest)
		return
	}

[...]

PR #3889, opened 2024-11-15
Update claims instead of overwrite from webhook response

3schwartz:
According to the [webhook documentation](https://www.ory.sh/docs/hydra/guides/claims-at-refresh#responding-to-the-webhook):

> To accept the token exchange without modification, return a 204 or 200 HTTP status code without a response body.

However, this behavior is currently not functioning as expected. When an empty response is returned, it overwrites the additional claims sent to the webhook, effectively erasing them.

## Related issue(s)

https://github.com/ory/hydra/issues/3879

## Checklist

- [X] I have referenced an issue containing the design document if my change
      introduces a new feature.
- [X] I am following the
      the changes.
- [X] I have added tests that prove my fix is effective or that my feature
      works.

### DIFF
--- a/oauth2/.snapshots/TestAuthCodeWithMockStrategy-strategy=jwt-case=0-description=should_pass_request_if_strategy_passes-should_call_refresh_token_hook_if_configured-hook=new.json
+++ b/oauth2/.snapshots/TestAuthCodeWithMockStrategy-strategy=jwt-case=0-description=should_pass_request_if_strategy_passes-should_call_refresh_token_hook_if_configured-hook=new.json
@@ -14,7 +14,8 @@
         "amr": null,
         "c_hash": "",
         "ext": {
-          "hooked": "legacy"
+          "hooked": "legacy",
+          "sid": ""
         }
       },
       "headers": {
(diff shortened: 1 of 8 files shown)


3schwartz:
Hi @aeneasr, @hperl, @alnr,

Would any of you have some time to take a look at this? I’d really appreciate your input.

aeneasr:
Thank you for the PR and sorry for the slow response! I think the problem here really is with incorrect documentation. It should read:

> To not make any changes to the token payload, mirror it in full:
>
> ```
requestBody := res.Body.Read()
// do nothing
send({
access_token_session: requestBody.session.ext
id_token_session: requestBody.session.id_token.ext
})
```

The issue with this PR is that it doesn't allow us to remove elements from the array and it's a breaking change in the behavior, hence I would prefer that we fix the incorrect documentation

3schwartz:
> Thank you for the PR and sorry for the slow response! I think the problem here really is with incorrect documentation. It should read:
> 
> > To not make any changes to the token payload, mirror it in full:
> > ```
> > requestBody := res.Body.Read()
> > // do nothing
> > send({
> > access_token_session: requestBody.session.ext
> > id_token_session: requestBody.session.id_token.ext
> > })
> > ```
> 
> The issue with this PR is that it doesn't allow us to remove elements from the array and it's a breaking change in the behavior, hence I would prefer that we fix the incorrect documentation

Hi @aeneasr,

Thank you for your response.

You mentioned that we should return `access_token_session: requestBody.session.ext` to match access token extra claims. However, I can only find this documentation regarding the payload: https://www.ory.sh/docs/hydra/guides/claims-at-refresh#webhook-payload, and it seems like it should actually be `access_token_session: requestBody.session.extra`.

The documentation also references the Accept Consent Request API: https://www.ory.sh/docs/hydra/reference/api#tag/oAuth2/operation/acceptOAuth2ConsentRequest. However, this doesn’t seem to align with what is described for adding extra claims.

Could you please clarify this?

aeneasr:
Closing this PR as per my previous comment

PR #3252, opened 2022-09-08
feat: OAuth 2.0 Device Authorization Grant #2416

supercairos:
This PR adds the Device Auth Grant to Ory/Hydra
RFC: https://www.rfc-editor.org/rfc/rfc8628

## Related issue(s)
This PR is related to https://github.com/ory/hydra/issues/2416

See also fosite PR:
https://github.com/ory/fosite/pull/701/files

## Checklist

- [X] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [X] I have referenced an issue containing the design document if my change
      introduces a new feature.
- [X] I am following the
      [contributing code guidelines](../blob/master/CONTRIBUTING.md#contributing-code).
- [X] I have read the [security policy](../security/policy).
- [X] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security. vulnerability, I
      confirm that I got green light (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [X] I have added tests that prove my fix is effective or that my feature
      works.
- [ ] I have added or changed [the documentation](https://github.com/ory/docs).

PR #3912, opened 2024-12-18
feat: implement RFC 8628

nsklikas:
This PR is a continuation of https://github.com/ory/hydra/pull/3851. I have created it from my own personal repo and I have invited people from Ory to contribute so that we can speed up things. I think that most of the comments in the old PR were resolved, but I can copy them to this PR if we wish to keep the discussion history. 

Implements the Device Authorization Grant to enable authentication for headless machines (see https://datatracker.ietf.org/doc/html/rfc8628) 

## Related issue(s)

Implements [RFC 8628](https://datatracker.ietf.org/doc/html/rfc8628). 

This PR is based on the work done on https://github.com/ory/hydra/pull/3252, by @supercairos and @BuzzBumbleBee. That PR was based on an older version of Hydra and was missing some features/tests.

We have prepared a [spec](https://github.com/canonical/hydra/wiki/Device-Flow-Specification), that describes our design and implementation. We have tried to mimic the existing logic in Hydra and not make changes that would disrupt the existing workflows

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

Notes: 
- The current implementation has been manually tested only for `memory` and `postgres` databases. The tests pass all of them.
- Fosite is installed from our fork to ease testing. Once the [relevant PR](https://github.com/ory/fosite/pull/826) in fosite is merged, we will update `go.mod`.

### Testing
To test this you need to built the hydra image:
```console
make docker
```
This will create an image with the name: `oryd/hydra:latest-sqlite`

To run the flow you can use our UI, from https://github.com/canonical/identity-platform-login-ui/tree/hydra-device-test:
```console
git clone git@github.com:canonical/identity-platform-login-ui.git -b hydra-device-test
cd identity-platform-login-ui/
# The image name is hard-coded in the docker-compose file
docker compose up --remove-orphans --force-recreate -d
```

Create a client for Hydra:
```console
docker exec -it identity-platform-login-ui-hydra-1 hydra create client   --endpoint http://localhost:4445   --grant-type authorization_code,refresh_token,urn:ietf:params:oauth:grant-type:device_code --scope openid,offline_access,email,profile --token-endpoint-auth-method client_secret_post
```

Use that client to perform the device flow:
```console
docker exec -it identity-platform-login-ui-hydra-1 hydra perform device-code --client-id <client-id> --client-secret <client-secret> -e http://localhost:4444 --scope openid,offline_access,email,profile
```

The user for logging in is:
- username: `test@example.com`
- password: `test`


consent/handler.go
```diff
@@ -1049,6 +1053,133 @@ func (h *Handler) getOAuth2LogoutRequest(w http.ResponseWriter, r *http.Request,
...
+	if err != nil {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(fosite.ErrServerError.WithWrap(err).WithHint(`'user_code' signature could not be computed`)))
+		return
+	}
+	userCodeRequest, err := h.r.OAuth2Storage().GetUserCodeSession(r.Context(), userCodeSignature, nil)
+	if err != nil {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(fosite.ErrNotFound.WithWrap(err).WithHint(`'user_code' session not found`)))
+		return
+	}
+	err = h.r.RFC8628HMACStrategy().ValidateUserCode(ctx, userCodeRequest, reqBody.UserCode)
```
aeneasr, line 1129: reqBody.UserCode is not actually being validated by validateusercode. We can probably remove it? Or should it be validated?

consent/handler.go
```diff
@@ -1049,6 +1053,133 @@ func (h *Handler) getOAuth2LogoutRequest(w http.ResponseWriter, r *http.Request,
...
+	reqURL, err := url.Parse(f.RequestURL)
+	if err != nil {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(err))
+		return
+	}
+	if reqURL.Query().Get("client_id") == "" {
+		q := reqURL.Query()
+		q.Add("client_id", userCodeRequest.GetClient().GetID())
+		reqURL.RawQuery = q.Encode()
+	}
```
aeneasr, line 1155: Should we not **always** use the client from the usercode request? Why is it possible to override the client here? This is a bit confusing to me as to why we need this.

consent/handler_test.go
```diff
@@ -301,3 +305,357 @@ func TestGetLoginRequestWithDuplicateAccept(t *testing.T) {
...
+	require.EqualValues(t, http.StatusOK, resp.StatusCode)
+
+	var result flow.OAuth2RedirectTo
+	require.NoError(t, json.NewDecoder(resp.Body).Decode(&result))
+	require.NotNil(t, result.RedirectTo)
+	require.Contains(t, result.RedirectTo, requestURL)
+	require.Contains(t, result.RedirectTo, "device_verifier")
+}
+
+func TestAcceptDuplicateDeviceRequest(t *testing.T) {
```
aeneasr, line 374: Is this testing if we can accept the same device challenge twice? If so, we need two tests:

1. Test if we can accept the same device challenge twice and that the device verifier is valid
2. Test if we can accept the device challenge once, use it, then try again, and fail

consent/handler_test.go
```diff
@@ -301,3 +305,357 @@ func TestGetLoginRequestWithDuplicateAccept(t *testing.T) {
...
+				}
+				return json.Marshal(ret)
+			},
+			getURL: func() string {
+				return ts.URL + "/admin" + DevicePath + "/accept?device_challenge=" + challenge
+			},
+			validateResponse: func(resp *http.Response) {
+				require.EqualValues(t, http.StatusBadRequest, resp.StatusCode)
+			},
+		},
```
aeneasr, line 644: Maybe you can add a test case here where a device auth code is already used

consent/strategy_default.go
```diff
@@ -41,6 +41,7 @@ import (
 )
 
 const (
+	DeviceVerificationPath      = "/oauth2/device/verify"
```
aeneasr, line 44: Is this URL coming from some spec?

consent/strategy_default.go
```diff
@@ -1132,11 +1159,11 @@ func (s *DefaultStrategy) HandleOAuth2AuthorizationRequest(
 	ctx, span := trace.SpanFromContext(ctx).TracerProvider().Tracer("").Start(ctx, "DefaultStrategy.HandleOAuth2AuthorizationRequest")
 	defer otelx.End(span, &err)
 
-	loginVerifier := strings.TrimSpace(req.GetRequestForm().Get("login_verifier"))
-	consentVerifier := strings.TrimSpace(req.GetRequestForm().Get("consent_verifier"))
+	loginVerifier := strings.TrimSpace(r.URL.Query().Get("login_verifier"))
+	consentVerifier := strings.TrimSpace(r.URL.Query().Get("consent_verifier"))
```
aeneasr, line 1163: Please revert this or provide a reason

consent/strategy_default.go
```diff
@@ -1155,6 +1182,76 @@ func (s *DefaultStrategy) HandleOAuth2AuthorizationRequest(
...
+	ctx context.Context,
+	w http.ResponseWriter,
+	r *http.Request,
+) (_ *flow.AcceptOAuth2ConsentRequest, _ *flow.Flow, err error) {
+	ctx, span := trace.SpanFromContext(ctx).TracerProvider().Tracer("").Start(ctx, "DefaultStrategy.HandleOAuth2DeviceAuthorizationRequest")
+	defer otelx.End(span, &err)
+
+	deviceVerifier := strings.TrimSpace(r.URL.Query().Get("device_verifier"))
+	loginVerifier := strings.TrimSpace(r.URL.Query().Get("login_verifier"))
+	consentVerifier := strings.TrimSpace(r.URL.Query().Get("consent_verifier"))
```
aeneasr, line 1196: use GetRequestForm

consent/strategy_oauth_test.go
```diff
@@ -1108,6 +1108,133 @@ func TestStrategyLoginConsentNext(t *testing.T) {
...
+			}
+			return *payload
+		})
+	}
+
+	createDefaultClient := func(t *testing.T) *client.Client {
+		c := &client.Client{GrantTypes: []string{"urn:ietf:params:oauth:grant-type:device_code"}}
+		return createClient(t, reg, c)
+	}
+	t.Run("case=should pass if both login and consent are granted and check remember flows as well as various payloads", func(t *testing.T) {
```
aeneasr, line 1163: Can you please expand this test suite?

1. Refresh token rotation should work
2. Add test to ensure only granted scope is actually added/used (e.g. omit openid to remove id token result from the request, or offline_access)
3. Re-use detection works (so doing the same request twice fails)

consent/handler.go
```diff
@@ -1049,6 +1053,133 @@ func (h *Handler) getOAuth2LogoutRequest(w http.ResponseWriter, r *http.Request,
...
+	userCodeSignature, err := h.r.RFC8628HMACStrategy().UserCodeSignature(r.Context(), reqBody.UserCode)
+	if err != nil {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(fosite.ErrServerError.WithWrap(err).WithHint(`'user_code' signature could not be computed`)))
+		return
+	}
+	userCodeRequest, err := h.r.OAuth2Storage().GetUserCodeSession(r.Context(), userCodeSignature, nil)
+	if err != nil {
+		h.r.Writer().WriteError(w, r, errorsx.WithStack(fosite.ErrNotFound.WithWrap(err).WithHint(`'user_code' session not found`)))
+		return
+	}
```
aeneasr, line 1128: If I read this code correctly, we check the user_code here (that is the code the user sees on the TV (qr code / plain text code)). We generate the hmac-sha signature of the code and then use that to look it up in the database.

So the database does not know the available user_codes, only their signatures. Correct?

oauth2/handler.go
```diff
@@ -689,6 +712,182 @@ func (h *Handler) getOidcUserInfo(w http.ResponseWriter, r *http.Request) {
...
+
+	// When this endpoint is called with a valid consent_verifier (meaning that the login flow completed successfully)
+	// there are 3 writes happening to the database:
+	// - The flow is created
+	// - The device auth session is updated (user_code is marked as accepted)
+	// - The OpenID session is created
+	// If there were multiple flows created for the same user_code then we may end up with multiple flow objects
+	// persisted to the database, while only one of them was actually used to validate the user_code
+	// (see https://github.com/ory/hydra/pull/3851#discussion_r1843678761)
+	// TODO: We should wrap these queries in a transaction
```
aeneasr, line 740: TODO is open here

oauth2/handler.go
```diff
@@ -689,6 +712,182 @@ func (h *Handler) getOidcUserInfo(w http.ResponseWriter, r *http.Request) {
...
+		x.LogAudit(r, err, h.r.AuditLogger())
+		h.r.Writer().WriteError(w, r, err)
+		return
+	}
+
+	if err != nil {
+		x.LogError(r, err, h.r.Logger())
+		h.r.Writer().WriteError(w, r, err)
+		return
+	}
```
aeneasr, line 757: Is this necessary? The audit logger is really not something we are using/maintaining. You can probably just log to the regular logger

oauth2/handler.go
```diff
@@ -689,6 +712,182 @@ func (h *Handler) getOidcUserInfo(w http.ResponseWriter, r *http.Request) {
...
+	req.SetUserCodeState(fosite.UserCodeAccepted)
+	session, err := h.updateSessionWithRequest(ctx, consentSession, f, r, req, req.GetSession().(*Session))
+	if err != nil {
+		h.r.Writer().WriteError(w, r, err)
+		return
+	}
```
aeneasr, line 770: Change this to `UPDATE ... SET accepted WHER nid=... AND request_id = ?` ?

oauth2/handler.go
```diff
@@ -689,6 +712,182 @@ func (h *Handler) getOidcUserInfo(w http.ResponseWriter, r *http.Request) {
...
+
+	// Update the OpenID Connect session if "openid" scope is granted
+	if req.GetGrantedScopes().Has("openid") {
+		err = h.r.OAuth2Storage().CreateOpenIDConnectSession(ctx, sig, req.Sanitize([]string{"grant_type",
+			"max_age",
+			"prompt",
+			"acr_values",
+			"id_token_hint",
+			"nonce",
```
aeneasr, line 792: This looks copy pasted - could you make this list a variable for re-use?

oauth2/handler.go
aeneasr: Oh, I'm sorry, it appears that there are quite a lot of values being copied around here

aeneasr: Do you know which values need to be updated here? Is it all of them? Or just some?

consent/strategy_oauth_test.go
aeneasr: I see that you may have been covering this in the oauth2 package. Please confirm :)

persistence/sql/migrations/20241609000001000000_device_flow.cockroach.up.sql
```diff
@@ -0,0 +1,46 @@
...
+
+    FOREIGN KEY (client_id, nid) REFERENCES hydra_client(id, nid) ON DELETE CASCADE,
+    FOREIGN KEY (nid) REFERENCES networks(id) ON UPDATE RESTRICT ON DELETE CASCADE,
+    PRIMARY KEY (device_code_signature, nid)
+);
+
+CREATE INDEX hydra_oauth2_device_auth_codes_request_id_idx ON hydra_oauth2_device_auth_codes (request_id, nid);
+CREATE INDEX hydra_oauth2_device_auth_codes_client_id_idx ON hydra_oauth2_device_auth_codes (client_id, nid);
+CREATE INDEX hydra_oauth2_device_auth_codes_challenge_id_idx ON hydra_oauth2_device_auth_codes (challenge_id);
+CREATE UNIQUE INDEX hydra_oauth2_device_auth_codes_user_code_signature_idx ON hydra_oauth2_device_auth_codes (user_code_signature, nid);
```
aeneasr, line 31: I only found queries for `device_code_signature` (primary key) and `request_id)` - do we need the challenge-id?

Also are we not missing a foreign key onto the oauth2 flow table (I'm not sure if we have that in the other tables)

Since you're referencing `nid` as a foreign key, it will also need an index where it's the prefix.

aeneasr (changes requested):
Very nice! I'll push some minor changes from my side and left a couple of comments.

I primarily cleaned up error handling to make it standards compliant

consent/strategy_default.go
nsklikas: No, the spec does not specify how the user_code should be verified.

consent/handler.go
nsklikas: As discussed in the fosite PR, this method makes sure that the user_code has not expired. The authenticity of the code has been verified by the call to `GetUserCodeSession` above.

consent/handler.go
nsklikas: The initial request should not contain the `client_id`. We don't know the client_id when the user gets to the device verification endpoint, we need to fetch the user_code from the database to get it.

The only reason we are adding the `client_id` to the request is to make the existing endpoints work without too many changes. The `client_id` is parsed in https://github.com/nsklikas/hydra/blob/canonical-master/consent/strategy_default.go#L1211 to construct an authorization request that will be used to do the rest of the flow. If we didn't have this hack, we would have to refactor part of the existing default_strategy. IMHO this is work that should be done after this PR is merged.

I will add a comment to make this clearer.

consent/strategy_default.go
nsklikas: `r` is not a fosite `Requester`. The device verification endpoint is not part of the RFC so it wasn't defined in fosite.

consent/handler.go
nsklikas: That is correct

oauth2/handler.go
nsklikas: Are referring to the logs on https://github.com/ory/hydra/pull/3912/files/14cf7cdc370102da2f487f12d7cdf4945b3e0f48#diff-dc5158991c1ce571f0a59c32b86d6fc64c55fbe11d67cdee44b1b0ae436569dfR743 and https://github.com/ory/hydra/pull/3912/files/14cf7cdc370102da2f487f12d7cdf4945b3e0f48#diff-dc5158991c1ce571f0a59c32b86d6fc64c55fbe11d67cdee44b1b0ae436569dfR748 (lines 743 and 748)?

I can change it, only reason I used that logger is because it is also used in `oAuth2Authorize` as well

oauth2/handler.go
nsklikas: I'm not sure to which line you are referring to.

oauth2/handler.go
nsklikas: It's taken from https://github.com/ory/fosite/blob/master/handler/openid/flow_explicit_auth.go#L29. I will make it a variable here, should I add a comment with a link to fosite?

persistence/sql/migrations/20241609000001000000_device_flow.cockroach.up.sql
nsklikas: > I only found queries for `device_code_signature` (primary key) and `request_id)` - do we need the challenge-id?

The challenge_id is not used as an index, the only reason it is there is to be uniform with the other `oauth2_token/code` tables. AFAICT it is not used there either, but thought maybe there is another reason for it.

> Also are we not missing a foreign key onto the oauth2 flow table (I'm not sure if we have that in the other tables)

Thought that there was some issue with adding that fk, but it seems to work now. Will try to add it.

> Since you're referencing `nid` as a foreign key, it will also need an index where it's the prefix.

Not sure why we need an index for that key. When will it ever be used? Do other tables have this index?

oauth2/handler.go
nsklikas: Yes, we discussed this with @zepatrik. We decided that there are no major issues with this and that it should be fixed later on

consent/strategy_default.go
nsklikas: reverted

consent/handler.go
aeneasr: Yes, understood

consent/handler.go
aeneasr: Ok I see, is it possible that we end up here where the user code / authorization flow and the client id could mismatch?

consent/handler_test.go
aeneasr: Yes, the device_challenge must be unique, otherwise we have a potential replay attack vector

consent/strategy_default.go
aeneasr: thanks! i think it's pretty consistent with existing api naming for hydra so LGTM :)

consent/strategy_oauth_test.go
aeneasr: ping @nsklikas

consent/handler.go
aeneasr: Thank you for clarifying!

.schema/config.schema.json
```diff
@@ -1064,6 +1104,17 @@
...
+        "device_authorization": {
+          "token_polling_interval": {
+            "description": "Sets the starting token polling interval.",
+            "default": "5s",
+            "allOf": [
+              {
+                "$ref": "#/definitions/duration"
+              }
+            ]
+          }
```
aeneasr, line 1116: What is currently the number of chars for the user code? Should we make it configurable? I would vote for 8 characters because 6 is just too likely to be attacked.

driver/config/provider.go
```diff
@@ -397,6 +404,26 @@ func (p *DefaultProvider) fallbackURL(ctx context.Context, path string, host str
...
+}
+
+// GetUserCodeLength returns configured user_code length
+func (c *DefaultProvider) GetUserCodeLength(ctx context.Context) int {
+	return 8
+}
+
+// GetDeviceAuthTokenPollingInterval returns configured user_code allowed symbols
+func (c *DefaultProvider) GetUserCodeSymbols(ctx context.Context) []rune {
+	return []rune(randx.AlphaUpper)
```
aeneasr, line 424: The RFC recommends: https://datatracker.ietf.org/doc/html/rfc8628#section-6.1

What do you think about offering security profiles here?

- alpha-upper
- alpha-upper-lower
- numeric-alpha-upper-lower
- numeric-alpha-upper
- numeric

or alternatively

```
entropy: high | medium | low
```

where high is case sensitive alpha numeric (8 length) medium is alpha upper (lenght 8) and low is numeric (length 9)

persistence/sql/migrations/20241609000001000000_device_flow.cockroach.up.sql
aeneasr: > Thought that there was some issue with adding that fk, but it seems to work now. Will try to add it.

My bad, we need it for the foreign key.

> Not sure why we need an index for that key. When will it ever be used? Do other tables have this index?

It's needed because we have a foreign key. Any foreign key needs in (most?) databases also an index or otherwise CASCADE DELETE takes forever.

However, I think you can change

```
CREATE UNIQUE INDEX hydra_oauth2_device_auth_codes_user_code_signature_idx ON hydra_oauth2_device_auth_codes (user_code_signature, nid);
```

to


```
CREATE UNIQUE INDEX hydra_oauth2_device_auth_codes_user_code_signature_idx ON hydra_oauth2_device_auth_codes (nid, user_code_signature);
```

and then `nid` is covered and you don't need an additional index

persistence/sql/src/YYYYMMDD000001_device_flow/20241609000001000000_device_flow.mysql.up.sql
```diff
@@ -0,0 +1,45 @@
+CREATE TABLE IF NOT EXISTS hydra_oauth2_device_auth_codes
```
aeneasr, line 1: You can remove all files in this directory

aeneasr (changes requested):
A few more comments, but I think we're getting quite close. Preventing replay attacks definitely needs to be addressed before merge

.schema/config.schema.json
nsklikas: Currently we use 8 upper case characters.

aeneasr: thanks!

driver/config/provider.go
nsklikas: I am not sure if we want to mix upper and lower alpha or alpha an numeric, as the spec states it may be hard for the user to differentiate `l` from `I` or `O` from `0`. 

Currently the entropy we get is 26^8, which is more than what is recommended in the RFC. But I guess some users may want to tinker with this config.

I think I prefer the security profiles you proposed to the entropy config. I will push an update for this later.

aeneasr: > I am not sure if we want to mix upper and lower alpha or alpha an numeric, as the spec states it may be hard for the user to differentiate l from I or O from 0.

In my view it really depends on the font, but you're right that they officially discourage it. How about:

- high: alpha-upper 9 chars
- medium: alpha-upper 8 chars
- low: numeric 9 chars

also open to other suggestions.

alnr: https://github.com/ory/x/pull/836

persistence/sql/migrations/20241609000001000000_device_flow.cockroach.up.sql
nsklikas: Understood, made `device_challenge` index unique and added a test to validate it (https://github.com/ory/hydra/pull/3912/commits/9e2e241ad8fd54912897f98bee5957330eeb9f5f). Also changed the user code index as you suggested to have `nid` first.

driver/config/provider.go
nsklikas: Nice, I will change it to:
- high: alpha-num-no-ambiguous 8 chars
- medium: alpha-upper 8 chars
- low: numeric 9 chars

aeneasr:
Let me know when this is good for another review

consent/strategy_oauth_test.go
nsklikas: Sorry, I thought I had answered but it seems my comment was never submitted. Yes they should all be covered in https://github.com/nsklikas/hydra/blob/canonical-master/oauth2/oauth2_device_code_test.go:
1. https://github.com/nsklikas/hydra/blob/canonical-master/oauth2/oauth2_device_code_test.go#L539
2. https://github.com/nsklikas/hydra/blob/canonical-master/oauth2/oauth2_device_code_test.go#L461 and https://github.com/nsklikas/hydra/blob/canonical-master/oauth2/oauth2_device_code_test.go#L487
3. https://github.com/nsklikas/hydra/blob/canonical-master/oauth2/oauth2_device_code_test.go#L736 and https://github.com/nsklikas/hydra/blob/canonical-master/oauth2/oauth2_device_code_test.go#L780

nsklikas:
@aeneasr sorry for the late response, everything should be good for another review now.

oauth2/handler.go
aeneasr: It's fine the way it is now :)

oauth2/handler.go
aeneasr: Damn, looks like my comment was saved on the wrong file. I'm resolving this

aeneasr:
I tried performing the flow using the included CLI tooling, but it failed with this:

```
go run . perform device-code --endpoint  http://127.0.0.1:4444 --client-id $code_client_id --client-secret $code_client_secret
Failed to perform the device authorization request
%!(EXTRA string=oauth2: cannot fetch token: 401 Unauthorized
Response: {"error":"invalid_client","error_description":"Client authentication failed (e.g., unknown client, no client authentication included, or unsupported authentication method). The OAuth 2.0 Client supports client authentication method 'client_secret_basic', but method 'client_secret_post' was requested. You must configure the OAuth 2.0 client's 'token_endpoint_auth_method' value to accept 'client_secret_post'."})exit status 1
```


1. It appears that the error message is not properly formatted (thus the "extra" part)
2. The client should per default use "client_secret_basic" like the other commands

aeneasr:
Or does device auth always use post form submit?

aeneasr:
Would it be possible to get https://github.com/ory/hydra-login-consent-node/pull/123 ready to merge so that the quickstart can be used for the device auth code flow? :)

aeneasr:
I've fixed up the node PR: https://github.com/ory/hydra-login-consent-node/pull/161

And made the CLI changes.

The node PR is still missing the device auth complete screen which we should add

Otherwise it seems like everything is working! I was able to perform the device auth code flow end to end :)

ps: the node pr includes a pre-release of the SDK that has the device auth code flow methods included

nsklikas:
> Or does device auth always use post form submit? From looking at the go oauth2 library it appears to be the case?

It is a little bit weird. I think there was a discussion around this in the original PR as well, but I can't fine it right now. [RFC section 3.1](https://datatracker.ietf.org/doc/html/rfc8628#section-3.1) says:
> The client authentication requirements of [Section 3.2.1 of [RFC6749]](https://datatracker.ietf.org/doc/html/rfc6749#section-3.2.1)
   apply to requests on this endpoint, which means that confidential
   clients (those that have established client credentials) authenticate
   in the same manner as when making requests to the token endpoint, and
   public clients provide the "client_id" parameter to identify
   themselves.

Reading this, I understand that the client must authenticate when making the device authz request. But the golang oauth2 lib does not provide client authentication in that request. There is an open issue about it https://github.com/golang/oauth2/issues/685, that hasn't been addressed for quite some time. My guess is that most people will use public clients for the device flow and that's why they don't care. I think that some of the public providers don't require client authn for this request either, so I was conflicted about whether I should add it or not.

We could make the CLI work by patching the http client used to add the basic auth header, I agree that this would provide better UX. I will give it a shot.

christiannwamba:
> One last thing we need is some docs. I was thinking a dedicated page for this grant type:
> 
> 1. Brief introduction
> 2. End to end example
> 3. Config options
> 4. Implementation of a custom ui for the verification and success screen
> 
> @christiannwamba anything I missed?

I think that covers it. Should go under the guides section.

aeneasr:
What I did:

1. Clean up the SQL migrations - there were a couple of NULL values that should not be null and inconsistencies between the database variants.
2. Change the config structure for device URLs.
3. Add code docs with my understanding of the business logic and simplify/clean up a bit.
4. Wrapped writes in `performOAuth2DeviceVerificationFlow` in a transaction to reduce latency.

What I found and what I believe needs to be addressed:

1. [ ] It does not appear to be possible to reject a device user auth code. I also noticed that the field `DeviceError` is unused (probably because the functionality is missing).
2. [ ] Please verify that  `device_authorization_grant_id_token_lifespan` (and refresh, access) are effective with a test.
3. [ ] `DeviceUserCodeAcceptedAt` is unused as far as I can tell. When in doubt, leave it out :)
4. [ ] Clean of up device table is missing - we are not removing the row like we do with the oidc tables post-use and there is no clean up process. This means that in production you'll quickly run into DB growth issues. Ideally, the `hydra_oauth2_device_auth_codes` row gets removed upon use.
5. [ ] The polling interval is purely informational and is part of the OAuth2 response. Is that understanding correct? If so, please document it as such in the config schema JSON.
6. [ ] Looking at the code I was wondering if we can somehow create a sequence where in `HandleOAuth2DeviceAuthorizationRequest` we can bypass device/login/consent verifiation? I've tried to hack around this but in my view it is not possible - please confirm:
  1. You can not get a login challenge without device verification, and you can't get a login verifier without a login challenge
  2. The same applies to the consent challenge - without a login verifier you can't get a consent verifier
7. [ ] Can you please add missing cypress e2e tests for a public client performing the flow?
8. [ ] In the database, `user_code` is leaking into `request_url` in the flow table (and potentially other tables). This isn't great because the user code should really only be available to the user.
9. [ ] It took me quite some time to understand what is happening with `CreateLoginRequest` which now accepts a flow as well. As far as I understand, this is because we now create the flow (sometimes) as part of the device flow which happens before login. I'm wondering if it wouldn't be simpler to just use the existing device flow and update the state, instead of calling `CreateLoginRequest`.
10. [ ] While we have a fallback URL for device success (`oauth2/fallbacks/device/done`), a fallback URL for device verification is missing (`oauth2/fallbacks/device`).

aeneasr:
Looks like I messed up the tests - very sorry about that!

nsklikas:
> Looks like I messed up the tests - very sorry about that! I‘m currently PTO but trying to squeeze it in :)

Nw, I will try to fix them

aeneasr:
Ideally we would not need the janitor (although the janitor should still try to clean the table) but instead the row is deleted when the token exchange happens:

- Add janitor clean up
- Clean up row after use

Without DB clean up it won't be possible to cut a release as it will cause very large DB growth very quickly

aeneasr:
Regarding rejection - I'm not sure how others handle it. Is it possible to reject the device screen in other implementations?

It probably should at least programmatically be possible - for example when the device verification endpoint decides "uhhhm nope, this should not be handled by that client". But might be a bit of a fabricated use case. I would be ok not having a reject endpoint, I think.

nsklikas:
> Ideally we would not need the janitor (although the janitor should still try to clean the table) but instead the row is deleted when the token exchange happens:
> 
>     * Add janitor clean up
> 
>     * Clean up row after use
> 
> 
> Without DB clean up it won't be possible to cut a release as it will cause very large DB growth very quickly

No problem with adding it, but can you point me to where this happens for the other oauth2 tokens as I can't really find it. 

To be clear what you suggest is that instead of marking the token as used after it is accepted/rejected (soft delete), we should instead delete the row (hard delete), right?

Also keep in mind that this won't protect you from someone malicious who will spam the device_authorization endpoint.

nsklikas:
> Regarding rejection - I'm not sure how others handle it. Is it possible to reject the device screen in other implementations?
> 
> It probably should at least programmatically be possible - for example when the device verification endpoint decides "uhhhm nope, this should not be handled by that client". But might be a bit of a fabricated use case. I would be ok not having a reject endpoint, I think.

I think that most provider simply have a button for accepting the code, no button for rejecting it. 

The [RFC}(https://datatracker.ietf.org/doc/html/rfc8628#section-3.3) says:
> The authorization server prompts the end user to
   identify the device authorization session by entering the "user_code"
   provided by the client.  The authorization server should then inform
   the user about the action they are undertaking and ask them to
   approve or deny the request.  Once the user interaction is complete,
   the server instructs the user to return to their device.

In Hydra it is possible to implement this behavior with the consent API. I am not sure if the reject endpoint will actually ever be used or not. IMHO it is a nice-to-have, but not required for an MVP.

aeneasr:
Hm the spec seems to clearly indicate we need a rejection option. At this point I would like to clarify that my personal expectation is to land this feature/PR with all important aspects covered - I understand everyone has different expectations on what MVP means but for Ory Hydra it implies stability, docs, tests and spec compliance :) That‘s why this project is so popular! As I said before - we are super close and I think from the business logic aspect we have almost all bases covered :)Am 11.02.2025 um 10:34 schrieb Nikos Sklikas ***@***.***>:﻿

Regarding rejection - I'm not sure how others handle it. Is it possible to reject the device screen in other implementations?
It probably should at least programmatically be possible - for example when the device verification endpoint decides "uhhhm nope, this should not be handled by that client". But might be a bit of a fabricated use case. I would be ok not having a reject endpoint, I think.

I think that most provider simply have a button for accepting the code, no button for rejecting it.
The [RFC}(https://datatracker.ietf.org/doc/html/rfc8628#section-3.3) says:

The authorization server prompts the end user to
identify the device authorization session by entering the "user_code"
provided by the client.  The authorization server should then inform
the user about the action they are undertaking and ask them to
approve or deny the request.  Once the user interaction is complete,
the server instructs the user to return to their device.

In Hydra it is possible to implement this behavior with the consent API. I am not sure if the reject endpoint will actually ever be used or not. IMHO it is a nice-to-have, but not required for an MVP.

aeneasr:
It would get removed on token exchange - so after the consent flow is finished. You can search for DeleteOpenIDSession (or similar) in Fosite for this!Am 11.02.2025 um 10:28 schrieb Nikos Sklikas ***@***.***>:﻿

Ideally we would not need the janitor (although the janitor should still try to clean the table) but instead the row is deleted when the token exchange happens:
* Add janitor clean up

* Clean up row after use

Without DB clean up it won't be possible to cut a release as it will cause very large DB growth very quickly

No problem with adding it, but can you point me to where this happens for the other oauth2 tokens as I can't really find it.
To be clear what you suggest is that instead of marking the token as used after it is accepted/rejected (soft delete), we should instead delete the row (hard delete), right?
Also keep in mind that this won't protect you from someone malicious who will spam the device_authorization endpoint.

nsklikas:
> It would get removed on token exchange - so after the consent flow is finished. You can search for DeleteOpenIDSession (or similar) in Fosite for this!

But that is different, `user_code`/`device_code` are similar to the `auth_code`. The `hydra_oauth2_oidc` holds information about the session. I don't see any of the oauth2 tokens being deleted when they are used (one could argue that you should keep the tokens for auditing reasons and delete them afterwards). `authorization_codes` and `refresh_tokens` stay in the database until janitor deletes them.

My question is why would you expect the database to grow that big with the device tokens and not with any of the other tokens?

nsklikas:
> Hm the spec seems to clearly indicate we need a rejection option. At this point I would like to clarify that my personal expectation is to land this feature/PR with all important aspects covered - I understand everyone has different expectations on what MVP means but for Ory Hydra it implies stability, docs, tests and spec compliance :) That‘s why this project is so popular! As I said before - we are super close and I think from the business logic aspect we have almost all bases covered :)

Agreed, what I meant is that IMO the current implementation is RFC compliant. `approve or deny the request` sounds like consent to me and that is already implemented. `deny the request != reject the user_code`, rejecting the user_code is the functionality that we are missing

AFAICT rejecting the `user_code` would be part of the consent UI or it would require a new UI screen to be implemented that would be shown before login (probably does not make much sense). We *could* enhance the reject consent API to also invalidate the `user_code`, but I am not sure I like this.

aeneasr:
> > Hm the spec seems to clearly indicate we need a rejection option. At this point I would like to clarify that my personal expectation is to land this feature/PR with all important aspects covered - I understand everyone has different expectations on what MVP means but for Ory Hydra it implies stability, docs, tests and spec compliance :) That‘s why this project is so popular! As I said before - we are super close and I think from the business logic aspect we have almost all bases covered :)
> 
> Agreed, what I meant is that IMO the current implementation is RFC compliant. `approve or deny the request` sounds like consent to me and that is already implemented. `deny the request != reject the user_code`, rejecting the user_code is the functionality that we are missing

Fair point - the current behavior is fine then in my view!

aeneasr:
> > It would get removed on token exchange - so after the consent flow is finished. You can search for DeleteOpenIDSession (or similar) in Fosite for this!
> 
> But that is different, `user_code`/`device_code` are similar to the `auth_code`. The `hydra_oauth2_oidc` holds information about the session. I don't see any of the oauth2 tokens being deleted when they are used (one could argue that you should keep the tokens for auditing reasons and delete them afterwards). `authorization_codes` and `refresh_tokens` stay in the database until janitor deletes them.
> 
> My question is why would you expect the database to grow that big with the device tokens and not with any of the other tokens?

The authorization code is kept around because we need it for re use detection. The oidc, pkce tables are transient tables which get cleaned up after the state is no longer required.

For the user code, do we need re use detection as well? Meaning that re using the code would invalidate the previous issued tokens? If not, we can and probably should delete it once the token exchange is completed so that the code is freed from the unique constraint. In my understanding, the user/device code is also transient and only needed to perform token exchange. Is that correct?

aeneasr:
ps: we have customers with several TB of data in the individual tables so cleaning them up is definitely important :)

nsklikas:
> The authorization code is kept around because we need it for re use detection. The oidc, pkce tables are transient tables which get cleaned up after the state is no longer required.
> 
> For the user code, do we need re use detection as well? Meaning that re using the code would invalidate the previous issued tokens? If not, we can and probably should delete it once the token exchange is completed so that the code is freed from the unique constraint. In my understanding, the user/device code is also transient and only needed to perform token exchange. Is that correct?

Understood, yes they are transient. Given that retries are part of the flow, I think that re-use detection is not needed. I think that this is something that should be added in fosite, I will open a PR to handle that.

nsklikas:
@aeneasr all of your comments should be resolved now except for:

- [ ] cypress e2e tests
- [ ] docs

AFAICT https://github.com/ory/hydra-login-consent-node/pull/161 needs to be merged, in order to implement the e2e tests. Is there anything that I should do on that end?

I will try to open a PR for the docs soon.

aeneasr:
> @aeneasr all of your comments should be resolved now except for:
> 
> * [ ]  cypress e2e tests
> * [ ]  docs
> 
> AFAICT [ory/hydra-login-consent-node#161](https://github.com/ory/hydra-login-consent-node/pull/161) needs to be merged, in order to implement the e2e tests. Is there anything that I should do on that end?
> 
> I will try to open a PR for the docs soon.

Just let me know if the 161 implementation is correct - then I‘ll merge that one!

nsklikas:
Pushed cypress tests for the device flow. I used https://github.com/ory/hydra-login-consent-node/pull/161 to run them, everything seems to work fine. Since there is no success page in that implementation, I set the success page to something in the client app to make the test implementation easier, I hope that's fine.

nsklikas:
The tests should be fixed when the hydra-login-consent-node PR is merged. I also opened a [PR](https://github.com/ory/docs/pull/2026) for the docs. It will need some work, but most of the required information should be there.

aeneasr:
I've pushed the missing success screen, sorry about that!

nsklikas:
AFAICT a new release is needed for `hydra-login-consent-node` as well in order for the tests to pass.

aeneasr:
It uses master IIRC, so no release required

nsklikas:
AFAICT it uses the latest release on the npm store (see https://github.com/ory/hydra/blob/master/test/e2e/oauth2-client/package.json#L17)

nsklikas:
> I see, I probably confused it with kratos CI - let’s see if this passes: https://github.com/ory/hydra-login-consent-node/releases/tag/v2.4.0-alpha.0

It failed, seems like it's missing a token or something

aeneasr:
Too bad, I will get it released today or tomorrow. Can you PTAL at the merge conflicts?

aeneasr:
https://www.npmjs.com/package/hydra-login-consent-logout/v/2.4.0-pre.3

aeneasr:
To test:

```
code_client=$(go run . create client \
    --endpoint http://127.0.0.1:4445 \
    --grant-type authorization_code,refresh_token,urn:ietf:params:oauth:grant-type:device_code \
    --response-type code,id_token \
    --token-endpoint-auth-method none \
    --format json \
    --scope openid --scope offline \
    --redirect-uri http://127.0.0.1:5555/callback)

code_client_id=$(echo $code_client | jq -r '.client_id')
code_client_secret=$(echo $code_client | jq -r '.client_secret')

go run . perform device-code --endpoint  http://127.0.0.1:4444  --client-id $code_client_id 
```

aeneasr (approved):
What a mountain of work. Thank you @nsklikas, @supercairos and @BuzzBumbleBee. I know it took way longer than you probably have ever thought, but I am very glad that Canonical stepped up and did not give up on this huge project. The result is truly amazing and I am very happy to approve and merge this PR. I will do one last last last round of sanity checks but the great collaboration in recent weeks gives me a ton of confidence that we have solved this huge challenge extremely well. Thank you everyoneryone, again!

aeneasr:
Ok, the latest changes make sense. Let's do it :)

meysam81:
wonderful job everyone :heart: 

@aeneasr when can we expect a github release? :pray:

oauth2/flowctx/encoding.go
```diff
@@ -25,6 +25,8 @@ type (
 const (
 	loginChallenge purpose = iota
 	loginVerifier
+	deviceChallenge
+	deviceVerifier
```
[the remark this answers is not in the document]
nsklikas: Shit, sorry I didn't think of that. Thanks for pointing it out, will try to be mindful of this in the future.

Issue #3916, opened 2025-01-05
bug: `softhsm` package should be `softhsm2`?


polarathene:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/hydra/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/hydra/blob/master/CONTRIBUTING.md).
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://www.ory.sh/l/sign-up-newsletter).

### Describe the bug

The Github Actions CI and Dockerfiles seem to rely on `softhsm` as a package for the HSM tests.

In a PR I recently opened this was a failure as the package could not be found. The version of Ubuntu is not pinned, although I'm not sure why the failure wasn't caught prior to my PR as my PR has nothing to do with that CI workflow AFAIK?

The Docker images use Debian as a base where `softhsm` is still a valid package name, there's also `softhsm2` there and the `softhsm` debian package does bring in `softhsm2` implicitly.

The last version of Ubuntu with the `softhsm` package appears to be 16.04. Github Actions test runner is using `ubuntu-latest`. All newer versions of Ubuntu pacakge that binary in the `softhsm2` package.

### Reproducing the bug

Both Ubuntu and Debian have a `softhsm2` package, only Debian has the `softhsm` pacakge (_brings in `softhsm2`_). Should be resolved and consistent with all package installs preferring the `softhsm2` package instead.

---

Ubuntu failure:

```console
$ docker run --rm -it -e DEBIAN_FRONTEND=noninteractive ubuntu:24.04 bash -c 'apt-get -qq update && apt-get -qq install -y softhsm'

E: Unable to locate package softhsm
```

Ubuntu success by using `softhsm2` instead of `softhsm`:

```console
$ docker run --rm -it -e DEBIAN_FRONTEND=noninteractive ubuntu:24.04 bash -c 'apt-get -qq update && apt-get -qq install -y softhsm2'

Creating config file /etc/softhsm/softhsm2.conf with new version
Setting up libsofthsm2 (2.6.1-2.2ubuntu3) ...
Setting up softhsm2 (2.6.1-2.2ubuntu3) ...
```

Debian success with `softhsm` or `softhsm2`:

```console
[...]

Issue #3685, opened 2023-12-28
`Dockerfile`: Remove `/etc/nsswitch.conf` workaround


polarathene:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/hydra/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/hydra/blob/master/CONTRIBUTING.md).
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

Some of the [`Dockerfile` files in this repo](https://github.com/ory/hydra/tree/master/.docker) (_and likely other Ory projects_) carry an old workaround:

https://github.com/ory/hydra/blob/f9cee328f4695b418542f068b29f6ff6372dc973/.docker/Dockerfile-alpine#L9-L11

https://github.com/ory/hydra/blob/f9cee328f4695b418542f068b29f6ff6372dc973/.docker/Dockerfile-scratch#L5-L7

This is no longer relevant since Alpine 3.16.3 which now includes `/etc/nsswitch.conf`, additionally since Go 1.16 the referenced logic has been changed to not prefer `dns files`, but instead `files dns` (_preferring `/etc/hosts` before querying DNS_).

### Describe your ideal solution

Remove the relevant lines from the `Dockerfile` files across Ory projects.

### Workarounds or alternatives

No issue, just communicating that the content is redundant to maintainers.

### Version

2.1.2

polarathene:
## Context

When `/etc/nsswitch.conf` does not exist, previously Go static builds (`CGO_ENABLED=0`) on Linux would perform a [host lookup](https://unix.stackexchange.com/questions/499792/how-do-etc-hosts-and-dns-work-together-to-resolve-hostnames-to-ip-addresses) by first [querying DNS with `/etc/hosts` as a fallback](https://github.com/golang/go/issues/35305#issuecomment-548791562).
[...]

PR #3921, opened 2025-01-10
fix: improve docker set up

### LINKED ISSUE #3683: `Dockerfile`: Remove `VOLUME` instruction
### Preflight checklist

### Describe your problem

Some of the `Dockerfile` files in this repo use the `VOLUME` instruction: https://github.com/ory/hydra/tree/master/.docker

<details>
<summary>Dockerfile references</summary>

https://github.com/ory/hydra/blob/f9cee328f4695b418542f068b29f6ff6372dc973/.docker/Dockerfile-build#L27

https://github.com/ory/hydra/blob/f9cee328f4695b418542f068b29f6ff6372dc973/.docker/Dockerfile-hsm#L54-L62

https://github.com/ory/hydra/blob/f9cee328f4695b418542f068b29f6ff6372dc973/.docker/Dockerfile-sqlite#L19-L27

</details>

PR that introduced `VOLUME`: https://github.com/ory/hydra/pull/2129

---

**[`VOLUME`](https://docs.docker.com/engine/reference/builder/#volume) is not necessary:**
- A container will persist it's internal state until it's destroyed (_`docker rm container_name`, `docker run --rm`, `docker compose down`, etc_). An anonymous volume is often mistakenly added despite being redundant.
- An implicit anonymous volume [copies data from the image to the host per container instance](https://github.com/rspamd/rspamd/issues/4400#issuecomment-1745881053) created. This is wasteful and **accumulates over runs**.
- When persistence via a volume is actually necessary, it should be explicit (_`docker run --volume ./host/path/:container/path`, or the `volume` key in `compose.yaml`, etc_).
  - **Anonymous** (`-v container/path`):
    > _I want to persist the volume without a human-friendly name to reference_
  - **Named** (`-v my_data:container/path`):
    > _I want a human-friendly name, but not easy access from the host filesystem, just persist my data externally from the container_
  - **Bind mount** (`-v host/path:container/path`):
    > _I want to persist the data at a known location on the host filesystem that I can easily access directly_

I also [provided justification in 2022 for Caddy](https://github.com/caddyserver/caddy-docker/issues/118#issuecomment-1091257278) to do the same, citing various sources from other popular official images that likewise dropped `VOLUME` (_which is effectively a legacy feature that causes more problems implicitly than benefits_).

Remove `VOLUME` from the `Dockerfile` lines referenced (_this would be applicable to other Ory projects too_).

- The redundant copy can be avoided.. if the user provides their own bind mount to the same mount point at runtime (_anonymous & named volumes copy container content by default, bind mounts replace_).
- Alternatively, the `--rm` option will remove the container on exit, additionally discarding the implicitly created anonymous volume. This doesn't prevent writing a copy of the volume data to disk, which if large slows startup.

Ideally though, without a real reason to keep `VOLUME`, it should just be removed from the `Dockerfile`s? 🤷‍♂️ 

### Version

2.1.2

aeneasr:
Improves the docker set up and removes some unused files.

Closes #3914
Closes https://github.com/ory/hydra/issues/3916
Closes https://github.com/ory/hydra/issues/3685
Closes #3683

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


polarathene:
Sorry for the delay in engagement, I've been caught up elsewhere 😅 

Thanks for tackling all of this 💪 I'll do a quick review in a moment.

.docker/Dockerfile-test-hsm
```diff
@@ -0,0 +1,32 @@
+FROM golang:1.23-alpine3.21 AS builder
+
+RUN apk -U --no-cache --upgrade --latest add build-base git gcc bash
```
polarathene, line 3: You shouldn't need `-U` as you don't care about cache? I've not encountered an need for `--upgrade` or `--latest` either, has there been any actual reasons to require that when building the image? Otherwise better to keep it simple:

```suggestion
RUN apk add --no-cache build-base git gcc bash
```

.docker/Dockerfile-test-hsm
```diff
@@ -0,0 +1,32 @@
...
+
+COPY go.mod go.sum ./
+COPY internal/httpclient/go.* ./internal/httpclient
+
+ENV CGO_ENABLED 1
+
+RUN go mod download
+COPY . .
+
+RUN go build -tags sqlite,json1,hsm -o /usr/bin/hydra
```
polarathene, line 16: You've got some `json1` usage here and elsewhere in the repo, but I recall a PR that made an effort to drop that and applied some other changes to favor proper static builds with sqlite as seen in the `Dockerfile-local-build` file:

```bash
go build -ldflags="-extldflags=-static" -tags sqlite,sqlite_omit_load_extension -o /usr/bin/hydra
```

Quoting the [merged commit](https://github.com/ory/hydra/commit/93edc9ad894771c67f46ae2c57ee7e50382d73cd) message from: https://github.com/ory/hydra/pull/3826

> remove json1 build tag everywhere
> This is no longer necessary since SQLite 3.38, and was in fact removed completely in `go-sqlite` v1.14.13

That usage still uses `CGO_ENABLED=1`, so I assume that's fine? The HSM support will still link to that library though (_might be possible to build static?_)

**EDIT:** Actually since the build is done before the package is installed, I assume there isn't any dynamic linking going on with the library, maybe `dlopen()`?

.docker/Dockerfile-local-build
```diff
@@ -18,15 +17,14 @@ RUN go build -ldflags="-extldflags=-static" -tags sqlite,sqlite_omit_load_extens
 
 #########################
 
-FROM gcr.io/distroless/static-debian12:nonroot AS runner
+FROM gcr.io/distroless/static-debian12:debug-nonroot AS runner
```
polarathene, line 20: While busybox for a shell is convenient, it kinda defeats the purpose of distroless.

.docker/Dockerfile-alpine
```diff
@@ -1,20 +1,21 @@
...
-RUN echo 'hosts: files dns' > /etc/nsswitch.conf
+    # Add a user/group for Ory with a stable UID + GID:
+    # NOTE: This only appears relevant for supporting hydra as non-root, otherwise unnecessary.
+    addgroup --system --gid 500 ory
+    adduser --system --uid 500 \
+      --gecos "Ory User" \
+      --home /home/ory \
+      --ingroup ory \
+      --shell /sbin/nologin \
+      ory
```
polarathene, line 15: Just for reference, the changes here don't completely reflect those from https://github.com/ory/hydra/pull/3914

Notably with the sqlite path being dropped. Which as was detailed in the referenced PR would lose the ownership for the `ory` user if a bind mount volume was attached to the same location, requiring the user to manually adjust.

Might want to make note to document these caveats (_or drop the non-root user to simplify and avoid the issue entirely_).

.docker/Dockerfile-alpine
```diff
@@ -1,20 +1,21 @@
 FROM alpine:3.20
 
-RUN addgroup -S ory; \
-    adduser -S ory -G ory -D -H -s /bin/nologin && \
-    apk upgrade --no-cache && \
+RUN <<HEREDOC
+    apk upgrade --no-cache
```
polarathene, line 4: Do you really need this?

```suggestion
```

Likewise the next apk command runs `--upgrade` which AFAIK isn't needed.

.docker/Dockerfile-local-build
```diff
@@ -18,15 +17,14 @@ RUN go build -ldflags="-extldflags=-static" -tags sqlite,sqlite_omit_load_extens
...
-COPY --from=builder --chown=nonroot:nonroot /var/lib/sqlite /var/lib/sqlite
 COPY --from=builder /usr/bin/hydra /usr/bin/hydra
 
-VOLUME /var/lib/sqlite
-
 # Declare the standard ports used by hydra (4444 for public service endpoint, 4445 for admin service endpoint)
 EXPOSE 4444 4445
 
+USER nonroot
+
```
polarathene, line 28: Perhaps it's useful to keep for awareness, but you don't actually need this line as the image already defaults to the `nonroot` user.

```suggestion
```

```console
$ docker inspect gcr.io/distroless/static-debian12:debug-nonroot | jq -r '.[].Config.User'
65532
```

Extending the image like done here will retain that `USER` directive.

quickstart.yml
```diff
@@ -12,6 +12,9 @@
 services:
   hydra:
     image: oryd/hydra:v2.2.0
+    build:
+      context: .
+      dockerfile: .docker/Dockerfile-local-build
```
polarathene, line 17: One notable difference that will happen here is due to the decision to continue using non-root, Alpine (_from Docker Hub pull_) will have `500` UID/GID, while Distroless (_from local build_) will have `65532` UID/GID, which will mess with ownership for the db volume if the user runs both variants at some point.

You should document that caveat or correct for it. Instead of the 500 this PR assigns in the Alpine image, it could probably be 65532 to match distroless.

.docker/Dockerfile-alpine
polarathene: From the referenced PR this was the sqlite related snippet (_useful for testing with sqlite even without a volume as DB state will still persist until container is destroyed, restarts are still valid for persistence_):

```bash
# Create the sqlite directory with ownership to that user and group:
# NOTE: This is required for read/write by SQLite.
# - Path may be a default value somewhere, or only explicitly provided via DSN?
# - Owner/Group is only relevant to permissions allowing the hydra process to read/write to the location.
# - Bind mount volumes will replace the ownership with that of the host directory, requiring correction.
install --owner ory --group ory --directory /var/lib/sqlite
```

polarathene (commented):
Bit more thorough than anticipated 😅 

Hope the feedback helps.

polarathene:
> Closes #3914
> Closes #3916
> Closes #3916

Also looks like you meant to reference another issue to close, not the same one twice

Makefile
```diff
@@ -71,7 +71,8 @@ test-resetdb: node_modules
 # Build local docker images
 .PHONY: docker
 docker:
-	DOCKER_BUILDKIT=1 DOCKER_CONTENT_TRUST=1 docker build --progress=plain -f .docker/Dockerfile-build -t oryd/hydra:${IMAGE_TAG}-sqlite .
+	DOCKER_BUILDKIT=1 DOCKER_CONTENT_TRUST=1 docker build --progress=plain -f .docker/Dockerfile-local-build -t oryd/hydra:${IMAGE_TAG} .
```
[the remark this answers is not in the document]
aeneasr: Thank you! We‘ll for sure remove it at some point but in my experience it’s better to keep a few older flags around as you never know what’s running on that Citrix environment people are trying out Ory on :)

README.md
```diff
@@ -977,10 +979,12 @@ If you wish to check your code changes against any of the docker-compose
 quickstart files, run:
 
 ```shell script
-make docker
-docker compose -f quickstart.yml up # ....
+docker compose -f quickstart.yml up --build # ....
```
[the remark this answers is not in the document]
aeneasr: The idea was to use the prod image per default snd the git state as an exception. I‘ve updated the guide accordingly in the docs

.docker/Dockerfile-local-build
aeneasr: True - but this really is only for local dev and I do know a few people including myself who sometimes need to exec locally i to the pod to do and test something. It should also point as a warning that this isn’t a prod image

.docker/Dockerfile-alpine
aeneasr: It works fine in my testing - other variants did not, which is why i‘ve set it up like this. But yeah, could be I missed something

quickstart.yml
aeneasr: That‘s a good point!!

Makefile
```diff
@@ -88,12 +89,12 @@ quicktest:
 
 .PHONY: quicktest-hsm
 quicktest-hsm:
-	DOCKER_BUILDKIT=1 DOCKER_CONTENT_TRUST=1 docker build --progress=plain -f .docker/Dockerfile-hsm --target test-hsm -t oryd/hydra:${IMAGE_TAG} --target test-hsm .
+	DOCKER_BUILDKIT=1 DOCKER_CONTENT_TRUST=1 docker build --progress=plain -f .docker/Dockerfile-test-hsm  --target test-hsm -t oryd/hydra:${IMAGE_TAG} --target test-hsm .
```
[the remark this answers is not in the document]
aeneasr: Addressed

Makefile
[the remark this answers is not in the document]
aeneasr: I misread, 2023 is definitely enough time

.docker/Dockerfile-alpine
aeneasr: done

quickstart.yml
aeneasr: changed to 65532 for alpine

Makefile
```diff
@@ -88,12 +89,12 @@ quicktest:
...
 
 .PHONY: test-refresh
 test-refresh:
 	UPDATE_SNAPSHOTS=true go test -failfast -short -tags sqlite,sqlite_omit_load_extension ./...
-	DOCKER_BUILDKIT=1 DOCKER_CONTENT_TRUST=1 docker build --progress=plain -f .docker/Dockerfile-hsm --target test-refresh-hsm -t oryd/hydra:${IMAGE_TAG} --target test-refresh-hsm .
+	DOCKER_BUILDKIT=1 DOCKER_CONTENT_TRUST=1 docker build --progress=plain -f .docker/Dockerfile-test-hsm  --target test-refresh-hsm -t oryd/hydra:${IMAGE_TAG} --target test-refresh-hsm .
```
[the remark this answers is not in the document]
aeneasr: addressed

README.md
```diff
@@ -977,10 +979,12 @@ If you wish to check your code changes against any of the docker-compose
...
 ```
 
+Warning, this will override your local image tag with the latest build and can
+lead to unexpected or confusing behavior.
```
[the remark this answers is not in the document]
aeneasr: https://github.blog/changelog/2023-12-14-new-markdown-extension-alerts-provide-distinctive-styling-for-significant-content/ :)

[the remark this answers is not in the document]
aeneasr: addressed

quickstart.yml
```diff
@@ -20,27 +23,32 @@ services:
     volumes:
       - type: volume
         source: hydra-sqlite
-        target: /var/lib/sqlite
+        target: /mount/sqlite
```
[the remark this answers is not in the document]
aeneasr: I was playing around with dokcer compose, reverted

.docker/Dockerfile-test-hsm
aeneasr: Removed, sqlite still requires CGO. There is a non-cgo version of sqlite but it's too difficult to use here.

polarathene:
For context to anyone landing here:
- Unexplained revert: https://github.com/ory/hydra/pull/3923
- PR remerged: https://github.com/ory/hydra/pull/3924

PR #3923, opened 2025-01-11
Revert "fix: improve docker set up"

aeneasr:
Reverts ory/hydra#3921


polarathene:
@aeneasr might I suggest you take a different merge strategy in future for PRs? You appear to be using the default merge commit which interleaves a PR branches individual commits into the target branch history.

Not only does it bring all the individual commits into the target branch, if the branch is long-lived it'll interleave the commits with those from other PRs merged based on timestamps of the individual commits. Without interactive rebase to clean-up commit history this tends to be even worse to navigate.

Merging this revert for example:

> image

I doubt those `synchronize workspaces` commits provide any value to `git blame` or the PR itself (_**EDIT:** Nevermind, they were review feedback changes, title was misleading for what they represented_).

In projects I maintain my personal preference is a squash merge commit. I leave commit history in the PR that is reviewed, it is very rare that the change warrants tailoring multiple pristine commits for the master branch, so this bundles the merge into a single logical commit with a reference to the PR.

Actually the revert commit history seems to lack a merge commit reference, instead the follow-up PR that brought back the changes with the shared revert commits in it's branch history was merged with the PR ref 🤔 

---

Anyway... just thought it might be worth mentioning as squash merge commits IMO make more sense for merging PRs, far easier to track the scope of changes and related discussions properly via `git blame` this way too.

aeneasr:
That was an accident because we activated merge queues recently. I reverted the commits, and squash merged the appropriate commit.

PR #3924, opened 2025-01-11
fix: improve docker set up

### LINKED ISSUE #3683: `Dockerfile`: Remove `VOLUME` instruction

aeneasr:
Improves the docker set up and removes some unused files.

Closes #3914
Closes https://github.com/ory/hydra/issues/3916
Closes https://github.com/ory/hydra/issues/3685
Closes #3683


polarathene:
For context, this PR was originally: https://github.com/ory/hydra/pull/3921#issuecomment-2585402366

PR #3932, opened 2025-01-30
feat: revoke token chain by consent challenge ID

alnr:
This change adds the ability to revoke token chains by "consent challenge ID".

## "Consent sessions"

Each time the user goes through a `GET /oauth2/auth?response_type=code&...` auth code flow, we persist a new "consent session" to the database.

This is independent of whether the user has previously logged in and/or granted consent, or whether the user was actively asked to grant consent by the consent app. A successful journey through the auth code flow results in a new "consent session".

This consent session is uniquely identified by its "consent challenge ID". This ID is obtained from the [`GET /admin/oauth2/auth/requests/consent?consent_challenge=...`](https://www.ory.sh/docs/reference/api#tag/oAuth2/operation/getOAuth2ConsentRequest) API. Note that it is not the same as the `consent_challenge=...` query parameter!

Any access and refresh tokens obtained from a token exchange following that particular user journey are bound to that consent session.

We call the totality of all refresh+access tokens derived from a particular consent session a "token chain".

## Token revocation

Revoking an access token (AT) is simple: send the AT to `/oauth2/revoke` and it is revoked. If this AT was birthed from a refresh token (RT), the RT that birthed it is not revoked.

Revoking a refresh token (RT) also revokes assocated access tokens.

What if I want to revoke a complete token chain given only an access token?

## Revocation by consent challenge ID

During an authorization code flow, save the consent challenge ID into the access token session data:

```
GET /admin/oauth2/auth/requests/consent?consent_challenge=abcdef
```
Response:
```
{
  "acr": ...,
  "challenge": "G_TIM3XABG14UwIgDoT1DRfipjhC1uix" # <- this is the ID we need
  ...
}
```

Accept the consent request:
```
PUT /admin/oauth2/auth/requests/consent/accept?consent_challenge=abcdef
{
  "remember": true,
  "remember_for": 3600,
  "session": {
    "access_token": {
      "ccid": "G_TIM3XABG14UwIgDoT1DRfipjhC1uix"
    }
  },
  ...
}
```

To revoke the token chain associated with this consent challenge ID, use

```
POST admin/oauth2/auth/sessions/consent?consent_challenge_id=G_TIM3XABG14UwIgDoT1DRfipjhC1uix
```

## PR notes

The persistence code and much of the test code are pretty bad. We test implementation not behavior. There are wrong abstractions.

I have deleted `sdk_test.go` because honestly I can't see the point of that whole file.


flow/flow.go
```diff
@@ -69,22 +69,16 @@ const (
...
-//
-// TODO: Deprecate the structs that are made obsolete by the Flow concept.
-// Context: Before Flow was introduced, the API and the database used the same
-// structs, LoginRequest and HandledLoginRequest. These two tables and structs
-// were merged into a new concept, Flow, in order to optimize the persistence
-// layer. We currently limit the use of Flow to the persistence layer and keep
-// using the original structs in the API in order to minimize the impact of the
-// database refactoring on the API.
+// Flow is an abstraction used in the persistence layer.
+// It has outlived its usefulness but is difficult to remove.
```
aeneasr, line 73: I think the explanation before was better to understand for anyone looking to still address this. The current code doc is not clear what "outlived it's usefulness" means. Please revert!

internal/httpclient/api/openapi.yaml
```diff
@@ -1032,7 +1032,7 @@ paths:
         explode: true
         in: query
         name: subject
-        required: true
+        required: false
```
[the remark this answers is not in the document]
alnr: Yes. We previously always required the subject when revoking a consent session.

Now, if we allow revocation by ID, the subject is not necessary any more.

We could, alternatively, keep the subject required even when deleting by consent ID. That would be less ergonomic (have to provide both subject and consent ID) and a bit weird.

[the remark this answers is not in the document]
aeneasr: makes sense thank you!

flow/flow.go
alnr: reverted

consent/sdk_test.go
```diff
@@ -1,263 +0,0 @@
...
-	"github.com/ory/hydra/v2/driver/config"
-	"github.com/ory/hydra/v2/x"
-	"github.com/ory/x/contextx"
-)
-
-func makeID(base string, network string, key string) string {
-	return fmt.Sprintf("%s-%s-%s", base, network, key)
-}
-
-func TestSDK(t *testing.T) {
```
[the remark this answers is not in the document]
alnr: We do use the SDK in a bunch of other tests, which is a good thing!

This particular test file just totally cuts across all layers of abstraction and interfaces. It tests If the SDK works against internal implementational details.

Those implementation details change in this PR, and this test is just completely broken.

The other tests which broke with this PR are actually testing behavior and required only some very minor adjustments.

Let's get rid of this file 💘

alnr:
> Is the consent challenge id set correctly during the get consent flow? Meaning it won't change after we accept the consent challenge? I believe that was the issue when I looked at it. Maybe an e2e test could bring assurance

Check out this test: https://github.com/ory/hydra/pull/3932/files#diff-6d883efffdabd9715dc9872121018df30a5843c81e25dc6c4af2c3edc13fb21cR1194

hperl (approved):
LGTM!

oauth2/oauth2_auth_code_test.go
```diff
@@ -1192,6 +1204,79 @@ func TestAuthCodeWithDefaultStrategy(t *testing.T) {
 				t.Run("strategy=jwt", run("jwt"))
 			})
 
+			t.Run("case=can revoke token chains with ID obtained from consent requests", func(t *testing.T) {
```
zepatrik, line 1207: It would be good to test a real "chain", i.e. `len(chain) > 1` here. Especially since we have the graceful refresh, we can get forks in the chain and I'd expect to have all branches revoked as well, right?

persistence/sql/persister_consent.go
```diff
@@ -306,10 +311,6 @@ func (p *Persister) VerifyAndInvalidateConsentRequest(ctx context.Context, verif
 		return nil, errorsx.WithStack(fosite.ErrInvalidRequest.WithDebug(err.Error()))
 	}
 
-	// We set the consent challenge ID to a new UUID that we can use as a foreign key in the database
-	// without encoding the whole flow.
-	f.ConsentChallengeID = sqlxx.NullString(uuid.Must(uuid.NewV4()).String())
```
zepatrik, line 311: Why exactly do we not update this anymore? I also see it removed in some tests.

zepatrik (commented):
I don't see a blocker, but let's clarify and in case follow up on these minor comments.

persistence/sql/persister_consent.go
alnr: This comment was misleading. There is no requirement for this ID to be a UUID.

We now generate the consent challenge ID once, at the very start of the flow, and keep it stable until we persist it to the database. That is what actually makes this change work.

Initially (before the stateless AEAD flow was introduced), there was no difference between a "challenge" and a "challenge ID"=="consent request ID". Because of compat reasons, these two concepts are still sometimes called by the same name ("challenge").

With this change, we assign a stable ID to each consent request (~user journey through `/oauth2/auth`).

The long URL parameters passed between parties during the flow (`?login_challenge=...&`, `?login_verifier=...`, `?consent_challenge=...`, `?consent_verifier=...`) are a distinct from these IDs.

oauth2/oauth2_auth_code_test.go
alnr: https://github.com/ory/hydra/issues/3941

PR #3938, opened 2025-02-06
feat: stateless logout

### LINKED ISSUE #3836: Make OP-initiated logout stateless
### Preflight checklist

### Ory Network Project

_No response_

### Describe your problem

OP-initiated logouts hit the database many times and leave stale rows in the database without a creation or expiration date.

Should be pretty straightforward to use the stateless AEAD-based flow used in the Authorization Code grant for this purpose as well, allowing us to completely drop this table and make logouts much faster as well.

Add expiration dates to logout requests and let TTL jobs/janitor handle the cleanup.

### Version

all

### Additional Context

_No response_

alnr:
Closes #3836 

Completely eliminates database writes during logout, except session deletion. We can drop table `hydra_oauth2_logout_request` completely -- I didn't in this PR tho.

There is one technically-incompatible SDK change: The LogoutRequest JSON no longer contains field `challenge`. That field hat no use whatsoever beforehand. But we could bring it back if we absolutely wanted to.

There now is no more reuse detection for logout verifiers. But double-submit use cases work as before, so I don't think there is a observable or at least no meaningful difference in behavior.

Deleted a bunch of tests, too 🥰 🔥


consent/handler_test.go
```diff
@@ -27,61 +27,6 @@ import (
 	"github.com/ory/x/sqlxx"
 )
 
-func TestGetLogoutRequest(t *testing.T) {
```
aeneasr, line 30: Shouldn't it still be possible to fetch a logout request?

consent/strategy_default.go
```diff
@@ -855,21 +854,18 @@ func (s *DefaultStrategy) issueLogoutVerifier(ctx context.Context, w http.Respon
...
-			Verifier:    uuid.New(),
-			RequestedAt: sqlxx.NullTime(time.Now().UTC().Round(time.Second)),
-			ExpiresAt:   sqlxx.NullTime(time.Now().UTC().Round(time.Second).Add(s.c.ConsentRequestMaxAge(ctx))),
-			RPInitiated: false,
-
-			// PostLogoutRedirectURI is set to the value from config.Provider().LogoutRedirectURL()
+		now := time.Now().UTC().Round(time.Second)
+		challenge, err := s.r.ConsentManager().CreateLogoutChallenge(ctx, &flow.LogoutRequest{
+			RequestURL:            r.URL.String(),
+			Subject:               session.Subject,
```
aeneasr, line 860: Shouldn't this have an ID to have one unique characteristic per request? Otherwise we might end up with two challenges that have the same payload given that we also round to time.Second

consent/strategy_default.go
```diff
@@ -939,6 +935,7 @@ func (s *DefaultStrategy) issueLogoutVerifier(ctx context.Context, w http.Respon
 		return nil, errorsx.WithStack(fosite.ErrInvalidRequest.
 			WithHint("Logout failed because none of the listed audiences is a registered OAuth 2.0 Client."))
 	}
+	cl.Secret = "" // We don't want to expose the client secret.
```
aeneasr, line 938: Was this a bug prior? Or is this just a safe guard?

consent/strategy_logout_test.go
```diff
@@ -277,20 +278,20 @@ func TestLogoutFlows(t *testing.T) {
...
 		// run once to log out
 		logoutAndExpectPostLogoutPage(t, browser, http.MethodGet, url.Values{}, defaultRedirectedMessage)
 
-		// run again to ensure that the logout challenge is invalid
-		_, _, err := adminApi.OAuth2API.GetOAuth2LogoutRequest(ctx).LogoutChallenge(logoutReq.GetChallenge()).Execute()
-		assert.Error(t, err)
+		require.NotZero(t, logoutChallenge)
 
-		v, _, err := adminApi.OAuth2API.AcceptOAuth2LogoutRequest(ctx).LogoutChallenge(logoutReq.GetChallenge()).Execute()
+		// double-submit: still works
```
aeneasr, line 293: This comment is misleading right or did the behavior change? From what I read here, double-submit results in the same behvaiour as before, does it not?

flow/consent_types.go
```diff
@@ -477,58 +474,27 @@ func (n *OAuth2ConsentRequestOpenIDConnectContext) Value() (driver.Value, error)
...
-	// Challenge is the identifier of the logout authentication request.
-	ID  string    `json:"challenge" db:"challenge"`
-	NID uuid.UUID `json:"-" db:"nid"`
+	// Challenge is used to retrieve/accept/deny the logout request.
+	Challenge string `json:"challenge" db:"challenge"`
 
-	// Subject is the user for whom the logout was request.
-	Subject string `json:"subject" db:"subject"`
+	// Subject is the user for whom the logout was requested.
+	Subject string `json:"subject,omitempty"`
```
aeneasr, line 481: I don't think this will ever be empty

```suggestion
	Subject string `json:"subject"`
```

persistence/sql/persister_consent.go
```diff
@@ -660,79 +660,73 @@ WHERE
...
 	defer otelx.End(span, &err)
 
-	count, err := p.Connection(ctx).
-		RawQuery("UPDATE hydra_oauth2_logout_request SET rejected=true, accepted=false WHERE challenge=? AND nid = ?", challenge, p.NetworkID(ctx)).
-		ExecWithCount()
-	if count == 0 {
-		return errorsx.WithStack(x.ErrNotFound)
-	} else {
-		return errorsx.WithStack(err)
+	_, err = flowctx.Decode[flow.LogoutRequest](ctx, p.r.FlowCipher(), challenge, flowctx.AsLogoutChallenge)
```
aeneasr, line 697: Rejection no longer works? This is a no-op and a breaking change unless rejection now works a different way.

aeneasr (commented):
Deletes a lot of code and DB = nice

Just a few comments / questions

persistence/sql/persister_consent.go
alnr: No, the behavior is the same. If we don't recognize the logout challenge, we return `x.ErrNotFound`. We we do recognize it, we return `nil`.

Previously, on database errors, we would return 500. Those no longer appear of course.

flow/consent_types.go
alnr: done

consent/handler.go
```diff
@@ -927,9 +927,7 @@ func (h *Handler) rejectOAuth2ConsentRequest(w http.ResponseWriter, r *http.Requ
 // Accept OAuth 2.0 Logout Request
 //
 // swagger:parameters acceptOAuth2LogoutRequest
-//
-//lint:ignore U1000 Used to generate Swagger and OpenAPI definitions
-type acceptOAuth2LogoutRequest struct {
+type _ struct {
```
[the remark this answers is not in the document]
alnr: This works! `make sdk` is clean.

consent/strategy_default.go
aeneasr: Right, the nonce makes it unique - thanks!

consent/strategy_logout_test.go
aeneasr: I think the new behavior is preferred/better

consent/strategy_default.go
aeneasr: I see, that makes sense. Client should instead really implement a custom JSON marshaller that removes the secret + a type alias which doesn't. But I see that this is not the case and we have a bunch of `c.Secret=""` everywhere.

consent/handler_test.go
aeneasr: I see now

persistence/sql/persister_consent.go
aeneasr: I think this is more applicable to the logout verifier and if re-using that one can cause these type of issues, than it is to the challenge. The challenge is only solved if the user hits "yes". I think that's fine and we don't need a new challenge every time we present the screen (unless it expired).

alnr: I need to revisit this in the medium term when I have time. I am 85% certain this is not a problem. Need to think about it though.

Note to self: @alnr 
make sure the login session ID is part of the login verifier, so different sessions cannot be revoked with the same verifier. Pretty sure this is already the case.

PR #3940, opened 2025-02-11
fix: cgroups auto-tune before driver configuration

tilgovi:
Move the cgroups auto-tuning into the driver factory so that its effect is automatically applied to the database connection pool defaults.

BREAKING CHANGE: Automatic GOMAXPROCS tuning based on cgroups influences the default database connection limits. Set connection limits explicitly based on the number of CPUs in the deployment to keep previous behavior.

## Related issue(s)

There is no issue filed because the behavior is not clearly documented one way or another. The database documentation states that the default connection limit is twice the number of CPUs, but it is undocumented that GOMAXPROCS changes this default.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/cmd/server/handler.go
+++ b/cmd/server/handler.go
@@ -29,7 +29,6 @@ import (
 	"github.com/rs/cors"
 	"github.com/spf13/cobra"
 	"github.com/urfave/negroni"
-	"go.uber.org/automaxprocs/maxprocs"
 
 	"github.com/ory/graceful"
 	"github.com/ory/x/healthx"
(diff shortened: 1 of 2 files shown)


tilgovi:
I can add tests and update documentation if it's determined that this would be a good change. I hesitate only because this will also influence the janitor and migration commands, and I don't know if it's appropriate to do that. If not, some greater refactoring would be needed due to the way the driver factory is responsible for constructing the configuration options.

tilgovi:
> I can add tests and update documentation if it's determined that this would be a good change. I hesitate only because this will also influence the janitor and migration commands, and I don't know if it's appropriate to do that. If not, some greater refactoring would be needed due to the way the driver factory is responsible for constructing the configuration options.

I guess I could just note the impact on the janitor and migrate commands as part of the breaking changes. Still, I wasn't sure if it was very deliberate to pull the actual execution of the auto-tuning up to the entry points and keep it out of the driver factory.

aeneasr:
I think the change is fine - to be honest it probably won’t have a large effect since nothing runs concurrently except the HTTP server. What‘s the underlying problem the change is addressing?

tilgovi:
> I think the change is fine - to be honest it probably won’t have a large effect since nothing runs concurrently except the HTTP server. What‘s the underlying problem the change is addressing?

I saw that the database configures maximum connections based on `GOMAXPROCS` (if less than `runtime.NumCPUs()`) and expected that the cgroups auto-tune feature would therefore configure maximum database connections based on cgroups quotas.

It's no problem to set `?max_conns` on the DSN and such, but the auto-tuning based on cgroups to align with the auto-tuned `GOMAXPROCS` is maybe nice? Mostly I opened the ticket in case this was the intended behavior and it was just a bug that it didn't work the way I expected.

Issue #3719, opened 2024-02-16
Calling end_session_endpoint with id_token_hint errors when JWK is rotated


dlpetrie:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/hydra/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/hydra/blob/master/CONTRIBUTING.md).
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

After creating/rotating a new JWK, when existing sessions call the `end_session_endpoint` with the `id_token_hint` Hydra is redirecting to the error page with the following error: 
```
error: invalid_request
error_description: The request is missing a required parameter, includes an invalid parameter value, includes a parameter more than once, or is otherwise malformed. go-jose/go-jose: error in cryptographic primitive
```

### Reproducing the bug

1. Successfully login and retrieve an `access_token` and `id_token` from Hydra.
2. Create a new JWK by making a `POST` call to `/admin/keys/hydra.openid.id-token`
3. Logout from the application being redirected to the `end_session_endpoint` with `id_token_hint` populated.
4. Hydra redirects to the error page with error: 
```
The request is missing a required parameter, includes an invalid parameter value, includes a parameter more than once, or is otherwise malformed. go-jose/go-jose: error in cryptographic primitive
```

### Relevant log output

```shell
time=2024-02-15T23:03:15Z level=error msg=An error occurred audience=application error=map[debug: message:invalid_request reason:go-jose/go-jose: error in cryptographic primitive status:Bad Request status_code:400]
```

### Version

2.2.0-pre.1

### On which operating system are you observing this issue?

None

### In which environment are you deploying?

None

### Additional Context

Digging into the code this is the trail I found:

1. Logout Handler calls: `h.r.ConsentStrategy().HandleOpenIDConnectLogout(...)`: https://github.com/ory/hydra/blob/0421fdad0b30b24a1aa5939e69c20e9d0d8b649f/oauth2/handler.go#L129
[...]

Issue #3573, opened 2023-07-13
Wrong KID in headers of ID Token for a Refresh Token


jvisker:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/hydra/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/hydra/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Network](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

When requesting a refresh token we are seeing a case where the headers of the included id_token state one KID value, but in reality it is signed by a newer KID. This behavior occurs with refresh tokens that existed before we added a new keyset for "hydra.openid.id-token". I have been trying to parse through the source code, but I think the crux of the issues is KID is saved in the session and when a new keyset is added it is unaware. If this was a bug that you fixed in a newer version that would be great to know as well. I acknowledge that this may be user error on our part, but I can't figure out how it could be.

### Reproducing the bug

For us the steps to reproduce are:
1. Go through an authcode flow and log in process that gives back an refresh token and id token
2. Add a new keyset for “hydra.openid.id-token”
3. Use the refresh token to get a new access token and id token
4. Inspect the id token and validate that the KID and signature do not work together

### Version

1.11.8

### On which operating system are you observing this issue?

Linux

### In which environment are you deploying?

Docker

aeneasr:
Hi, please upgrade to 2.x as we have done some major reworking in the JWKs logic. Thanks!

tilgovi:
I can confirm a similar experience with JWT access tokens. After performing a key rotation, refresh tokens still use the old `kid` but are signed by the new key.

tilgovi:
I'm not sure if I'm following right, but a quick attempt to debug leads me to maybe these places:

- https://github.com/ory/fosite/blob/master/handler/oauth2/strategy_jwt.go#L131
- https://github.com/ory/fosite/blob/master/token/jwt/jwt.go#L150

When generating an access token with the JWT strategy, it looks like the headers from the existing JWT session get copied over.

[...]

PR #3942, opened 2025-02-12
fix: use up-to-date `kid` in JWT header when refreshing

NilsBarlaug:
When you rotate any of the JWT signing keys (`hydra.jwt.access-token`/`hydra.openid.id-token`, Hydra will (as expected) sign all new JWTs with the new key - including any JWTs from a refresh token grant flow. However, for any refresh token chain that existed before the key rotation Hydra will incorrectly use the `kid` of the old key in the JWT header. This effectively breaks/poisons all issued refresh tokens on key rotation since all future access/id tokens they provide will be invalid. See the new test for a reproducible example.

I consider this to be a _critical_ bug.

## Related issue(s)

#3573, #3719

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.
- [ ] (Not relevant) I have added or changed [the documentation](https://github.com/ory/docs).

## Further Comments

The underlying problem is that the JWT signing mechanism doesn't control the `kid` in the JWT header - it's picked separately and stored in `Session`. So ideally, we would  either instruct Fosite which key to use or respect the key that Fosite chooses. As pointed out in #3573, if we do not set the KIDs Fosite will use the correct `kid`. However, because we include the KIDs in the token hook upfront before we sign the JWTs it would be a breaking change to omit it. It seems non-trivial (for me at least) to refactor the codebase so the KIDs used in the actual JWT signing code is the source of truth for the KIDs provided in the token hook - mostly because the hook must run before the signing.

Therefore, I've opted for being consistent with how it's done for the initial authorization/token exchange. On the start of every refresh token grant we set the KIDs in the `Session` to correspond to the newest keys. This is not a perfect solution, but is a huge improvement over the current situation without making big changes. It goes from "guaranteed" breaking of all JWTs from existing refresh tokens to "very unlikely". Given the severity of this bug, I think it makes sense to first make a smaller fix and then follow up later with larger structural improvements.

What do you think?

### DIFF
--- a/oauth2/handler.go
+++ b/oauth2/handler.go
@@ -1021,6 +1021,32 @@ func (h *Handler) oauth2TokenExchange(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	if accessRequest.GetGrantTypes().ExactOne(string(fosite.GrantTypeRefreshToken)) {
+		var accessTokenKeyID string
+		if h.c.AccessTokenStrategy(ctx, client.AccessTokenStrategySource(accessRequest.GetClient())) == "jwt" {
+			accessTokenKeyID, err = h.r.AccessTokenJWTStrategy().GetPublicKeyID(ctx)
+			if err != nil {
+				h.logOrAudit(err, r)
+				h.r.OAuth2Provider().WriteAccessError(ctx, w, accessRequest, err)
+				events.Trace(ctx, events.TokenExchangeError, events.WithRequest(accessRequest))
+				return
+			}
+		}
+
+		openIDKeyID, err := h.r.OpenIDJWTStrategy().GetPublicKeyID(ctx)
+		if err != nil {
+			h.logOrAudit(err, r)
+			h.r.OAuth2Provider().WriteAccessError(ctx, w, accessRequest, err)
+			events.Trace(ctx, events.TokenExchangeError, events.WithRequest(accessRequest))
+			return
+		}
+
+		if sess, ok := accessRequest.GetSession().(*Session); ok {
+			sess.KID = accessTokenKeyID
+			sess.DefaultSession.Headers.Add("kid", openIDKeyID)
+		}
+	}
+
 	for _, hook := range h.r.AccessRequestHooks() {
 		if err = hook(ctx, accessRequest); err != nil {
 			h.logOrAudit(err, r)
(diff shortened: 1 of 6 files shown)


aeneasr (approved):
Great fix! I think this is the best workaround-y solution to address this :)

Since I‘m reviewing this on my phone it would be good for someone else to also review this. The tests give me decent confidence that this is working as intended.

oauth2/handler.go
```diff
@@ -1021,6 +1021,32 @@ func (h *Handler) oauth2TokenExchange(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	if accessRequest.GetGrantTypes().ExactOne(string(fosite.GrantTypeRefreshToken)) {
```
alnr, line 1024: Maybe a short comment explaining why we do this here?

alnr (approved):
I like it. Optionally add a comment to explain the workaround.

oauth2/handler.go
aeneasr:
```suggestion
	// When refreshing tokens, we want to ensure to use the latest key-id available for signing the
	// potentially JWT-formatted tokens.
	if accessRequest.GetGrantTypes().ExactOne(string(fosite.GrantTypeRefreshToken)) {
```

oauth2/handler.go
```diff
@@ -1021,6 +1021,32 @@ func (h *Handler) oauth2TokenExchange(w http.ResponseWriter, r *http.Request) {
...
+			accessTokenKeyID, err = h.r.AccessTokenJWTStrategy().GetPublicKeyID(ctx)
+			if err != nil {
+				h.logOrAudit(err, r)
+				h.r.OAuth2Provider().WriteAccessError(ctx, w, accessRequest, err)
+				events.Trace(ctx, events.TokenExchangeError, events.WithRequest(accessRequest))
+				return
+			}
+		}
+
+		openIDKeyID, err := h.r.OpenIDJWTStrategy().GetPublicKeyID(ctx)
```
aeneasr, line 1036: Can you please add an if condition to only fetch this key if the flow is an openid flow. I think it should be enough to check if GrantedScope contains `openid`. This will reduce DB load in cases where we don't have an OpenID flow.

aeneasr (commented):
Thanks! To remarks, otherwise LGTM

tilgovi:
It may be that this can be solved more simply by letting fosite handle this, now that ory/fosite#799 is released. I put up a [PR](https://github.com/ory/hydra/pull/3973) to see about that. I kept the E2E tests and the commit authorship, but added myself as a co-author. Instead of adding go code to set the right `kid`, I just removed all of the code that sets a `kid` at all.

aeneasr:
I agree that #3973 is the better approach now that this is solved in fosite!

PR #3947, opened 2025-02-18
fix: revoke by consent request ID

### LINKED ISSUE #3941: Follow-ups to #3932
https://github.com/ory/hydra/pull/3932#discussion_r1946487284

https://github.com/ory/hydra/pull/3932#discussion_r1946538791

alnr:
This is a follow-up to #3932.

This makes several improvements, chiefly restoring backwards compatibility to Hydra before #3932. We now return both the `challenge` as well as the `consent_request_id` from our APIs and distinguish between them clearly.

Closes #3941

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,7 +93,7 @@ quicktest-hsm:
 
 .PHONY: test-refresh
 test-refresh:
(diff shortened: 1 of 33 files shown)


hperl (approved):
LGTM!

PR #3949, opened 2025-02-20
fix: include client_assertion in token hook payload

phooijenga:
We want to use [JWTs for client authentication](https://www.ory.sh/docs/hydra/guides/jwt#jwts-for-client-authentication) in combination with a token hook. 

## Related issue(s)

When a client authenticates via the `private_key_jwt` auth method the payload includes

    client_assertion_type=urn:ietf:params:oauth:client-assertion-type:jwt-bearer

and 

    client_assertion=<signed-jwt>

This works, and the token hook is called, but it does not include the assertion.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

I didn't see any tests for the `private_key_jwt` auth method, so I copied and modified some of the existing tests for _JWT as Authorization Grant_.


aeneasr:
That‘s on purpose - like we don’t include the client_secret, auth code, or other tokens.

phooijenga:
`client_assertion` is not a secret though, and `assertion` _is_ included.

aeneasr:
Would it not be possible to replay the auth grant with the assertion?

phooijenga:
A replay attempt would be rejected because the `jti` was already used; which is required for this flow.

aeneasr:
I see, it will probably more difficult to implement, but what do you think of instead including the client assertion _claims_ and not the token itself, so that the webhook target does not need to verify the JWT?

phooijenga:
Yeah, that would work; I'll see if I can figure out how to do that.

phooijenga:
Getting the claims also requires some changes in Fosite, for which I have created https://github.com/ory/fosite/pull/844. I based it on v0.49.0 to avoid adding unnecessary changes in this MR, which now has a `replace` in the `go.mod` to demonstrate the tests passing.

I'm not sure the approach I took to make the claims available is the right one, and it should probably have a test.

mitar:
I am curious, why do you need access to the assertion itself/claims in the hook?

phooijenga:
We want to validate/verify some of the claims in the assertion, and copy them to the access token.

phooijenga:
Yes, we are aware :)

mitar:
Maybe they have some out-of-band way of verifying content, like additional signatures and stuff. Or maybe it is non-critical stuff like first and last name they allow one just to pick arbitrary.

Issue #3969, opened 2025-03-27
oauth2 webhook: `granted_scopes` always empty for `authorization_code`


sousandrei:
### Preflight checklist

- [x] I could not find a solution in the existing issues, docs, nor discussions.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/ory/hydra/blob/master/CODE_OF_CONDUCT.md).
- [x] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/hydra/blob/master/CONTRIBUTING.md).
- [x] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [x] I am signed up to the [Ory Security Patch Newsletter](https://www.ory.sh/l/sign-up-newsletter).

### Describe the bug

Related to #3620 
But when running OIDC flows with a token hook, the call does not include `granted_scopes`

### Reproducing the bug

Same steps as #3620 

### Relevant log output

```shell

```

### Relevant configuration

```yml

```

### Version

2.3.0

### On which operating system are you observing this issue?

None

### In which environment are you deploying?

None

PR #3891, opened 2024-11-18
Include Requested Scopes in Webhook Requests

3schwartz:
Add requested scopes to the webhook request, allowing the webhook to condition access based on the scopes requested by the user.

## Related issue(s)

Fixes https://github.com/ory/hydra/issues/3620, particularly implementing the idea proposed in this comment: https://github.com/ory/hydra/issues/3620#issuecomment-1745442682.

If this PR is accepted, I can provide an update to the documentation.

## Checklist

- [X] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [X] I have referenced an issue containing the design document if my change
      introduces a new feature.
- [X] I am following the
      [contributing code guidelines](../blob/master/CONTRIBUTING.md#contributing-code).
- [X] I have read the [security policy](../security/policy).
- [X] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security vulnerability, I
      confirm that I got the approval (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [ ] I have added tests that prove my fix is effective or that my feature
      works.
- [ ] I have added or changed [the documentation](https://github.com/ory/docs).

PR #3970, opened 2025-03-27
fix: add scopes to token hook on authorization_code

sousandrei:
## Related issue(s)

#3969 
#3891 

Adding `authorization_code` to conditional of adding scopes to the requests for the token hook

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/oauth2/handler.go
+++ b/oauth2/handler.go
@@ -1178,7 +1178,8 @@ func (h *Handler) oauth2TokenExchange(w http.ResponseWriter, r *http.Request) {
 
 	if accessRequest.GetGrantTypes().ExactOne(string(fosite.GrantTypeClientCredentials)) ||
 		accessRequest.GetGrantTypes().ExactOne(string(fosite.GrantTypeJWTBearer)) ||
-		accessRequest.GetGrantTypes().ExactOne(string(fosite.GrantTypePassword)) {
+		accessRequest.GetGrantTypes().ExactOne(string(fosite.GrantTypePassword)) ||
+		accessRequest.GetGrantTypes().ExactOne(string(fosite.GrantTypeAuthorizationCode)) {
 		var accessTokenKeyID string
 		if h.c.AccessTokenStrategy(ctx, client.AccessTokenStrategySource(accessRequest.GetClient())) == "jwt" {
 			accessTokenKeyID, err = h.r.AccessTokenJWTStrategy().GetPublicKeyID(ctx)


oauth2/handler.go
```diff
@@ -1178,7 +1178,8 @@ func (h *Handler) oauth2TokenExchange(w http.ResponseWriter, r *http.Request) {
 
 	if accessRequest.GetGrantTypes().ExactOne(string(fosite.GrantTypeClientCredentials)) ||
 		accessRequest.GetGrantTypes().ExactOne(string(fosite.GrantTypeJWTBearer)) ||
-		accessRequest.GetGrantTypes().ExactOne(string(fosite.GrantTypePassword)) {
+		accessRequest.GetGrantTypes().ExactOne(string(fosite.GrantTypePassword)) ||
+		accessRequest.GetGrantTypes().ExactOne(string(fosite.GrantTypeAuthorizationCode)) {
```
tilgovi, line 1182: I think the reason that authorization code is left out of here is that a lot of this information is already determined when handling the authorization endpoint. I think the token endpoint should not be receiving the scope and audience parameters in the authorization code flow.

Line 1306 calls `Handler.updateSessionWithRequest` which should grant the scopes. If the scopes aren't available in the token hook, I think something else must be going on.

tilgovi: And I hope the drive-by is welcome! I'm not a maintainer here or anything, just someone who saw your note about this PR on Slack and was curious and has spent some time recently staring at some of this code.

tilgovi: I think what may be happening here is that fosite is supposed to handle this but it's called in two phases. First it's called to handle the flow, then it's called to issue the access token. I think the token hook might be getting called in between these two phases. Moving some code between the two phases in fosite fixes the issue, I think.

I put up a [draft fosite PR](https://github.com/ory/fosite/pull/848) that seems to fix this for me, but haven't checked yet whether tests need to be updated or looked into whether there are other consequences to be aware of.

tilgovi: Just to explain that a little more clearly, maybe. The token endpoint handler does these things:

1. Calls `fosite.OAuth2Provider.NewAccessRequest()`, which calls the fosite code to handle the token request. This is just above your patch.
2. Runs the code around where you've patched here, to handle some things that fosite does not.
3. Calls all the hooks returned from `Registry.AccessTokenHooks()`, just below the code in your patch. That's where the token hook gets called.
4. Calls `fosite.OAuth2Provider.NewAccessResponse()`, which calls the fosite code to generate a response.

So, I think my patch in fosite is moving the code that reads the granted scope and audience, already set and persisted by the consent flow, and updates the token request. It moves this from step 4 to step 1 so that it's visible in the token hook in step 3.

sousandrei: I think you are on to something!
I'll keep this PR in draft until we have a resolution from your's as well if you don't mind

tilgovi: That sounds like the right move!

aeneasr: Hello, this definitely is not correct. This code part is only for machine interaction. User interaction has to be done differently.

aeneasr (changes requested):

PR #3971, opened 2025-04-02
feat: add error reason to OAuth2TokenExchangeError event

alnr:


x/events/events.go
```diff
@@ -68,6 +70,7 @@ const (
 	attributeKeyOAuth2TokenFormat           = "OAuth2TokenFormat"           //nolint:gosec
 	attributeKeyOAuth2RefreshTokenSignature = "OAuth2RefreshTokenSignature" //nolint:gosec
 	attributeKeyOAuth2AccessTokenSignature  = "OAuth2AccessTokenSignature"  //nolint:gosec
+	attributeKeyReason                      = "Reason"
```
aeneasr, line 73: How about `ErrorReason` ?

alnr: Symmetry with Kratos:
https://github.com/ory/kratos/blob/ef2ad444573e4fabd19703e6a0d41f50e009baea/x/events/events.go#L65

jonas-jonas (approved):

x/events/events.go
aeneasr: Unfortunate, I see. I would like us to use ErrorReason though to clearly scope errors like we do with all other variables. Maybe we add errorreason to kratos and keep reason for bc ?

alnr: Come on, this is pure bikeshedding. We have this working fine in Kratos. Reason is every bit as descriptive as ErrorReason, since `Error` is already in the event name.

alnr: renamed

aeneasr: Naming sucks, I agree, but not doing enough diligence on - in particular event attributes and names - gets us into an inconsistent mess of attribute names according to the gusto of the engineer that worked on it that day.

PR #3973, opened 2025-04-05
fix: use up-to-date `kid` in JWT header when refreshing

tilgovi:
This PR is a variation on [#3942](https://github.com/ory/hydra/pull/3942) that attempts to solve the problem by letting fosite set the `kid` header of tokens and removing all of the code to explicitly set this header in JWTs.


tilgovi:
Not sure if there's anything I should do about the CodeQL scanning results task failing. It seems spurious.

oauth2/.snapshots/TestAuthCodeWithMockStrategy-strategy=jwt-case=5-description=should_pass_with_prompt=login_when_authentication_time_is_recent-should_call_refresh_token_hook_if_configured-hook=new.json
```diff
@@ -17,10 +17,7 @@
           "hooked": "legacy"
         }
       },
-      "headers": {
-        "extra": {
-        }
-      },
+      "headers": null,
```
aeneasr, line 20: Please keep the headers.extra keys, as changing them to null will potentially break webhook receivers.

cypress/integration/oauth2/refresh_token.js
```diff
@@ -100,6 +100,61 @@ describe("The OAuth 2.0 Refresh Token Grant", function () {
...
+          }).then(({ body: tokensBefore }) => {
+            const kidsBefore = {
+              accessToken: validateJwtAndGetKid(tokensBefore.access_token),
+              idToken: validateJwtAndGetKid(tokensBefore.id_token),
+            }
+
+            rotateJwks("hydra.jwt.access-token")
+            rotateJwks("hydra.openid.id-token")
+
+            cy.refreshTokenBrowser(client, tokensBefore.refresh_token).then(
```
aeneasr, line 142: This test isn't actually checking that the `kid` is set correctly, it only validates its non-equality. Can you please make sure that the kid is set correctly - either with a regex (expect a non-nil uuid) or some other way?

From the snapshots it looks like it was `public:hydra.jwt.access-token` before. What is the value now?

aeneasr (changes requested):
Thank you and good approach, some ideas to improve it further. I accepted the CodeQL issues

tilgovi:
I pushed up three commits to restore the `extra` key, validate the UUID format, and remove some of the extraneous diff.

tilgovi:
Also, please let me know how you'd like me to resolve conflicts, if at all. I can merge or rebase or squash at your preference.

tilgovi:
I realized the checks wouldn't even run due to the conflict, so I rebased and squashed.

tilgovi:
And, I think the CI jobs flaked, but I can't re-run them. 😞

tilgovi:
@aeneasr anything more I can do here?

bateller (approved):

permfl:
@aeneasr is it possible to get this changed merged?

jvisker:
I am also very interested in seeing this get merged

PR #3975, opened 2025-04-11
chore: improve oauth2 event data

jonas-jonas:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/consent/handler.go
+++ b/consent/handler.go
@@ -521,7 +521,7 @@ func (h *Handler) acceptOAuth2LoginRequest(w http.ResponseWriter, r *http.Reques
 		return
 	}
 
-	events.Trace(ctx, events.LoginAccepted, events.WithClientID(request.Client.GetID()), events.WithSubject(request.Subject))
+	events.Trace(ctx, events.LoginAccepted, events.WithClientID(request.Client.GetID()), events.WithSubject(loginRequest.Subject))
 	h.r.Writer().Write(w, r, &flow.OAuth2RedirectTo{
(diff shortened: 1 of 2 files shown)


consent/handler.go
```diff
@@ -521,7 +521,7 @@ func (h *Handler) acceptOAuth2LoginRequest(w http.ResponseWriter, r *http.Reques
 		return
 	}
 
-	events.Trace(ctx, events.LoginAccepted, events.WithClientID(request.Client.GetID()), events.WithSubject(request.Subject))
+	events.Trace(ctx, events.LoginAccepted, events.WithClientID(request.Client.GetID()), events.WithSubject(loginRequest.Subject))
```
jonas-jonas, line 524: It was noticed, that sometimes the subject is empty. I believe this because we don't transfer over the subject in all cases, but I wasn't able to reproduce.

alnr:
```suggestion
	events.Trace(ctx, events.LoginAccepted, events.WithClientID(request.Client.GetID()), events.WithSubject(handledLoginRequest.Subject))
```

The logic when the subject might be empty is quite convoluted, but this should always work I think. See

https://github.com/ory/hydra/pull/3975/files#diff-567166c33485b3996843a300681ef6f4cba3e7c340c63f91a96edef0582eb908L463

jonas-jonas: Ah, that's better, thanks.

alnr (approved):
Very good improvements! Nice

PR #3980, opened 2025-04-16
feat: add handler for /.well-known/oauth-authorization-server.

jhickmanit:
In order to support OAuth2.1 and some specific integrations that leverage the `/.well-known/oauth-authorization-server` endpoint, this PR adds a handler for the specific endpoint. The `/.well-known/openid-configuration` endpoint already supports all configuration items that conform to this endpoint as seen here: https://datatracker.ietf.org/doc/html/rfc8414


aeneasr (changes requested):
Test please to ensure this works :)

You can probably reuse the existing test for the oidc endpoint

jonas-jonas (approved):

aeneasr (approved):

PR #3988, opened 2025-05-09
chore: fix down migrations for device flow

adamwalach:
changes:
- ~added new file for postgresql~ removed "drop foreign key" statements because removing foreign keys uses different syntax [than mysql](https://github.com/ory/hydra/blob/master/persistence/sql/migrations/20241609000001000000_device_flow.down.sql) and migration fails with:
   ```
   ERROR: syntax error at or near "FOREIGN" (SQLSTATE 42601)
   error executing migrations/20241609000001000000_device_flow.down.sql, sql: ALTER TABLE 
   hydra_oauth2_device_auth_codes DROP FOREIGN KEY IF EXISTS hydra_oauth2_device_auth_codes_challenge_id_fk;
   ```


persistence/sql/migrations/20241609000001000000_device_flow.postgres.down.sql
```diff
@@ -0,0 +1,17 @@
...
+ALTER TABLE hydra_oauth2_flow DROP COLUMN IF EXISTS device_code_request_id;
+ALTER TABLE hydra_oauth2_flow DROP COLUMN IF EXISTS device_verifier;
+ALTER TABLE hydra_oauth2_flow DROP COLUMN IF EXISTS device_csrf;
+ALTER TABLE hydra_oauth2_flow DROP COLUMN IF EXISTS device_was_used;
+ALTER TABLE hydra_oauth2_flow DROP COLUMN IF EXISTS device_handled_at;
+ALTER TABLE hydra_oauth2_flow DROP COLUMN IF EXISTS device_error;
+
+ALTER TABLE hydra_client DROP COLUMN device_authorization_grant_id_token_lifespan;
+ALTER TABLE hydra_client DROP COLUMN device_authorization_grant_access_token_lifespan;
+ALTER TABLE hydra_client DROP COLUMN device_authorization_grant_refresh_token_lifespan;
```
zepatrik, line 17: Why are these not `IF EXISTS`?

adamwalach: I copied the file from the default version. I see no reason for that

adamwalach: Updated both files to include it

persistence/sql/migrations/20241609000001000000_device_flow.postgres.down.sql
```diff
@@ -0,0 +1,17 @@
+ALTER TABLE hydra_oauth2_device_auth_codes DROP CONSTRAINT IF EXISTS hydra_oauth2_device_auth_codes_challenge_id_fk;
+ALTER TABLE hydra_oauth2_device_auth_codes DROP CONSTRAINT IF EXISTS hydra_oauth2_device_auth_codes_client_id_fk;
+ALTER TABLE hydra_oauth2_device_auth_codes DROP CONSTRAINT IF EXISTS hydra_oauth2_device_auth_codes_nid_fk_idx;
```
hperl, line 3: IIRC this would already be covered by `DROP TABLE ...`, right?

hperl (approved):

persistence/sql/migrations/20241609000001000000_device_flow.postgres.down.sql
adamwalach: good catch, I removed those lines and now we again have one "down" migrations for all database types

PR #3989, opened 2025-05-13
test: parallelize and improve

zepatrik:
Some stuff I stumbled upon, there should be no application code affected. Tests for the oauth2 package ran ~5x faster on my machine.

### DIFF
--- a/persistence/sql/migratest/migration_test.go
+++ b/persistence/sql/migratest/migration_test.go
@@ -231,8 +231,8 @@ func TestMigrations(t *testing.T) {
 				})
 
 				t.Run("case=hydra_oauth2_refresh", func(t *testing.T) {
-					rs := []sql.OAuth2RequestSQL{}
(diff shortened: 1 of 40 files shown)


persistence/sql/migratest/migration_test.go
```diff
@@ -231,8 +231,8 @@ func TestMigrations(t *testing.T) {
 				})
 
 				t.Run("case=hydra_oauth2_refresh", func(t *testing.T) {
-					rs := []sql.OAuth2RequestSQL{}
-					require.NoError(t, c.RawQuery(`SELECT signature, nid, request_id, challenge_id, requested_at, client_id, scope, granted_scope, requested_audience, granted_audience, form_data, subject, active, session_data, expires_at	FROM hydra_oauth2_refresh`).All(&rs))
```
zepatrik, line 235: Previously not all fields where in the snapshots.

alnr (approved):

PR #3992, opened 2025-05-16
feat: graceful refresh count limit

zepatrik:
This change adds a counter to the refresh, so that graceful refresh can be limited to a specific number of tries only.

### DIFF
--- a/driver/config/provider.go
+++ b/driver/config/provider.go
@@ -110,7 +110,8 @@ const (
 	KeyExcludeNotBeforeClaim                     = "oauth2.exclude_not_before_claim"
 	KeyAllowedTopLevelClaims                     = "oauth2.allowed_top_level_claims"
 	KeyMirrorTopLevelClaims                      = "oauth2.mirror_top_level_claims"
-	KeyRefreshTokenRotationGracePeriod           = "oauth2.grant.refresh_token.rotation_grace_period" // #nosec G101
+	KeyRefreshTokenRotationGracePeriod           = "oauth2.grant.refresh_token.rotation_grace_period"      // #nosec G101
+	KeyRefreshTokenRotationGraceReuseCount       = "oauth2.grant.refresh_token.rotation_grace_reuse_count" // #nosec G101
 	KeyOAuth2GrantJWTIDOptional                  = "oauth2.grant.jwt.jti_optional"
 	KeyOAuth2GrantJWTIssuedDateOptional          = "oauth2.grant.jwt.iat_optional"
 	KeyOAuth2GrantJWTMaxDuration                 = "oauth2.grant.jwt.max_ttl"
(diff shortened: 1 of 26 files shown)


driver/config/provider.go
```diff
@@ -758,10 +759,24 @@ func (p *DefaultProvider) cookieSuffix(ctx context.Context, key string) string {
...
+	reuseCount := p.getProvider(ctx).IntF(KeyRefreshTokenRotationGraceReuseCount, 0)
+	if reuseCount > math.MaxInt32 {
+		reuseCount = math.MaxInt32
+	} else if reuseCount < 0 {
+		reuseCount = 0
+	}
 	gracePeriod := p.getProvider(ctx).DurationF(KeyRefreshTokenRotationGracePeriod, 0)
-	if gracePeriod > time.Minute*5 {
-		return time.Minute * 5
+	if reuseCount == 0 && gracePeriod > 5*time.Minute {
```
aeneasr, line 775: We'll need to remove the max time check here

aeneasr (changes requested):
- config schema missing
- add a default max value like 3

driver/config/provider.go
aeneasr: nvm

oauth2/oauth2_auth_code_test.go
```diff
@@ -1605,6 +1606,22 @@ func TestAuthCodeWithDefaultStrategy(t *testing.T) {
...
+
+						token := issueTokens(t)
+						token.Expiry = time.Now().Add(-time.Hour * 24)
+
+						for range 3 {
+							_, err := conf.TokenSource(ctx, token).Token()
+							require.NoError(t, err)
+						}
+						_, err := conf.TokenSource(ctx, token).Token()
+						assert.Error(t, err, "Rotating a used refresh token is not possible after the limit is exceeded")
```
aeneasr, line 1623: Do we need to add additional tests to make sure token chains are revoked as well, or is this covered with other tests? We have a bunch of those tests for the grace period

persistence/sql/persister_oauth2.go
```diff
@@ -491,18 +492,21 @@ func (p *Persister) GetRefreshTokenSession(ctx context.Context, signature string
...
 		return row.toRequest(ctx, session, p)
-	} else if gracePeriod > 0 &&
+	}
+
+	if graceful := p.r.Config().GracefulRefreshTokenRotation(ctx); graceful.Period > 0 &&
 		row.FirstUsedAt.Valid &&
-		row.FirstUsedAt.Time.Add(gracePeriod).After(time.Now()) {
+		row.FirstUsedAt.Time.Add(graceful.Period).After(time.Now()) &&
+		(graceful.Count == 0 || // no limit
+			(row.UsedTimes.Valid && row.UsedTimes.Int32 < graceful.Count)) {
```
aeneasr, line 509: I think we need to make explicit the definition of count in the naming, as I'm reading this here the counter starts at 0 and needs to be less than the configured value (that's fine) but it should be clear that the configured value is `<`

aeneasr (commented):
Generally LGTM, needs some docs updates and clarification on how the counter works

persistence/sql/persister_oauth2.go
zepatrik: This check runs before the counter is increased, hence setting the `count` config to 2 means you can reuse the token exactly twice (i.e. create two concurrent chains). If you have a better naming, happy to adjust.

aeneasr: Ok, that's fine! Then we just need to document it that way :)

oauth2/oauth2_auth_code_test.go
aeneasr: It would be - in this file - "token generations" - where we create multiple generations of tokens and then see if revoking one revokes all. I think it should just work, but I'm wondering if we should still add a test case specifically for this.

zepatrik: Added a case now.

PR #4019, opened 2025-08-15
fix: improve JWK key selection logic to select keys based on `use` parameter

subosito:
## Implement JWK key selection based on `use` parameter

- Add `FindPrivateKeyWithUse()` and `FindPublicKeyWithUse()` functions
- Update `GetOrGenerateKeys()` to use "sig" key selection for signing operations
- Maintain backward compatibility with existing functions
- Add comprehensive tests demonstrating the improvement

> With the help of GitHub Copilot Agent


aeneasr:
Hydra does currently not support encryption at all, hence we encourage to not store keys with `use: enc` in Hydra as those would be a no-op. Since ensuring this doesn't break existing use cases and customers is difficult, we will put this on the backlog until we support JWT encryption in hydra (not on the roadmap at the moment).

aeneasr:
Thank you for your great work though here - if we didn’t have thousands of people using Hydra it would be an easy merge!

PR #4028, opened 2025-09-17
feat: add subject query parameter to token deletion endpoint

RacceGatel:
## Related issue(s)

[#discussion 3284](https://github.com/ory/hydra/discussions/3284)

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

When using a token with introspection verification, I needed to be able to disable all access tokens for a specific user for immediate refresh. This doesn't apply to the user's Kratos session, as I only needed to prompt the client to refresh their current token without terminating the session or reauthenticating. I hope this contribution will help others resolve similar cases.

### DIFF
--- a/internal/httpclient/api/openapi.yaml
+++ b/internal/httpclient/api/openapi.yaml
@@ -1265,15 +1265,25 @@ paths:
       - oAuth2
   /admin/oauth2/tokens:
     delete:
-      description: This endpoint deletes OAuth2 access tokens issued to an OAuth 2.0
-        Client from the database.
+      description: |-
+        This endpoint deletes OAuth2 access tokens issued to an OAuth 2.0 Client or Subject from the database.
+
+        Note that you must specify either the `client_id` or the `subject` parameter.
       operationId: deleteOAuth2Token
       parameters:
       - description: OAuth 2.0 Client ID
         explode: true
         in: query
         name: client_id
-        required: true
+        required: false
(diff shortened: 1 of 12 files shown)


RacceGatel:
Unfortunately, this isn't suitable for the "updating permissions" scenario in an id_token or access JWT, as the user must reconfirm the permissions granted to the client on the consent page each time they update(without skip_consent param). In my case, I expect to re-acquire the access token using a refresh token.

I understand the issue with the large table; I'll try to come back with a PR that would be suitable.

aeneasr:
> Unfortunately, this isn't suitable for the "updating permissions" scenario in an id_token or access JWT, as the user must reconfirm the permissions granted to the client on the consent page each time they update(without skip_consent param). In my case, I expect to re-acquire the access token using a refresh token.

It sounds a bit like you're using Access Tokens or ID tokens as login sessions which is not what they are: https://www.ory.sh/blog/oauth2-openid-connect-do-you-need-use-cases-examples.

Generally, you would not revoke access tokens unless the user "revokes" consent to the application, the tokens have been leaked, the client application was hacked. Revoking access tokens because the user "logs out" is misusing access tokens as sessions, if that's what you want to achieve here.

As such it's not very likely at the moment to get a "revoke all user access tokens without impacting consent" feature merged as it is a rare occurrence.

PR #4050, opened 2025-12-11
fix: add omitempty to optional URI fields in Client JSON serialization

### LINKED ISSUE #4044: Allow setting default values for optional DCR values
### Preflight checklist

- [x] I could not find a solution in the existing issues, docs, nor discussions.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/ory/hydra/blob/master/CODE_OF_CONDUCT.md).
- [x] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/hydra/blob/master/CONTRIBUTING.md).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://www.ory.sh/l/sign-up-newsletter).

### Ory Network Project

_No response_

### Describe your problem

When using Dynamic Client Registration (DCR), Hydra responds with empty values (empty string or empty array) for optional client values such as:

- client_uri
- logo_uri
- contacts
- tos_uri
- policy_uri

Several MCP clients ([python-mcp](https://github.com/modelcontextprotocol/python-sdk) and [typescript-mcp](https://github.com/modelcontextprotocol/typescript-sdk)) validate these values and often fail with an error stating that the provided value is not valid (empty string is not a valid url, ...). 
While Hydra allows setting these values when creating a new client manually, they are always empty when using DCR. 

I suggest we should be able to either set default values for these fields via the configuration file, or hydra should not respond with these fields at all if they are not set.

Some links which discuss this problem:

- https://github.com/modelcontextprotocol/typescript-sdk/issues/940 (workaround for some of the fields, but not all)
- https://github.com/modelcontextprotocol/typescript-sdk/issues/543 (validation issue for contacts)
- https://github.com/open-webui/open-webui/pull/19144/files (workaround in application using a MCP client)

### Version

v25.4.0

### Additional Context

_No response_

panbanda:
## Summary

- Add `omitempty` JSON tag to optional URI fields (`policy_uri`, `tos_uri`, `client_uri`, `logo_uri`) and `contacts` in the Client struct
- These fields are now omitted from JSON responses when empty/nil instead of being serialized as `""` or `null`

## Problem

When using Dynamic Client Registration (RFC 7591), Hydra returns empty strings for optional URI fields and `null` for the contacts array when they are not set:

```json
{
  "client_id": "...",
  "logo_uri": "",
  "tos_uri": "",
  "policy_uri": "",
  "client_uri": "",
  "contacts": null
}
```

This breaks strict JSON validators (e.g., Zod) that expect optional fields to either have valid values or be omitted entirely. RFC 7591 specifies these fields as optional, so omitting them when empty is the correct behavior.

### Real-world impact

This issue affects [mcp-remote](https://github.com/geelen/mcp-remote), a tool for connecting to remote MCP (Model Context Protocol) servers. When mcp-remote uses DCR to dynamically register OAuth clients with Hydra, the response fails Zod validation because of the unexpected empty strings and null values for optional fields.

## Changes

- `client/client.go`: Added `omitempty` to 5 optional fields
- `client/client_test.go`: Added test `TestClientJSONOmitsEmptyOptionalURIFields` to verify behavior

## Testing

```bash
go test -v -race ./client/ -run TestClientJSONOmitsEmptyOptionalURIFields
```

Fixes https://github.com/ory/hydra/issues/4044


damianoneill:
Thanks for this fix, we're directly affected by this. Our identity-bridge service acts as a DCR proxy in front of Hydra for MCP clients (Claude, etc.) and we've had to implement response filtering to strip the empty-string URI fields and null contacts just to get past client-side Zod validation.

The omitempty approach is exactly right per RFC 7591, these fields are optional and should be absent when not set, not serialised as empty. The change is minimal and targeted. Hope this can get a maintainer review soon.

KayleeTheMech:
@damianoneill we hit that snag as well and we also strip non-spec compliant empty strings. We're also interested to get this merged.

simonresch:
Thanks for the PR @panbanda .

You're right that the registration response shouldn't return empty strings for optional fields. However, we have to be mindful of existing integrations which could be relying on certain fields always being present in the hydra API. Since the `Client` object is embedded in every flow payload and the admin API this would be a significant breaking change.

We're taking the same fix scoped to the `/oauth2/register` endpoints, which covers the MCP client case and keeps the breaking change contained. I'll link that PR here and close this one once it lands.

PR #4076, opened 2026-03-04
feat: add configuration to skip copying assertion audience in JWT bearer grant type

elatt:
Introduce a configuration option to control whether the audience from the assertion JWT is copied into the resulting access token in the JWT bearer grant type. By default, this behavior remains enabled. The implementation includes updates to the configuration interface, handler logic, and tests to ensure expected functionality.

## Summary by CodeRabbit

* **New Features**
  * Added JWT-bearer grant configuration option to control whether assertion audience claims are propagated to the access token (default: false for backward compatibility).

* **Configuration**
  * New `oauth2.grant.jwt.omit_assertion_audience` boolean setting available for OAuth2 JWT-bearer grant configuration.

* **Tests**
  * Added test coverage for JWT assertion audience handling behavior.


fosite/config_default.go
```diff
@@ -160,6 +161,10 @@ type Config struct {
 	// GrantTypeJWTBearerIssuedDateOptional indicates, if "iat" (issued at) claim required or not in JWT.
 	GrantTypeJWTBearerIssuedDateOptional bool
 
+	// GrantTypeJWTBearerCopyAssertionAudience indicates whether the audience from the assertion JWT should be
+	// copied into the resulting access token. Defaults to true for backwards compatibility.
+	GrantTypeJWTBearerCopyAssertionAudience *bool
```
elatt: I don't like how we need to have this be a pointer just to handle the default unset behavior. I'm thinking of just flipping the name of this param. However, I also wouldn't like inverting the logic of the flag because then you are dealing with a double negative 🤔

elatt:
```suggestion
	// GrantTypeJWTBearerOmitAssertionAudience indicates whether the audience from the assertion JWT should be
	// copied into the resulting access token. Defaults to false for backwards compatibility.
	GrantTypeJWTBearerOmitAssertionAudience bool
```

PR #4124, opened 2026-08-28
fix: copy device verification query into the synthetic authorize request

### LINKED ISSUE #4118: Device verification GET query parameters are not propagated to the authorization request
## Preflight checklist

- [x] I could not find a solution in the existing issues, docs, pull requests, or discussions.
- [x] I agree to follow this project's Code of Conduct.
- [x] I have read and am following this repository's Contribution Guidelines.

## Describe the bug

`HandleOAuth2DeviceAuthorizationRequest` reads control parameters such as `device_verifier` and `client_id` directly from `r.URL.Query()`, but later assigns `ar.Form = r.Form` without first parsing the GET request form.

For a GET request, `r.Form` is not populated until `ParseForm` is called. As a result, authorization parameters carried by `/oauth2/device/verify`, including `prompt=login` and `max_age=0`, are absent from the synthetic authorization request. `requestAuthentication` therefore may reuse a remembered login session instead of requiring fresh authentication.

## Reproducing the bug

1. Complete a device authorization flow and accept the login with `remember=true`, using a browser client that retains Hydra's session cookie.
2. Start a second device authorization flow for the same client.
3. Open the second `verification_uri_complete` with `&prompt=login&max_age=0` appended.
4. Inspect the second login request.

Actual result: the second login request can have `skip=true`, and the resulting ID token can retain the earlier authentication time.

Expected result: the query parameters are propagated to the authorization request, so `prompt=login` causes the second login request to have `skip=false` and produces a fresh authentication time.

## Proposed fix

Populate the synthetic authorization request from the verification request query:

```go
ar.Form = r.URL.Query()
```

A regression test can run two device flows with the same cookie jar, add `prompt=login&max_age=0` to the second verification URL, and assert that both login requests are non-skipped.

## Relevant log output

No specific error is logged. The observable symptom is `skip=true` on the second login request and an unchanged `auth_time` in the resulting ID token.

## Version

Confirmed on Hydra `v25.4.0` and still present on `master` at `4174065ff`.

## Environment

- OS: Linux
- Deployment: self-hosted Kubernetes

## Additional context

A downstream build with the one-line change and the regression test was exercised through a real device login and reauthentication flow. The patched build produced a new authentication time at the explicit reauthentication event and atomically rotated the relying application's session as expected.

AI assistance disclosure: OpenAI Codex was used to inspect the code path, draft this report, and help prepare the regression test. The behavior and proposed fix were independently exercised against a self-hosted deployment.

BetterAndBetterII:
## Related issue(s)

Fixes #4118

`HandleOAuth2DeviceAuthorizationRequest` copied `r.Form` onto the synthetic authorize request. Device verification is a GET, so `r.Form` is empty until `ParseForm`, and query values such as `prompt=login` and `max_age=0` never reached `requestAuthentication`. A remembered login session could then skip authentication on a later device flow.

Copy `r.URL.Query()` instead. A regression in `TestStrategyDeviceLoginConsent` runs two device flows with the same cookie jar, appends `prompt=login&max_age=0` on the second verification URL, and asserts `skip=false`.

## Checklist

- [x] I have read the contributing guidelines.
- [x] I have referenced an issue containing the design document if my change introduces a new feature.
- [x] I am following the contributing code guidelines.
- [x] I have read the security policy.
- [x] I confirm that this pull request does not address a security vulnerability.

## Summary by CodeRabbit

* **Bug Fixes**
  * Device authorization requests now correctly preserve URL parameters such as `prompt`, `max_age`, and `id_token_hint`.
  * Device-flow re-authentication now works as expected when explicitly requested.

* **Tests**
  * Added coverage for login prompts and forced re-authentication during repeated device authorization flows.


BetterAndBetterII:
I have read the CLA Document and I hereby sign the CLA

